PR #3, opened 2015-01-07
Updated README

dyv:
Updated Readme to have information regarding third party dependencies. Noted current installation error that does not indicate a broken installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Install C library dependencies.
 
 OpenSSL: [OpenSSL](https://www.openssl.org/) 
 
-Pairing-Based Cryptographay: [PBC](http://crypto.stanford.edu/pbc/)
+Pairing-Based Cryptography: [PBC](http://crypto.stanford.edu/pbc/)
     
     Mac (with homebrew):   $ brew install pbc
     Otherwise: Install from source.
--- a/README.md
+++ b/README.md
@@ -11,10 +11,36 @@ for details on the library's purpose and functionality.
 Installing
 ----------
 
-To install, first make sure you have [Go](https://golang.org)
-version 1.3 or newer installed, then run:
+First make sure you have [Go](https://golang.org)
+version 1.3 or newer installed.
+
+Install C library dependencies.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,10 +11,36 @@ for details on the library's purpose and functionality.
...
-To install, first make sure you have [Go](https://golang.org)
-version 1.3 or newer installed, then run:
+First make sure you have [Go](https://golang.org)
+version 1.3 or newer installed.
+
+Install C library dependencies.
+
+OpenSSL: [OpenSSL](https://www.openssl.org/) 
+
+Pairing-Based Cryptographay: [PBC](http://crypto.stanford.edu/pbc/)
```
bford: Typo: "Cryptographay"

bford: Also, PBC is actually optional and not actually needed by anything in the crypto library we're currently using, so maybe we should add a note to that effect.  (All that will happen is that directory will fail to build/test.)  We actually need to figure out the "right" way to deal with optional dependencies like this in the Go environment; let me know if you run across any relevant wisdom.

README.md
```diff
@@ -11,10 +11,36 @@ for details on the library's purpose and functionality.
...
 
 	go get github.com/dedis/crypto
+    cd $GOPATH/src/github.com/dedis/crypto
+    go get ./... # install 3rd-party dependencies
+
+It is ok if you get the error:
+
+    # github.com/dedis/crypto/sodium/ed25519
+    sodium/ed25519/curve.go:310: cannot use suite (type \*suite) as type abstract.Suite in return argument:
+        *suite does not implement abstract.Suite (missing Sponge method)
```
bford: Oops, looks like a slight bug due to partly-integrated code; I need to fix that.  (Fine to leave this note in for now until it's fixed.)

README.md
dyv: The best way I can think of to optionally build the packages is to add a build flag (http://golang.org/pkg/go/build/). Then you can do conditional compilation by building with a tag.




You would add:




// +build pbc




To the files you want to ignore. Then when you want to build with pbc you would run  "go build -tags ‘pbc’"

PR #9, opened 2015-01-10
Vanilla ElGamal

iuliatamas:
I attempted a Vanilla ElGamal by keeping all the functionality of sig.go in elgam.go, except for the non-vanilla parts. I tested using a sig_test.go test and 2 tests that were integrating functionality from sig.go and elgam.go to see that they produce the same results. It might not be vanilla enough, case in which I could go from something that looks more like http://en.wikipedia.org/wiki/ElGamal_signature_scheme


bford:
Good, I see you figured out the basic formulation and how to use the crypto library's primitives to implement it.  I'll attach some comments to the code, but two high-level comments:

- The library doesn't really need basic ElGamal signing as a publicly exported function, since it's subsumed by the ring-signature code and the discrete-log proof code - but what the library does need is a clean, simple, educational _example_ of basic ElGamal signing.  For now this could go in the top-level 'crypto' directory alongside 'dh_test.go' and 'en_test.go', which are intended to be "showcase examples" of simple uses of the library's key facilities.  (These examples may need to be moved in a possible future library reorganization, but we'll cross that bridge when we get there.)  So could you package all your code up into a single 'sig_test.go' file, and try to make it as simple and clean as possible?

- In general, you need more comments, and more useful comments.  "generate signature" within the signing function isn't a very useful comment for example since that's what the whole function does.

anon/elgam.go
```diff
@@ -0,0 +1,82 @@
+package anon
+
+import (
+	"bytes"
+	"errors"
+	"crypto/cipher"
+	"github.com/dedis/crypto/abstract"
+)
+
+type basicSig struct {
```
bford: You don't really need this typedef, since you're going to be encoding the signature into a []byte before it leaves the Sign function anyway.

anon/elgam.go
```diff
@@ -0,0 +1,82 @@
...
+	S  abstract.Secret
+}
+
+func signH1preElGam(suite abstract.Suite, message []byte) []byte {
+	H := suite.Hash()
+	H.Write(message)			
+	return H.Sum(nil)
+}
+
+func signH1ElGam(suite abstract.Suite, H1pb []byte, PG abstract.Point) abstract.Secret {
```
bford: I don't see why this two-stage hash is needed; why not just include the message rather than the hash of the message in this second one?

anon/elgam.go
```diff
@@ -0,0 +1,82 @@
...
+	H1.Write(H1pb)
+	H1.Write(PG.Encode())
+
+	b := H1.Sum(nil)
+	s := suite.Stream(b[:suite.KeyLen()])
+	return suite.Secret().Pick(s)
+}
+
+
+func SignElGam(suite abstract.Suite, random cipher.Stream, message []byte,
```
bford: Please add documentation-comments, at least brief ones, before each function (or at least public functions).  

In particular, since this is an implementation of a well-known crypto primitive, include in the comment a pointer (URL) to whatever formulation you used here or found to be most clear.

anon/elgam.go
```diff
@@ -0,0 +1,82 @@
...
+	publicKey abstract.Point , privateKey abstract.Secret) []byte {
+
+	// generate signature
+	H := signH1preElGam(suite, message)
+
+	u := suite.Secret().Pick(random)
+	UB := suite.Point().Mul(nil,u)
+
```
bford: Where do these variable names come from?  Unless they're intended to match up with some very well-known, definitive formulation of ElGamal, I would prefer if variable naming followed the Camenisch-Stadler terminology: e.g., v for the random private value, T for the public point commitment to that value, just c rather than c0 for the challenge, and r rather than s for the response.  

On the other hand, if you did find some very definitive formulation, it might indeed be better to be consistent with that.

anon/elgam_test.go
```diff
@@ -0,0 +1,141 @@
...
+	// 00000000  aa ed 0d ac 14 0d 60 d9  88 8b 3c c2 86 0b f5 79  |......`...<....y|
+	// 00000010  bd 1e ec 0c e8 03 83 44  9b 3b 64 c5 14 4e 87 c4  |.......D.;d..N..|
+	// 00000020  58 8c a0 be c8 4c 15 c9  83 56 70 60 23 7a bd 2a  |X....L...Vp`#z.*|
+	// 00000030  1b 64 f1 71 09 ea 0f b3  0d 3d 05 a7 c0 95 c0 94  |.d.q.....=......|
+	// Signature verified against correct message.
+	// Verifying against wrong message: invalid signature
+}
+
+// testing ElGam, ElGam  generated signature, sig.go verified signature
+func TestElGamAgainstSig1( t *testing.T) {
```
bford: Good, but since this is just an example code, just one small test is probably adequate - again see crypto/enc_test.go.  Simplicity and understandability are more important than exhaustiveness here.

bford:
Very nice, thanks Iulia!

PR #16, opened 2015-01-12
Continuous integration support for absolute paths.

dyv:
This pull request attempts to fix the absolute paths problem. It at the very least does not break anything and gives us more control over Travis. It ignores all of the defaults and runs our own commands instead.


bford:
Thanks Dylan, let's give it a try!

PR #19, opened 2015-01-23
Add missing initialization in PubPoly.Decode.

jackowitzd2:
Outstanding fix that never got pushed. ```PubPoly.Decode()``` after ```Init()``` tries to work with nil points and then fails. Adding the initialization in ```Decode``` seems consistent with the other functions, but there may be argument for putting it in ```Init``` instead.


jackowitzd2:
Whoops, my branch is a bit behind. Either needs to be cherry picked or I'll put it on top of the most recent HEAD.

bford:
Oops!  Thanks for catching this.  Happy to merge it; why did you close?  You can update your poly_decode branch wrt the master - and this pull request will automatically track it - just by doing 'git merge origin/master' followed by a 'git push' back to the poly_decode branch.

I notice I neglected ever to create a test-suite for this package; since it sounds like you've been exercising it anyway, would you (or someone) mind throwing a few tests and/or benchmarks into a sharing_test.go according to the standard Go testing model?  William Bailey might be interested in helping, since he's working on the life-insurance work-item that's supposed to use this polynomial code.

WEB3-GForce:
I'll start working on the test suite. I'll create a branch and make a pull request so we can have a place for further discussion.

jackowitzd2:
Tracked down why the builds were failing - should be all good to merge now.

bford:
Looks good, thanks!

PR #20, opened 2015-01-24
Initial commit for the testing suite.

WEB3-GForce:
I've opened up this pull request for the test suite for poly/sharing  so that it will be easier to discuss the code as I work on it.

I'll be adding to the branch to flesh out the code.

Daniel Jackowitz, do you have any general suggestions or tips?


WEB3-GForce:
How thorough do we want to be in the testing suite (particularly in regards to checking input)?
For example, in the Pick function for PriPoly, what should the function do if it is handed:
- A negative k
- a null abstract.Group or cipher.Stream

Should we be concerned about these cases or focus on testing core functionality?

bford:
Hi William,

- When you shout out to someone in particular, e.g., Danny Jackowitz, look up and use their github name e.g. @jackowitzd2  to make sure they get E-mail notification.

- Don't worry about testing bad-input behaviors that will automatically and naturally just be caught by Go and produce runtime panics (e.g., negative k or null input pointers).  Focus on semantic tests.  DO test for bad input scenarios that could arrive *over the network*: e.g., truncated or otherwise malformed messages received to a parsing function, etc.  (That probably doesn't apply to this particular package though since it currently does no message parsing.)

Cheers
Bryan

bford:
Also, since you're touching code that wasn't properly gofmt'ed before, go ahead and gofmt the files you touch (which should make the Travis CI check happy).  But commit the gofmt as a separate git commit from your other "semantic" changes so that we can 'git diff' to separate the semantic changes from the formatting changes if necessary.  Thanks.

jackowitzd2:
Sorry for the delay in replying. Looks good. We probably also want to add tests of Encode() and Decode(), to verify that the pending fix in #19 is correct, once I get that merged.

Danny

WEB3-GForce:
Thanks Professor Ford! Will do.

Sounds good Danny. I'll be working on tests for public polynomials and shares today. I'll work on the encode/ decode tests when the changes are merged.

WEB3-GForce:
@jackowitzd2
@bford 

I'm getting closer to being done. I'm currently blocked by issue #21. That will need to be resolved before I can verify the tests for PubPoly and PubShares. Once Push #19 is out, I can make sure my tests works for decode/encode.

Concerning the Eval function for PriPoly and PubPoly, do you all have any suggestions for testing this? The ideas I can think of at the moment would be to reproduce the calculations in the function (which wouldn't be as good) or to do it on paper and make sure the result was expected (general pointers for how I could do this would be helpful).

However, I do test Eval indirectly through my tests of the split/ combine methods for private and public polynomials. This process relies on Eval. Hence, I have indirectly verified it. Is this sufficient for our purposes?

bford:
Tests for the Eval functions:

1. Invoke PriPoly.Eval for a few values of parameter i, and PubPoly.Eval on the equivalent polynomial-commitment for the same values of parameter i, then test that the results correspond: i.e., the invariant P(i) = s(i)*B should hold.

2. Invoke PriPoly.Eval(0) and make sure that yields the committed secret (i.e., term 0).

That's probably reasonable, especially given that Eval gets tested indirectly as well.  Does this help?

B

WEB3-GForce:
I was able to get part 1 working. However, number 2 is giving me some issues. When I run the test, it always fails. My guess is that in the code (see below), we always have the x coordinate set to at least 1. Hence, we can never get just the constant part at the end (aka the original secret).

Would you like me to drop the second test and leave the code as is? Given the convention of starting array indices at 0, I can see it being useful to map i =0 to xi = 1. I could imagine a case where someone enters 0 to produce the share for party 0.

func (p *PriPoly) Eval(i int) abstract.Secret {
	g := p.g
	k := len(p.s)
	xi := g.Secret().SetInt64(1 + int64(i)) // x-coordinate of this share
	sv := g.Secret().Zero()
	for i = k - 1; i >= 0; i-- {
		sv.Mul(sv, xi)
		sv.Add(sv, p.s[i])
	}
	return sv
}

WEB3-GForce:
Tests pass. Travis is happy. Ready to merge.

poly/sharing_test.go
```diff
@@ -0,0 +1,544 @@
...
+func producePriPoly(g abstract.Group, k int, s abstract.Secret) *PriPoly {
+	return new(PriPoly).Pick(g, k, s, random.Stream)
+}
+
+func produceTwoPriPoly(g abstract.Group, k int, s abstract.Secret) (*PriPoly,
+	*PriPoly) {
+	return producePriPoly(g, k, s), producePriPoly(g, k, s)
+}
+
+func produceThreePriPoly(g abstract.Group, k int, s abstract.Secret) (*PriPoly,
```
bford: Do you really need three versions of this function?  How about just keeping producePriPoly and calling it 2 or 3 times in the few places where you need 2 or 3 polynomials?

poly/sharing_test.go
```diff
@@ -0,0 +1,544 @@
...
+// secret it was initialized with.
+func TestPriPolySecret(t *testing.T) {
+	testPoly := producePriPoly(group, k, secret)
+	if !secret.Equal(testPoly.Secret()) {
+		t.Error("This secret differs from the one given to it.")
+	}
+}
+
+// Verify that Equal returns true for two polynomials that are the same
+func TestPriPolyEqual_Same(t *testing.T) {
```
bford: I think in this and a lot of your test functions where you have multiple TestPriPoly<func>_<testcase> functions, it would probably be better and produce more concise and simpler test-code if you just merge all the <testcase> variant functions into a single TestPriPoly<func> function that performs all of the test cases.  In this case, for example, just have a TestPriPolyEqual function that first performs a same-polynomial test and then a different-polynomial test.

poly/sharing_test.go
```diff
@@ -0,0 +1,544 @@
...
+	testPoly := producePriPoly(group, k, secret)
+	testPolyCopy := testPoly
```
bford: No need for a 'testPolyCopy' variable, just use 'testPoly' in the Equal below: there's no need to test whether Go's assignment operator produces a valid copy of testPoly.

poly/sharing_test.go
```diff
@@ -0,0 +1,544 @@
...
+	if !testPoly.Equal(testPolyCopy) {
+		t.Error("Polynomials are expected to be equal.")
+	}
+}
+
+// Verify that Equal returns false for two polynomials that are diffferent
+func TestPriPolyEqual_Different(t *testing.T) {
+	testPoly1, testPoly2 := produceTwoPriPoly(group, k, secret)
```
bford: As suggested above, merge this with the above function; you can reuse the 'testPoly' created above and just create a 'testPoly2' for the different-poly test.

poly/sharing_test.go
```diff
@@ -0,0 +1,544 @@
...
+	if testPoly1.Equal(testPoly2) {
+		t.Error("Polynomials are expected to be different.")
+	}
+}
+
+// Verify that Equal panics if the polynomials are of different degrees.
+func TestPriPolyEqual_Error1(t *testing.T) {
```
bford: Convert the _Error1 and _Error2 functions to a single function as discussed above as well, using an internal helper function to handle the 'defer' clause for each case.  For example, something like:

func TestPriPolyEqual_Error2(t *testing.T) {
        test := func(p1, p2 *PriPoly) {
   	      defer deferTest(t, "The Equal method should have panicked.")
              p1.Equal(p2)
        }

        // different degrees
	testPoly1 := producePriPoly(group, k, secret)
	testPoly2 := producePriPoly(group, k+10, secret)
	test(testPoly1, testPoly2)

        // different groups
	testPoly1 := producePriPoly(group, k, secret)
	testPoly2 := producePriPoly(altGroup, k, altSecret)
	test(testPoly1, testPoly2)
}

poly/sharing_test.go
```diff
@@ -0,0 +1,544 @@
...
+// Verify that Equal panics if the polynomials are of different groups.
+func TestPriPolyEqual_Error2(t *testing.T) {
+	defer deferTest(t, "The Equal method should have panicked.")
+	testPoly1 := producePriPoly(group, k, secret)
+	testPoly2 := producePriPoly(altGroup, k, altSecret)
+	testPoly1.Equal(testPoly2)
+}
+
+// Verify that the add function properly adds two polynomials together.
+func TestPriPolyAdd_Success(t *testing.T) {
```
bford: Combine the _Success, _Error1, _Error2 variants here and in other functions similarly to above.



Also, instead of re-creating a bunch of standard "test polynomials" in each of a bunch of test functions, you can just move those out to global variables in this _test.go module.  Those variables will only ever exist in versions of this package compiled for testing, so they won't pollute the namespace or create any binary code-bloat when this package is used "for real" by other packages.

poly/sharing_test.go
WEB3-GForce: DOne.

poly/sharing_test.go
WEB3-GForce: Done.

bford:
Hi William,

Once you've finished this round of changes, don't forget to 'gofmt -w' your code and check that the Travis CI build succeeds.

Thanks
Bryan

poly/sharing_test.go
WEB3-GForce: Done. All tests for a function are now contained in a single test function.

poly/sharing_test.go
WEB3-GForce: Done.

poly/sharing_test.go
WEB3-GForce: DOne.

poly/sharing_test.go
WEB3-GForce: I've created generic test polynomials and shares that can be used in tests that just need a standard polynomial. For tests that need to modify this polynomial, create a more customized polynomial, or create a pub/pri-share based on a specific polynomial, I kept the production function calls. I also kept the production calls in places where it would make the code easier to understand (such as when testing equality errors with polynomials of different groups or lengths).

WEB3-GForce:
Finished with the changes.

Issue #22, opened 2015-01-29
crypto/protobuf package needs to go away in favor of using dedis/protobuf


bford:
Danny @jackowitzd2 - I didn't notice until now that you had recently made some changes to the crypto/protobuf package, instead of to the newer top-level protobuf repo (https://github.com/DeDiS/protobuf), which we forked off of the crypto repo because at least one external user wanted to use it separately from the crypto library.  I had meant to delete the old crypto/protobuf package from the crypto library once the fork-off was complete, but forgot to do that.  So can you:

1. Propagate your recent bug fix into the protobuf repo if needed and make sure some (non-crypto-specific) test in that repo covers it, assuming Alec Thomas's work didn't already do this.

2. Move your crypto-specific Encoding test into, say, an addition to the TestSuite function in the crypto/test package, so that *all* suites will automatically get tested for functional Encoding support.

3. Create a temporary github branch of crypto and a Pull request announcing the impending deletion of the crypto/protobuf package and warning anyone else to move to using the top-level protobuf package.  @dyv and @iuliatamas might also currently have code that depends on it and needs to be updated, for example.

Thanks

jackowitzd2:
Bryan,

I'm pretty sure all changes I made here were backported from the protobuf fork - I'll verify that and then proceed as suggested.

Danny

jackowitzd2:
Related, do we want to consider moving away from the Encoding interface to the the more standard ```Binary(Un)Marshaller```?

jackowitzd2:
It also looks like e78c4e6 broke the encoding of ```poly.PubPoly``` since it no longer satisfies the Encoding interface (Missing ```EncodeTo``` and ```DecodeFrom```). I'll fix and add test cases for that as well.

bford:
It might indeed make sense to migrate toward compatibility with Binary(Un)Marshaler.  At the moment I'm thinking:

- Keep the Encoding interface in the crypto/abstract package, but rename it to Marshaling; it'll be a superset of BinaryMarshaler and BinaryUnmarshaler.

- Change Encode/Decode to the appropriate MarshalBinary/UnmarshalBinary signatures.

- Rename EncodeTo/DecodeFrom to MarshalTo/UnmarshalFrom.

- Rename Len to MarshalSize, for naming consistency with the Binary(Un)Marshaler interfaces and with the convention I've observed across most of the Go API to name such functions Size rather than Len (e.g., BlockSize etc).

- Drop String from the interface.

[...]

PR #26, opened 2015-01-29
Protobuf cleanse

jackowitzd2:
In response to #22, ```crypto/protobuf``` is going away in favor of the standalone package. This branch starts the migration process and anyone currently using protobuf should begin transitioning.


jackowitzd2:
This branch now has the ```crypto/protobuf``` package removed. I'll wait for @bford to determine when to actually roll it out.

bford:
Awesome, thanks Danny - I see that Neg in genPoint was indeed the source of the y-coordinate problem.

PR #34, opened 2015-02-12
Cipher API respin for "helper" methods

bford:
@davidiw (and anyone else) do you think this is a way to implement the Cipher API?  Basically makes Cipher a struct representing a very thin wrapper around a separate CipherState interface representing the underlying cipher state/implementation object.  This way (a) we can add helper functions as methods of Cipher, and (b) we can make the Cipher struct backward-compatible with the Reader, Writer, and cipher.Stream interfaces without widening the underlying CipherState interface.

(Not yet ready to merge; neither documented nor tested sufficiently.)


davidiw:
Bryan, if you're using VIM, might I recommend https://github.com/fatih/vim-go

It seems you're not running gofmt before committing. I just run this plugin within vim, and it yells at me if something is amiss and automatically gofmts as well.

bford:
Good idea, thanks - I'll check it out.  (But I didn't bother gofmt'ing this branch yet because I already explicitly stated it as being "not ready to merge" for other reasons as well. :) )

bford:
Since the only feedback I got on this seemed positive, I finished the documentation changes needed for this approach and plan to merge it in.  Any further issues before I do?

PR #36, opened 2015-02-21
added test that shows null point unmarshaling fails

iuliatamas:


iuliatamas:
I opened an issue about it, and the test fails like this:
nullPoint_test.go:19: invalid elliptic curve point

bford:
Hi Iulia, after a brief inspection of the source code to Go's IsOnCurve method, indeed it looks like the problem is that that function does not consider the neutral ("point-at-infinity") element to be "on the curve" for purposes of its return.  So could you:

1. apply something like this fix to your 'nullpoint' branch:

 func (p *curvePoint) Valid() bool {
-       return p.c.IsOnCurve(p.x, p.y)
+       // The IsOnCurve function in Go's elliptic curve package
+       // doesn't consider the point-at-infinity to be "on the curve"
+       return p.c.IsOnCurve(p.x, p.y) ||
+               (p.x.Sign() == 0 && p.y.Sign() == 0)
 }

2. gofmt -w the nist package (looks like it hasn't been)

3. add your null-point marshal/unmarshal test to the testGroup function in the crypto/test package.

4. Verify that the above test indeed fails on the P256 curve without the fix in #1 above.

Thanks
Bryan

bford:
Argh, excuse the crazy numbers - github's formatter does really, really bad things I guess when you try to cut-and-paste a diff into a text-entry box. :/

bford:
Aha, after a bunch of stupid trial-and-error, looks like the way to embed code in github markdown syntax is just use HTML <pre>...</pre> tags.  Nice to know. :)

dyv:
This is a good resource for formatting GitHub markdown: you can use 3 backticks in order to specify a code block, and specify the language of that code block as well.

https://help.github.com/articles/github-flavored-markdown/

iuliatamas:
After adding the test for marshaling/ unmarshaling null points in testGroup, the results:
 * with the nist Valid function unchanged: 
      * nist P256 fails
      * openssl P224 fails
      * suites test fails
 * with the Valid function corrected: 
      * openssl P224 fails

The openssl failure happens when binary marshaling:
```go
	b, _ := pzero.MarshalBinary()
```

bford:
Aha, the general test-case proves fruitful already. :)

I took a look at the internal implementation of EC_CURVE_point2oct in the openssl code, and it turns out that it special-case encodes the "point at infinity" as a *single* 0 byte, rather than the usual length.  Argh.  (Terrible idea in all sorts of ways.)

I believe that applying this patch to the openssl code should fix it (and keep the go crypto library's point-marshaling format constant size despite openssl's weirdness):

```
diff --git a/openssl/curve.go b/openssl/curve.go
index a406281..cafdf11 100644
--- a/openssl/curve.go
+++ b/openssl/curve.go
@@ -215,18 +215,27 @@ func (p *point) MarshalSize() int {
 func (p *point) MarshalBinary() ([]byte, error) {
        l := 1 + p.c.plen
        b := make([]byte, l)
+
+       // Note: EC_POINT_point2oct encodes the "point at infinity"
+       // as a single 0 byte, hence returning a length of 1.
        if C.EC_POINT_point2oct(p.c.g, p.p, C.POINT_CONVERSION_COMPRESSED,
                (*_Ctype_unsignedchar)(unsafe.Pointer(&b[0])),
-               C.size_t(l), p.c.ctx) != C.size_t(l) {
+               C.size_t(l), p.c.ctx) == 0 {
                panic("EC_POINT_point2oct: " + getErrString())
        }
+
        return b, nil
 }
 
 func (p *point) UnmarshalBinary(buf []byte) error {
+
+       l := len(buf)
+       if buf[0] == 0 { // Special case: point at infinity
+               l = 1 // single 0 byte
+       }
        if C.EC_POINT_oct2point(p.g, p.p,
                (*_Ctype_unsignedchar)(unsafe.Pointer(&buf[0])),
-               C.size_t(len(buf)), p.c.ctx) == 0 {
+               C.size_t(l), p.c.ctx) == 0 {
                return errors.New(getErrString())
        }
        return nil
```

iuliatamas:
Done

iuliatamas:
We will go ahead and merge this because it seems that this problem is solved and is holding up the evaluation.

bford:
Looks good, thanks Iulia.

PR #38, opened 2015-02-23
Cipher fix, cipher test, and a note...

davidiw:
I can break this up into individual pull requests if there are any issues with any of the commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,19 @@ of NIST-standardized elliptic curves and symmetric cryptosystems.
 [Stanford Pairing-Based Crypto (PBC) library](http://crypto.stanford.edu/pbc/),
 which will of course only work if the PBC library is installed.
 
+Issues
+------
+
+- Traditionally, ECDH (Elliptic curve Diffie-Hellman) derives the shared secret
+from the x point only. In this framework, you can either manually retrieve the
+value or use the MarshalBinary method to take the combined (x, y) value as the
+shared secret. We recommend the latter process for new softare/protocols using
+this framework as it is cleaner and generalizes across differen types of
+groups (e.g., both integer and elliptic curves), although it will likely be
+incompatible with other implementations of ECDH.
+http://en.wikipedia.org/wiki/Elliptic_curve_Diffie%E2%80%93Hellman
+
+
 Copyright (C) 2014 Yale DeDiS Group
 -----------------------------------
 
--- a/test/cipher.go
+++ b/test/cipher.go
@@ -1,11 +1,14 @@
 package test
 
 import (
+	"bytes"
+	"crypto/cipher"
+	"crypto/rand"
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/subtle"
 	"hash"
 	"testing"
-	"crypto/cipher"
 )
-
 
 func HashBench(b *testing.B, hash func() hash.Hash) {
 	b.SetBytes(1024 * 1024)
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -67,6 +67,16 @@ of NIST-standardized elliptic curves and symmetric cryptosystems.
 [Stanford Pairing-Based Crypto (PBC) library](http://crypto.stanford.edu/pbc/),
 which will of course only work if the PBC library is installed.
 
+Issues
+------
+
+- Traditionally, ECDH uses only the x point for the shared secret. In this
```
bford: This text is very unclear and out-of-the-blue; it needs some context, and in particular it's not obvious what exactly "uses" means here.  Can you expand the term "ECDH", and point to the most-relevant document specifying ECDH's standard behavior.  And you need to clarify here that "uses" means "forms the Diffie-Hellman shared secret from..."

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
-			c.XORKeyStream(data,data)
+			c.XORKeyStream(data, data)
 		}
 	}
 }
 
 // Benchmark a block cipher operating in counter mode.
+/*
+XXX Broken
 func BlockCipherBench(b *testing.B, keylen int,
```
bford: It looks like the Travis build is failing because this function is referenced but its definition is missing?

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+
+// Compares the bits between two arrays returning the fraction
+// of differences. If the two arrays are not of the same length
+// no comparison is made and a -1 is returned.
+func BitDiff(a, b []byte) float64 {
+	if len(a) != len(b) {
+		return -1
+	}
+
+	mask1 := byte(1)
```
bford: Argh, you can do better. :)  Don't unroll all this stuff 8 times here and in the for loop; just use nested loops with a bit of trivial arithmetic and the code size will be cut in half.

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+	mask3 := byte(4)
+	mask4 := byte(8)
+	mask5 := byte(16)
+	mask6 := byte(32)
+	mask7 := byte(64)
+	mask8 := byte(128)
+
+	count := 0
+	for i := 0; i < len(a); i++ {
+		if (a[i] & mask1) != (b[i] & mask1) {
```
bford: try if ((a[i] ^ b[i]) & (1 << j)) != 0 where j is the bit number from the inner loop

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+		if (a[i] & mask8) != (b[i] & mask8) {
+			count += 1
+		}
+	}
+
+	return float64(count) / float64(len(a)*8)
+}
+
+// Tests a Cipher can encrypt and decrypt
+func BCHelloWorldHelper(t *testing.T,
```
bford: Where does this name come from?  What is "BC"?

test/cipher.go
davidiw: I did it because I'm willing to bet $1 that this is faster albeit less elegant.

test/cipher.go
davidiw: Interesting, I'll take a second look, but from what I can recall, this code wasn't compiling.

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+		}
+	}
+}
+
+// Tests a Cipher:
+// 1) Encryption / decryption work
+// 2) Encryption / decryption with different key don't work
+// 3) Changing a bit in the ciphertext or mac results in failed mac check
+// 4) Different keys produce sufficiently random output
+func AuthenticateAndEncrypt(t *testing.T,
```
bford: "TestAuthenticatedEncryption"?

test/cipher.go
davidiw: BlockCipher

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+	hashsize := bc.HashSize()
+	mac := make([]byte, hashsize)
+
+	ncrypts := make([][]byte, n)
+	nkeys := make([][]byte, n)
+	nmacs := make([][]byte, n)
+
+	// Encrypt / decrypt / mac test
+	for i := range nkeys {
+		nkeys[i] = make([]byte, keysize)
```
bford: Looks like there's a bunch of key-setup code that's getting duplicated in each test; factor it out.

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+		}
+	}
+
+	deltacopy := make([]byte, cryptsize)
+
+	// Bit flipping test
+	for i := range ncrypts {
+		copy(ncrypts[i], deltacopy)
+
+		deltacopy[0] ^= 255
```
bford: A few comments please about what's going on from one code block to the next?

test/cipher.go
davidiw: I avoided starting anything with a Test name since it seems to be special
to Go's unit test. I suspect the signature would likely make sure that Go
didn't get confused, but someone reading the code might.

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+		bc = newCipher(nkeys[i])
+		bc.Message(decrypted, deltacopy, deltacopy)
+		bc.Message(mac, nmacs[i], nil)
+		if subtle.ConstantTimeAllEq(mac, 0) == 1 {
+			t.Log("MAC Check passed")
+			t.FailNow()
+		}
+		deltacopy[0] = ncrypts[i][0]
+
+		deltacopy[len(deltacopy)/2-1] ^= 255
```
bford: Looks like each of these blocks of code is mostly duplicate but testing a slightly different case; factor out the common code into a separate function.  Doesn't need to be a separate top-level function if an internal higher-order function def would be useful, i.e.,

```
deltatest := func (...) { ...}
```

very handy because you can also refer to internal variables in the outer test function from the inner higher-order function.

test/cipher.go
```diff
@@ -21,25 +24,274 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+	AuthenticateAndEncrypt(t, newCipher, n, bitdiff, kb)
+
+	mb := make([]byte, 2^20)
+	for i := 0; i < len(mb); i++ {
+		mb[i] = byte(i & 256)
+	}
+	AuthenticateAndEncrypt(t, newCipher, n, bitdiff, mb)
+}
 
+func BlockCipherTest(t *testing.T,
```
bford: Why "TestBlockCipher"?  These aren't block ciphers we're testing; we're testing the "abstract.Cipher" interface, which I've been calling a "Message Cipher" interface although that may not be the best name - but in any case the name "block cipher" is certainly not appropriate here.  For now just say TestCipher probably.

test/cipher.go
bford: Oh geesh, you're working about unrolling and micro-optimizing a function that's only used in a *test suite*???

test/cipher.go
bford: Indeed, I think Go will avoid getting confused due to the function signature, and people can probably take a hint from the function signature as well. :)  (But I've seen the practice of attaching "...Helper" to the name and stuff like that to clarify it further.)

test/cipher.go
davidiw: I'm not sure I follow the exact location of this comment. There are a handful of comments before each block and if the code isn't self-documenting, then it likely ought to be improved. Of course, it could be a third choice which is that regardless of how elegant the code is, it still needs more documenting :).

test/cipher.go
davidiw: I had thought about it when writing. I'm almost in favor of completely removing the Hello, World test, since it seems quite redundant anyway. But factoring out the code in this specific scenario seems like it may result in more work then we really want to invest. I.e., should we go as far to construct a struct that contains an array of keys, encrypted messages, macs, etc?

davidiw:
I've addressed most of your comments in the code and when not, I've left comments above. Let me know what you think. If you're happy, let me know, I'll squash and rebase my commits.

test/cipher.go
bford: Global variables are perfectly fine in a _test.go package.  Those don't get compiled into the non-test version of package so no need to worry about code bloat or startup performance due to them.

README.md
```diff
@@ -67,6 +67,17 @@ of NIST-standardized elliptic curves and symmetric cryptosystems.
 [Stanford Pairing-Based Crypto (PBC) library](http://crypto.stanford.edu/pbc/),
 which will of course only work if the PBC library is installed.
 
+Issues
+------
+
+- Traditionally, ECDH (Elliptic curve Diffie-Hellman) derives the shared secret
+from the x point only. In this framework, you can either manually retrieve
+the value or take the combined (x, y) value. The latter is the recommended
+process as it can be done without loss of security and done more cleanly,
```
bford: Avoid passive voice in technical writing!  (ALWAYS!!! :) )  "is the recommended process" - recommended by whom, and in what context?  And why can the latter be done more cleanly?  I would say:

```...or use the MarshalBinary method to take the combined (x, y) value as the shared secret.  We recommend the latter process for new software/protocols using this framework as it is cleaner and generalizes across different types of groups (e.g., both integer and elliptic-curve groups), ...```

File: test/cipher.go [hunk 12]
```diff
@@ -21,25 +24,230 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+func BitDiff(a, b []byte) float64 {
+	if len(a) != len(b) {
+		return -1
+	}
+
+	count := 0
+	for i := 0; i < len(a); i++ {
+		for j := 0; j < 8; j++ {
+			if ((a[i] ^ b[i]) & (1 << uint(j))) != 0 {
+				count += 1
```
bford: count++.

Or better yet, replace the whole if statement with:

```count += ((a[i] ^ b[i]) >> uint(j)) & 1```

File: test/cipher.go [hunk 13]
```diff
@@ -21,25 +24,230 @@ func HashBench(b *testing.B, hash func() hash.Hash) {
...
+		bc = newCipher(nkeys[index])
+		bc.Message(decrypted, testmsg, testmsg)
+		bc.Message(mac, testmac, nil)
+		if subtle.ConstantTimeAllEq(mac, 0) == 1 {
+			t.Log("MAC Check passed")
+			t.FailNow()
+		}
+	}
+
+	for i := range ncrypts {
```
bford: Some comments in here please?

test/cipher.go
davidiw: I understand the above comment, but I don't see the application in this context. The code sample has some redundancy with a far simpler test case in another function. We initialize variables each time, so that we can make the function generalizable to different ciphers, cleartext, and number of keys tested in parallel. Using global variables would not reduce the amount of initialization while keeping this generality.

test/cipher.go
davidiw: Sorry, I've used Python too much recently :)

test/cipher.go
bford: Sorry, maybe I'm being too nit-picky. :)  No big deal.

davidiw: After more careful examination, I realized I may have been misunderstanding you. You want the key generation to be done once in a separate init style function and store the results to global variables. I can easily implement that, but then we'd have a separate helper function that performs basically the following:

```
var keys [][]byte
func TestCipherInit(key_count int, key_size int) {
  keys = make([][]byte, key_count)
  for i := 0; i < key_count; i++ {
    keys[i] = make([]byte, key_size)
    rand.Read(keys[i])
  }
}
```

We'd still need to zero out the ncrypts and nmacs in each test case, since values are Xored with existing content and in not assigned. Perhaps I'm missing something, but I think this type of change would make the test code harder to read and it doesn't seem to be very critical to performance.

bford: Yes, that's something like what I had in mind - but use your judgment; if it doesn't feel like it would improve conciseness/clarity then don't bother with it.

davidiw: I feel mixed about it, because I had the same reaction as I was writing it and similar thoughts during our discussion. I tend to prefer test code that would guide users on how to use the code or that matches my intuition on good coding practice. So if we wanted to really invest the time into making this better, we ought to pass in the keys and other init variables as parameters into the function, but then I feel like I'm starting to do the same level of optimization as I did with the bit comparison :).

davidiw:
I've squashed the previous commits and these new (rebased / squashed) commits should address all concerns.

bford:
Sounds good, thanks David!

PR #40, opened 2015-02-27
Promises

WEB3-GForce:
@bford 
@jackowitzd2 

This is the branch in which I am working on the cryptographic object for the insurance policy. I opened up this pull request to serve as a forum of discussion while I work. Please feel free to make comments/ suggestions as I develop.


WEB3-GForce:
@bford 

I finished my first attempt at trying to write BinaryMarshal/ BinaryUnmarshal for a cryptographic object (PromiseSignature). Before I move on to writing the marshalling code for Promise, it would be helpful if you could offer suggestions on how I handled the PromiseSignature marshalling. Could you look over the code when you get a chance?

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+
+/* An initialization function for preparing a PromiseSignature for unmarshalling
+ *
+ * Arguments
+ *    s   = the signing suite
+ *
+ * Returns
+ *   An initialized PromiseSignature ready to be unmarshalled
+ */
+func (p *PromiseSignature) UnMarshalInit(suite abstract.Suite) *PromiseSignature {
```
bford: Unmarshal is typically spelled with only the U and not the M capitalized.

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ * of a given promise. After receiving a promise and verifying that their share
+ * is good, guardians can then produce a signature to send back to the promiser.
+ *
+ * Upon receiving this, the promiser can then add the signature to its lists of
+ * signatures to server as proof that the promiser has gained a sufficient
+ * number of guardians.
+ */
+type PromiseSignature struct {
+
+	// The index of the guardian producing the signature
```
bford: Index in what list/array?

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ * the original server goes offline.
+ *
+ * The Promise struct handles the logic of creating private shares, splitting
+ * these shares up for a given number of servers to act as guardians, verifying
+ * promises, and providing proof that guardians have indeed taken out approved
+ * of backing up the promse.
+ *
+ * Terms:
+ *   promiser = the server making the promise. The own who owns this object.
+ *   promised = recipients of the promise (aka the clients of the promiser)
```
bford: Let's just say "client" rather than "promised"

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ *   guardian = another server who receives a share of the promise and can help
```
bford: Let's say "insurer" here as we've been doing in other discussion.  (or "trustee" might also be OK.)

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ * Development note: The guardians, secrets, and signatures arrays should
+ *                   remain synchronized. In other words, the guardians[i],
+ *                   secrets[i], and signatures[i] should all refer to the same
+ *                   server.
+ */
+type Promise struct {
+
+	// The id of the promise. In the format:
+	//   PromiserPublicKey.String() + TimeOfCreation + RandomNumber
+	id string
```
bford: Why do you need to invent a new kind of ID string here?  What is it used for?

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+	
+	// The minimum number of shares needed before the policy can become
+	// active. t <= r
+	r int
+	
+	// The total number of shares to send.
+	n int
+	
+	// The public key of the promiser.
+	pubKey abstract.Point
```
bford: Is this the long-term public key of the promiser or the short-term public key whose private key gets secret-shared and potentially revealed if the promiser fails?  (The two need to be distinguished.)

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+	
+	// The public polynomial that is used to verify that a secret share
+	// given did indeed come from the appropriate private key.
+	pubPoly *poly.PubPoly
```
bford: Is there a reason this needs to be a pointer rather than just an inline struct-instance?

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+	
+	// Primarily for the promised, the number of shares that have been 
+	// revealed so far.
+	numShares int
```
bford: Looks like you're mixing "promise state" and "reconstruction state" in this struct - please separate the two.  Any specifically reconstruction-related state you need in the reconstruction process should be kept elsewhere.

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ *    guardians = a list of the public keys of servers to act as guardians.
+ *
+ * Returns
+ *   The initialized promise
+ *
+ * Note
+ *   Since shares will be multiplied by Diffie-Hellman keys, they need to be the
+ *   same group as the keys.
+ */
+func (p *Promise) Init(keyPair *config.KeyPair, t, r int,
```
bford: Does this KeyPair need to include the private key?  Is this Init function only supposed to be called by the promiser or also by nodes that receive and validate a promise?  If it's only for use by the promiser, it's better to name it accordingly to make that clear.

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ * the correct guardian. VerifyShare handles "semantic" verification.
+ *
+ * Arguments
+ *   i        = the index of the share
+ *   gKeyPair = the key pair of the guardian of the share
+ *
+ * Returns
+ *    Whether the share is valid or invalid.
+ */
+func (p *Promise) validShare(i int, gKeyPair *config.KeyPair) bool {
```
bford: Where is the actual validation of the share here?  I don't even see reference to an actual "share" in this function or its arguments, only a share index.

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ * Return
+ *   whether the decrypted secret properly passes the public polynomial.
+ *
+ * Note
+ *   Make sure that the proper index and key is specified. Otherwise, the
+ *   function will return false because diffieHellmanDecrypt gave the wrong
+ *   result. In short, make sure to verify only shares that are allotted to you.
+ */
+func (p *Promise) VerifyShare(i int, gKeyPair *config.KeyPair) bool {
+	if !p.validShare(i, gKeyPair) {
```
bford: Ah, I see - this is the real verify function.  But if this is the only place where validShare is called (or is ever likely to be), why the separate validShare function?  Do you anticipate there being other meaningful uses of validShare?

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+		return false
+	}
+	diffieBase := p.shareSuite.Point().Mul(p.pubKey, gKeyPair.Secret)
+	share := p.diffieHellmanDecrypt(p.secrets[i], diffieBase)
+	return p.pubPoly.Check(i, share)
+}
+
+/* Verifies whether a Promise objects has enough signatures to be valid.
```
bford: Let's not use the term "valid" to mean "enough signatures"; usually "valid" refers to something immediate and self-evident (like syntactically valid), not something that is acquired gradually over time.  Perhaps a Promise becomes "certified" once it has enough signatures?  Or some other similar term?

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ * Note:
+ *   The signature message will always be of the form:
+ *      Guardian approves PromiseId
+ *
+ *   It is assumed that the guardian has called VerifyShare first and hence
+ *   it is assumed that the input to the function is trusted.
+ */
+func (p *Promise) Sign(i int, gKeyPair *config.KeyPair) *PromiseSignature {
+	set        := anon.Set{gKeyPair.Public}
+	approveMsg := gKeyPair.Public.String() + " approves " + p.id
```
bford: Ah, now I see what 'id' is used for.  But I think the signature message should just be a binary message of some kind: that'll probably be more efficient and possibly more secure due to being fixed-length.

Also, the approval message doesn't need to *contain* the public key of the signer, since the signature will be *verified* using the public key of the signer anyway.  And by not making the approval message depend on the signer, the it'll be simpler and more efficient to verify multiple signatures on the same approval message (i.e., on the same promise): just one message hash, multiple signatures.

A further improvement we might want to consider is actually using multi-signatures here in similar fashion to what CoCo/CoSi does at larger scale, so that validating a certified promise requires only a single check of a composite signature.  Not critical and will require a bit of thought as to how to deal with the need for a "collective challenge", but I have some ideas...  we can get back to this laster.

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+/* Verifies a signature from a given guardian
+ *
+ * Arguments
+ *    sig = the PromiseSignature object containing the signature
+ *
+ * Return
+ *   whether or not the signature is valid
+ */
+func (p *Promise) VerifySignature(sig *PromiseSignature) bool {
+	if !p.validSignature(sig) {
```
bford: Again, this validSignature helper seems to be looking for a reason to exist.

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ *       found at the corresponding index in the Promise.
+ *    3. The secret provided passes the public polynomial
+ *
+ * Arguments
+ *    psecret = the PromiseShare to verify
+ *
+ * Return
+ *   Whether the secret is valid
+ */
+func (p *Promise) PromiseShareVerify(psecret *PromiseShare) bool {
```
bford: This should return an 'error' (or nil if no error) rather than a 'bool'.

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ *
+ *   1. The provided share when encrypted with the diffie key must equal the
+ *   point provided at index bi of the promise. This verifies that the share
+ *   was actually intended for the guardian.
+ *
+ *   2. The provided share must fail to pass pubPoly.Check. This ensures that
+ *   the share is actually corrupted and the guardian is not just lying.
+ */
+
+type PromiseShare struct {
```
bford: It's not yet clear to me exactly what role(s) this PromiseShare type is supposed to serve.  It doesn't like the Promise struct directly contains or refers to this type.  What exactly is it for?

poly/promise/promise.go
```diff
@@ -0,0 +1,744 @@
...
+ *
+ * Arguments
+ *    i   = the index of the Promise share the guardian is revealing
+ *    s   = the share being revealed
+ *    d   = the Diffie-Hellman key between the guardian and promiser
+ *
+ * Returns
+ *   An initialized PromiseShare
+ */
+func (p *PromiseShare) Init(i int, s abstract.Secret, d abstract.Point) *PromiseShare {
```
bford: Aha, now I see - apparently 'PromiseShare' is only used/needed in the reveal process, hence it's really part of the reveal/reconstruction state, correct?  In that case I would eliminate the PromiseShare type in favor of a Reveal struct or something like that representing *all* the state needed for a complete reveal process: e.g., including the reconstruction-related fields currently in the Promise struct, as well as a slice of shares that have been validated and revealed so far.  Once this refactoring is done, I don't think any equivalent of the PromiseShare struct will be needed at all, because the 'i' will be implicit in the position of the revealed share in the slice of revealed shares, and the DH key will only be needed ephemerally during the actual share-decryption process and won't need to be stored once the share is decrypted and checked.  Make sense?

bford:
Good start, see all my inline comments. :)

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: Per our discussion, I've started separating the two. I'll refine it and polish it up a little more.

poly/promise/promise.go
WEB3-GForce: Made the changes. I was initially thinking about combining the two functions but hadn't gotten around to it quite yet.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: This was primarily an artifact of an initial design. I was planning on merging the two and did so in a recent commit.

poly/promise/promise.go
WEB3-GForce: As in later commits, this is taken care of.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: I removed the index (which must now be tracked by the caller). In short, PromiseSignature is a simple struct that makes keeping track of signatures in BlameProofs and PromiseState easier. It handles the marshalling/ unmarshalling of the signature itself, which helps make other code simpler.

poly/promise/promise.go
WEB3-GForce: Initially, I was thinking this could help support multiple promises per server. The promise id could be used to differentiate them and as a hash key to allow quick look-ups. However, there are other ways of differentiating promises (namely the new Equal method I created). Furthermore, fast look-ups probably won't be much of a concern as a given server is unlikely to have too many active Promises at once.

poly/promise/promise.go
WEB3-GForce: I changed this to be the long term key.

poly/promise/promise.go
WEB3-GForce: Not particularly. I changed it to be inline.

poly/promise/promise.go
WEB3-GForce: Perhaps, a better name for this function would be ConstructPromise.

The function is used to create a new promise. The function in the the latest version of the code accepts two keypairs: the secret keypair which contains the secret to be shared (its suite and private key are needed) along with a long term keypair containing the long term credentials of the server (which are used for the Diffie-Hellman encryption).

poly/promise/promise.go
WEB3-GForce: I changed the signing to use predefined messages.

My thought behind using the more specific messages was to make it harder for malicious signature reuse. A malicious user who wanted to frame a server could produce a false BlameProof message. It could then get the victim to produce a valid signature for a valid share it has produced. It could then take that signature and attach it to the BlameProof. Hence, it could pretend that the server had unjustly blamed it. Having fine grain messages would prevent a malicious server from reusing signatures in such fashion since the malicious server would need to get the victim to sign a message saying "publicKey blames publicKey" instead of "publicKey approves publicKey".

Do you think this will be a concern?

However, the approach above wouldn't work well if we wanted to do collective signing as I can imagine the message soon growing to an unwieldy size.

poly/promise/promise.go
WEB3-GForce: I've changed all such methods to produce errors instead.

poly/promise/promise.go
WEB3-GForce: Please see comments below.

poly/promise/promise.go
WEB3-GForce: I think the issue is that I am combining network logic with crypto logic.

PromiseShare is more of a networking struct, since insurers will have to send the secret and the index over the network to clients who ask. I should move this code to prifi.

I've reworked promise.go to contain these structures:

    - PromiseState => Keeps track of signatures certifying the promise, secrets revealed,    
    and blame proofs received
 
    - PromiseSignature => a struct responsible for marshalling/ unmarshalling signature
    messages (along with keeping track of their suites). This makes it easier for
    PromiseState to keep track of signatures.

    - BlameProof => an object containing enough information to prove that a promiser
    created an invalid secret share

WEB3-GForce:
@bford 

I've finished with the next iteration of the code, particularly

    - Responding to your comments
    - Implementing a more robust blame system
    - Providing more extensive documentation
    - Cleaning up code

The code is ready for another review and hopefully a merge soon.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/anon"
+	"github.com/dedis/crypto/config"
+	"github.com/dedis/crypto/poly"
+	"github.com/dedis/crypto/proof"
+	"github.com/dedis/crypto/random"
+)
+
+/* This package implements the Promise cryptographic primitive, which is based
```
bford: Make sure this overview text gets included in a sensible way in godoc-generated documentation.  It might be appropriate for the "package-level" documentation, for example, which should go just above the 'package' statement if so.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ * will make more sense.
+ *
+ * Code using this package will typically have the following flow (please see
+ * "Key Terms" below for a definition of terms used):
+ *
+ * 1) The promiser constructs a new Promise and stores it withing a PromiseState
+ *
+ * 2) The promiser sends the Promise to the insurers
+ *
+ * 3) The insurers verify the Promise is well-formed and make sure that their
```
bford: Combine steps 3,4,5 into 3 with sub-bullets 3a and 3b.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *    secret shares are valid.
+ *
+ * 4) If the secret share is invalid, an insurer creates a BlameProof and sends
+ *    it back.
+ *
+ * 5) If the share is valid, an insurer creates a PromiseSignature to send back
+ *    to the promiser.
+ *
+ * 6) The promiser receives the message from the insurer.
```
bford: similarly for 6,7,8?

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ * 14) The insurers can then check if the promiser is indeed unresponsive.
+ *
+ * 15) If so, an insurer reveal its share and send it to the client.
+ *
+ * 16) The client repeats steps 13-15 until enough shares are recovered to
+ *     reconstruct the secret.
+ *
+ * 17) The client can then reconstruct the secret and take over for the promiser
+ *
+ *
```
bford: In general, you need more structure above: any list of 17 steps is by definition pretty much impenetrable to a reader.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+
+/* Promise structs are mechanisms by which a server can promise other servers
+ * that an abstract.Secret will be availble even if the secret's owner goes
+ * down. The secret to be promised will be sharded into shared secrets that can
+ * be combined using Lagrange Interpolation to produce the original secret.
+ * Other servers will act as insurers maintaining a share. If a client ever
+ * needs the secret to be reconstructed, it can contact the insurers to regain
+ * the shares and reconstruct the secret.
+ *
+ * Note to Developers:
```
bford: The text "note to developers" is redundant since it is only developers who tend to read API documentation like this.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+	p.pubKey = longPair.Public
+	p.insurers = insurers
+	p.secrets = make([]abstract.Secret, p.n, p.n)
+
+	// Verify that t <= r <= n
+	if p.n < p.t {
+		panic("Not enough insurers for the secret")
+	}
+	if p.r < p.t {
+		p.r = p.t
```
bford: I think all three invariant checks should just panic on failure; that's an API demand and shouldn't be violated by the caller.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+/* Given a Diffie-Hellman shared secret, encrypts a secret.
+ *
+ * Arguments
+ *    secret      = the secret to encrypt
+ *    diffieBase  = the DH shared secret
+ *
+ * Return
+ *   the encrypted secret
+ */
+func (p *Promise) diffieHellmanEncrypt(secret abstract.Secret,
```
bford: How about just having this function return the diffieSecret without doing an Add.  Then you can eliminate diffieHellmanDecrypt, use this one function for both encryption and decryption, and do the Add or Sub in the caller.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *   * VerifyBlame
+ */
+type Promise struct {
+
+	// The id of the promise used to differentiate it from others
+	// The id is the short term public key of the private key being promised
+	id string
+
+	// The cryptographic suite to use for the shared secrets
+	shareSuite abstract.Suite
```
bford: Can shareSuite and keySuite ever be different?  I don't think so, because the secret-shared Secret needs to be usable and thus in an appropriate modulus for both secret-sharing and DH key exchange with the long-term keys.  I think you could probably just have a single 'suite' config variable here.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+	pripoly := new(poly.PriPoly).Pick(p.shareSuite, p.t,
+		secretPair.Secret, random.Stream)
+	prishares := new(poly.PriShares).Split(pripoly, p.n)
+	p.pubPoly = poly.PubPoly{}
+	p.pubPoly.Commit(pripoly, nil)
+
+	// Populate the secrets array with the shares encrypted by a Diffie-
+	// Hellman shared secret between the promiser and appropriate insurer
+	for i := 0; i < p.n; i++ {
+		diffieBase := p.keySuite.Point().Mul(insurers[i], longPair.Secret)
```
bford: Here is (I think) why there can only be one suite: longPair.Secret gets used both in secret-sharing and in DH key agreement.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *    secretPair   = the keypair of the secret to be promised
+ *    longPair     = the long term keypair of the promiser
+ *    t            = minimum number of shares needed to reconstruct the secret.
+ *    r            = minimum PromiseSignatures needed to certify the Promise
+ *    insurers     = a list of the long-term public keys of the insurers.
+ *
+ * Returns
+ *   A newly constructed Promise
+ */
+func (p *Promise) ConstructPromise(secretPair *config.KeyPair,
```
bford: I think there should be only one constructor/init function, such as this, which is called both on the promiser node and on other nodes that will receive and unmarshal a promise.  All other nodes should also call this Init function first with valid suite, t, r, and insurers; let's call those fields "configuration information" that all nodes should know and agree on a priori.  Really the *only* difference in initialization here should be that the promiser includes the secret to be secret-shared, whereas on other nodes that secret can be 'nil' (and in that case this function will skip the secret-sharing part).

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+/* Initializes a Promise for unmarshalling
+ *
+ * Arguments
+ *    shareSuite = the suite used for shared secrets within the Promise
+ *    keySuite   = the suite used for long term keys within the Promise
+ *
+ * Returns
+ *   An initialized Promise ready to be unmarshalled
+ */
+func (p *Promise) UnmarshalInit(shareSuite, keySuite abstract.Suite) *Promise {
```
bford: This should be unnecessary once the Init function above is generalized.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *    promiserKey = the long term key the caller believes the Promise to be from
+ *
+ * Return
+ *   an error if the promise is malformed, nil otherwise.
+ *
+ * TODO Consider more ways to verify (such as making sure there are no duplicate
+ *      keys in p.insurers or that the promiser's long term public key is not in
+ *      p.insurers).
+ */
+func (p *Promise) VerifyPromise(promiserKey abstract.Point) error {
```
bford: The UnmarshalBinary function should call this automatically and fail if a received promise is invalid.  If that's happening, then maybe this can just be an internal helper and no longer needs to be a public method?

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+	}
+	// There should be a secret and public key for each of the n insurers.
+	if len(p.insurers) != p.n || len(p.secrets) != p.n {
+		return errors.New("Insurers and secrets array should be of length promise.n")
+	}
+	return nil
+}
+
+// Returns the id of the Promise
+func (p *Promise) GetId() string {
```
bford: Go practice names getter methods simply Id() instead of GetId().

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *    i         = the index of the share to verify
+ *    gKeyPair  = the long term key pair of the insurer of share i
+ *
+ * Return
+ *  an error if the share is malformed, nil otherwise.
+ *
+ * Note
+ *   Make sure that the proper index and key is specified.
+ */
+func (p *Promise) VerifyShare(i int, gKeyPair *config.KeyPair) error {
```
bford: Perhaps this should be implicitly done as part of the UnmarshalBinary method for shares, and hence does not need to be a public method?

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ * clients or others who request an insurer to reveal its shared secret.
+ *
+ * Arguments
+ *    i     = the index of the share
+ *    share = the share to validate.
+ *
+ * Return
+ *   Whether the secret is valid
+ */
+func (p *Promise) VerifyRevealedShare(i int, share abstract.Secret) error {
```
bford: An insurer should verify the share it receives *before* it signs a received Promise; the API should guarantee this sequence of events.  RevealShare should only ever get called on a Promise object that's been signed, and hence has been verified to be valid.  Thus, RevealShare should only ever return a valid share (or else panic if it got called out-of-sequence or on a Promise that failed to verify).  Hence, there should be no need for a public VerifyRevealedShare method.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *
+ * Arguments
+ *    i         = the index of the malicious shared secret
+ *    gKeyPair  = the long term key pair of the insurer of share i
+ *
+ * Return
+ *   A BlameProof that the promiser is malicious or nil if an error occurs
+ *   An error object denoting the status of the proof construction
+ */
+func (p *Promise) Blame(i int, gKeyPair *config.KeyPair) (*BlameProof, error) {
```
bford: Similarly, there should be no need for public Blame or VerifyBlameProof methods - they should probably be private.  The only public method should produce a "PromiseResponse" struct or some such that can be either a signature (if a received-and-unmarshalled promise was valid) or a blame proof (if it wasn't).

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *
+ * Returns
+ *   The error status of the unmarshalling (nil if no error)
+ */
+func (p *Promise) UnmarshalBinary(buf []byte) error {
+	pointLen := p.keySuite.PointLen()
+	secretLen := p.shareSuite.SecretLen()
+
+	// Decode n, r, t
+	p.n = int(binary.LittleEndian.Uint32(buf))
```
bford: no need to marshal/unmarshal n, r, t: as per above, they should be well-known config parameters that receiving nodes can be expected to "know" up-front.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *
+ * Arguments
+ *    r = the reader to use for unmarshalling
+ *
+ * Returns
+ *   The number of bytes read
+ *   The error status of the read (nil if no errors)
+ */
+func (p *Promise) UnmarshalFrom(r io.Reader) (int, error) {
+	// Retrieve p.n and p.t and then initialize p.PubPoly
```
bford: This should probably be just a trivial wrapper around UnmarshalBinary(); this much (duplicated?) code shouldn't be needed.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *    To reconstruct the secret, do:
+ *
+ *       p.PriShares.Secret()
+ *
+ *    Be warned that Secret will panic unless there are enough shares to
+ *    reconstruct the secret. (See poly/sharing.go for more info)
+ *
+ * TODO Consider if it is worth adding a String function
+ */
+type PromiseState struct {
```
bford: Since this is in a 'promise' package, just name this 'State' instead of 'PromiseState': otherwise external code will be referring to this as 'promise.PromiseState' which is long and redundant.  Same with all the other Promise-prefixed public types etc.

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+ *    i   = the index in the signature array this signature belongs
+ *    sig = the PromiseSignature to add
+ *
+ * Postcondition
+ *   The signature has been added
+ *
+ * Note to users of this code
+ *   Be sure to call ps.Promise.VerifySignature before calling this function
+ */
+func (ps *PromiseState) AddSignature(i int, sig *PromiseSignature) {
```
bford: As per above, there should only be one 'AddSignature' or 'AddResponse' or some such, for a single public type that can represent either a signature (attesting that the promise was good) or a blame-proof (attesting that the promise was bad).

poly/promise/promise.go
```diff
@@ -0,0 +1,1355 @@
...
+
+	// The suite used throughout the BlameProof
+	suite abstract.Suite
+
+	// The Diffie-Hellman shared secret between the insurer and promiser
+	diffieKey abstract.Point
+
+	// A HashProve proof that the insurer properly constructed the Diffie-
+	// Hellman shared secret
+	diffieKeyProof []byte
```
bford: you can just call this 'proof'

poly/promise/promise.go
WEB3-GForce: I moved this right above the "package promise" statement.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: I divided the list into 5 main steps with each step have ~3-4 subpoints.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: Done

poly/promise/promise.go
WEB3-GForce: I was thinking that the long term and the short term keys might be of different suites. For example, the long term key may be more cryptographically secure but slower since it will be used infrequently while the short term might be less secure but faster since it will be used more frequently once insurance is up. However, it would make the protocol much simpler just to have one suite.

I made the change to only have a single suite.

poly/promise/promise.go
WEB3-GForce: I made the change. Thanks! It made the code much simpler.

poly/promise/promise.go
WEB3-GForce: My initial thought on having ConstructPromise and UnmarshalInit be two different functions was to follow the model of sharing.go (an example is included at the bottom of the comment).

I was also hoping to make the code generalizable for future iterations. With the insurers list for example, it would be difficult for all nodes to know which insurers an individual promiser selected. This would likely differ from server to server since we would gain more security overall if each server had a different set of insurers. Hence, the nodes would need to wait to unmarshal the promise to figure out its insurers.

Concerning the longPair, an insurer would also have to remember that longPair should not be its own long term key pair but the pair from the server who owned the Promise they were trying to unmarshal.

Since what you would input for secretPair, longPair, and insurers would differ depending on whether you were a promiser taking out your own secret or an insurer unmarshalling a promise, I thought it might be easier for future users of this code base to have two different functions.

    // Initialize to an empty polynomial for a given group and threshold (degree),
    // typically before using Decode() to fill in from a received message.
    func (pub *PubPoly) Init(g abstract.Group, k int, b abstract.Point) {  
  	    pub.g = g
   	    pub.b = b
	    pub.p = make([]abstract.Point, k)
    }

    // Initialize to a public commitment to a given private polynomial.
    // Create commitments as encryptions of a given base point b,
    // or the standard base if b == nil.
    func (pub *PubPoly) Commit(pri *PriPoly, b abstract.Point) *PubPoly {
	    g := pri.g
	    k := len(pri.s)
	    p := make([]abstract.Point, k)
	    for i := 0; i < k; i++ {
		    p[i] = g.Point().Mul(b, pri.s[i])
	    }
   	    pub.g = g
	    pub.b = b
   	    pub.p = p
	    return pub
    }

poly/promise/promise.go
WEB3-GForce: Please see my comments above for a more general explanation of this method.

From an unmarshalling perspective, the only thing that is needed to unmarshal the Promise is the suite since everything else is sent over the wires.

However, it might be helpful to give t, r, and n to the function for verification purposes. These could be used to denote the minimum values this server is willing to tolerate.

 If the buffer contains values less than this, it would fail to marshal. If it contained values greater or equal to this (hence providing the expected amount of redundancy or more), the unmarshalling would succeed (assuming everything else is okay).

Does that sound like a good idea?

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: It would be more difficult to make this private.

VerifyShare needs to know the index and keypair of the insurer who was unmarshalling. I would need to store this information within the Promise since I couldn't alter the marshalling interface to include them. However, I thought that these two would be better suited as arguments versus struct variables.
 
Different actions will be taken depending upon whether or not this function succeeds. If it fails, then the share and/or Promise was improperly constructed (likely maliciously). Hence, the insurer could construct a BlameProof to show that the Promise is untrustworthy at best or malicious at worst.
It would be more difficult to make this private.

VerifyShare needs to know the index and keypair of the insurer who was unmarshalling. I would need to store this information within the Promise since I couldn't alter the marshalling interface to include them. However, I thought that these two would be better suited as arguments versus struct variables.
 
Different actions will be taken depending upon whether or not this function succeeds. If it fails, then the share and/or Promise was improperly constructed. Hence, the insurer could construct a BlameProof to show that the Promise is untrustworthy at best or malicious at worst. If it succeeds, then the Promise is good and a PromiseSignature can be constructed.

poly/promise/promise.go
WEB3-GForce: VerifyRevealedShare is used to protect against malicious insurers. Although the Promise could be valid, a malicious insurer could give a client a bad share. Hence, it would not be able to recover the secret.

The function provides a way for clients to ensure that an insurer has given it the proper share.

poly/promise/promise.go
WEB3-GForce: In my revisions of creating a PromiseResponse, I no longer need to make BlameProof public. The name "proof" would conflict with crypto/proof, so I named the struct blameProof.

However, I changed diffieKeyProof  to proof.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: DOne.

poly/promise/promise.go
WEB3-GForce: Done.

poly/promise/promise.go
WEB3-GForce: Done. Made the marshalling code much simpler.

poly/promise/promise.go
WEB3-GForce: I initially was not treating n or t as configuration parameters. Hence, the insurers array and public polynomial could be of variable length. I needed to find p.n and p.t before I could determine the buffer size. Now that these are configuration parameters, I can just call MarshalSize directly.

poly/promise/promise.go
WEB3-GForce: However, I made sure that good insurers won't accidentally reveal a bad share. RevealShare will panic unless the Promise is certified. 

Since the State struct keeps track of Response's to determine whether a Promise is certified, I made Promise.revealShare an internal function and State.RevealShare a public wrapper.

poly/promise/promise.go
WEB3-GForce: Update: in response to your latter comments on creating a Response struct, I created a new method Promise.ProduceResponse. This method produces a new response.

I wait to verify the response until State.PromiseCertified is called. Any invalid responses are simply ignored. Hence, I made verifyShare private.

poly/promise/promise.go
WEB3-GForce: I also created a Response struct that is a union of PromiseSignature and BlameProof (now the private structs signature and blameProof). This makes sure that a Response is only created after verifyShare is called and that the proper response is made based on whether the share is valid or invalid.

WEB3-GForce:
@bford 

I responded to the changes you mentioned and asked some additional questions. I've altered the code to produce a new version. Could you look over them when you get a chance and let me know if you would like me to change the code more?

I'll proofread the documentation again after I've made all the changes you like

WEB3-GForce:
@bford 
@davidiw 

I've added a few more features to Promise so that it could support the networking code properly. It now has all the features to support the first networking version.

The code has been tested and ready for merge after any additional reviews that you would like to do.

bford:
William, just noticed - your comment-writing style is incompatible with godoc and produces all kinds of bad effects.  See for example:

http://godoc.org/github.com/DeDiS/crypto/poly/promise

Can you fix this?  You can start your own godoc server locally via:

'''godoc -http=:6060'''

and edit your code-comments appropriately until the resulting godoc output looks decent.  Thanks!

B

PR #44, opened 2015-09-14
Shamir

nikkolasg:
Implementing a Joint Threshold signature :
 - create a Shared secret between M peers 
 - using the shared secrets protocols to implement a distributed schnorr signature


bford:
Thanks Nicolas.  I see this is in-progress so won't comment for now, except to bring up one more question to think about:  is the term "promise" the right name for what the "promise" code does, or is there a better name?  It was intended to represent a polynomial and corresponding set of encrypted secret-shares that some dealer "promises" are good and correspond to each other; whether they actually are good can of course only be verified by the individual share-receivers, who can validate or refute them in the response.  But perhaps "promise" is a bit vague a term and not the most closely related to the function the Promise data structure is serving.

A couple possible alternatives that came to mind are "Sharing" (as an, "a sharing" by a polynomial-dealer), or "Dealing" or just "Deal" (as in a polynomial and set of secret shares that have been "dealt" by a "dealer"; I think this is at least somewhat consistent with verifiable secret sharing terminology).  Thoughts?

nikkolasg:
Indeed we can find à more appropriate name. Deal is fine for me, because I have named my structs for the threshold as 'Dealer' and 'Receiver' so a 'Dealer' will now send M 'deals' to 'Receivers' which makes the whole thing rather explicit .

bford:
Stylistic nit: can we standardize on 80-column lines, at least for comments and when there isn't otherwise a good reason lines need to be longer?  Thanks.

poly/helper.go
```diff
@@ -0,0 +1,107 @@
+package poly
```
bford: Is this code mostly or only used by _test.go test-suite code?  If so, it should also have a _test.go name, so that it doesn't get compiled into the non-test "deployment" code-base.  For example, 'help_test.go'?

poly/joint_test.go
```diff
@@ -0,0 +1,99 @@
+package poly
+
+import (
+	"fmt"
+	"testing"
+)
+
+/////// TESTING ///////
+
+func TestReceiverAddDealer(t *testing.T) {
```
bford: Tests specifically related to a particular type, e.g., Dealer, should be called 'TestDealer...' with the type name first after Test, so that we can 'go test -run Dealer*' to run only the tests related to Dealer.

util/encoding.go
```diff
@@ -0,0 +1,31 @@
+package util
+
+import (
+	"bytes"
+	"encoding/binary"
+)
+
+// Helpers functions regarding the encoding of basic types ( integers ...)
```
bford: Is there a reason 'binary.LittleEndian.PutUint64', 'binary.LittleEndian.Uint64', etc won't work instead of these?

poly/helper.go
nikkolasg: Done. helper_test.go

poly/joint_test.go
nikkolasg: This function runs the test for the Receiver struct for its method "AddDealer". So that should follow the guidelines right ?

util/encoding.go
nikkolasg: Indeed i've tried with uint => UInt64.
The reason i've fallen back to Int is for "general usage" & compatibility. For example, in promise.go, you must add a Response at a specific index i which is an int. I could write another method with UInt, but then i'd had to do a lot of cast to operate with the rest of the code. 
I agree that using uint is much better and that was my first approach, but since the rest of the library only uses int, it becomes more of a burden to switch to uint that to stay with int.
What do you think ?

poly/joint_test.go
bford: Oops, yes - I should have put this note on a different function, e.g., TestRightDealer... below. :)

util/encoding.go
bford: When encoding/decoding, especially in size-sensitive binary formats, I think it's best to use only the size-specific types (e.g., int64 or uint64) to make it abundantly clear from the code exactly what the size and hence encoding is.  For Varint and other variable-length encodings it might make more sense to use encode/decode functions that just take 'int' types.

nikkolasg: Would it be okay then if I implement methods like "UInt64ToBuf" and "BufToUInt64", but still stay with the same "int" in my structs like in PartialSchnorrSig.Index ? I would just transform the Index into a uint64 before passing it to UInt64ToBuf. That way, the binary format would know we use a "uint64" but the structs still use the same good old "int"

poly/schnorr.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// Init is needed before decoding a PartialSchnorrSig so it can now about the length of a secret
+func (pss *PartialSchnorrSig) Init(suite abstract.Suite) *PartialSchnorrSig {
+	pss.Suite = suite
+	return pss
+}
+
+// Marshal a PartialSchnorrSig into a packet
+func (pss *PartialSchnorrSig) MarshalBinary() ([]byte, error) {
```
bford: Are you sure you even need these custom MarshalBinary/UnmarshalBinary methods? At least for PartialSchnorrSig, it looks to me like the generic abstract.Read and abstract.Write functions should work fine (after eliminating the abstract.Suite element from PartialSchnorrSig, which your comment says was only needed for marshal/unmarshal anyway).

poly/joint.go
```diff
@@ -0,0 +1,286 @@
...
+	return util.IntSize(p.N) +
+		util.IntSize(p.R) +
+		util.IntSize(p.T)
+}
+
+func (p *PolyInfo) Equal(p2 PolyInfo) bool {
+	return p.N == p2.N && p.R == p2.R && p.T == p2.T && p.Suite.String() == p2.Suite.String()
+}
+
+func (p *PolyInfo) MarshalBinary() ([]byte, error) {
```
bford: PolyInfo also looks like a type that shouldn't need its own custom marshal/unmarshal functions (similarly after eliminating the abstract.Suite pointer from it).  Note that abstract.Read and abstract.Write takes a separate Group interface parameter, which is the essential part of Suite for the purpose of marshaling/unmarshaling Points and Secrets.  I want to generalize it further at some time, but it should already be adequate for these purposes at the moment.

bford: And if you do for some reason want/need to implement your own MarshalBinary function here, it would be cleaner and more efficient to do it something like:

var buf bytes.Buffer
var val [3]uint64 = {(uint64)p.T, (uint64)p.R, (uint64)p.N}
err := abstract.Write(&buf, &val, suite)
...

bford: I've been meaning to change abstract.Read and abstract.Write slightly as well to take a variable-length list of obj interface{} which would/will make it even more convenient; i.e., then barring int type issues you could just

var buf bytes.Buffer
err := abstract.Write(&buf, suite, &p.T, &p.R, &p.N)

util/encoding.go
bford: Yes but I don't see the utility in having helper functions like this that return fresh byte-arrays; it seems both more efficient and just as clean and simple either to (a) just store directly into a caller-supplied byte array as LittleEndian.PutUint64 does, if the caller wants to handle layout directly; or (b) use something like abstract.Read/abstract.Write instead, which handles the layout itself into a Buffer or other stream object.

poly/schnorr.go
nikkolasg: Indeed you are right. I didn't know about abstract/encoding.go and just understood a lot better your abstract/ folder. I've been trying to make a little example out of the abstract.Read / Write method to see if they work like I expect them to.... But unfortunatly I can not have a working example.  Would you be willing to see what's wrong in this example ?
```Go
package main

import (
	"bytes"
	"fmt"
	"github.com/dedis/crypto/abstract"
	"github.com/dedis/crypto/edwards"
)

var suite = edwards.NewAES128SHA256Ed25519(true)

type CustomStruct struct {
	Index int32
	//Secret abstract.Secret
	Index2 int32
}

func main() {
	testReadWrite()
}

func (cs *CustomStruct) Equal(cs2 *CustomStruct) bool {
	if cs.Index != cs2.Index {
		fmt.Printf("Index 1 is not equal\n")
		return false
	}
	if cs.Index2 != cs2.Index2 {
		fmt.Printf("Index 2 is not equal\n")
		return false
	}
	//&& cs.Secret.Equal(cs2.Secret)
	return true
}

func testReadWrite() {
	//sec := suite.Secret().Zero()
	cs := CustomStruct{
		Index: 4,
		//Secret: sec
		Index2: -4,
	}
	var buf *bytes.Buffer = new(bytes.Buffer)
	err := abstract.Write(buf, cs, suite)
	if err != nil {
		fmt.Printf("Hoho error Writing : %v\n", err)
	}

	bytesBuffer := buf.Bytes()

	buf2 := bytes.NewBuffer(bytesBuffer)
	cs2 := CustomStruct{}
	err = abstract.Read(buf2, cs2, suite)
	if err != nil {
		fmt.Printf("Hoho error Reading : %v\n", err)
	}

	if cs.Equal(&cs2) {
		fmt.Printf("YOUPI\n")
	} else {
		fmt.Printf("NOOOOAAAAA\n")
	}
}
```

I've tried with simple "int", then "int32", then removing the "secret" but nothing works ><. 
I am willing to use the abstract.Read / Write if I understand them correctly though. In the meantime, I think the priority like we said is in the comparison baseline.

poly/joint_test.go
nikkolasg: Changed to TestDealer....

poly/joint.go
nikkolasg: Indeed this is a nice addition to have, with the inclusion of that method into the Group itself so we can get rid of it in the argument list.
However, as I've done some tests, it seems it doesn't work with  basic integers type, but only for Secrets or points or others abstract structures. (only tested with secrets)
If we can get it working one way or the other, **ALL** my structs are indeed already abstract.Write-able / Read-able, since every fields can be written by abstract.

Daeinar: Hi,

ah nice it seems you already figured out the problem regarding abstract.Read/Write. I also looked into it a little bit earlier today but then ran out of time to post my findings. In any case, I came to the same conclusion that abstract.Read/Write seems not to be working with basic types. See examples below. 

Furthermore, it seems to be a known approach when using the encoding/binary package that you have to Write/Read all the elements of a struct step-by-step. See left side of: https://bochs.info/p/cxvm9

Cheers,
Philipp

```go
package main

import (
    "bytes"
    "fmt"
    "reflect"
    "github.com/dedis/crypto/abstract"
    "github.com/dedis/crypto/edwards"
)

var suite = edwards.NewAES128SHA256Ed25519(true)

type Struct01 struct {
    Secret abstract.Secret
    Point abstract.Point
}

// this test passes since the to-be-(un)packed struct has only abstract.Secret
// and abstract.Point elements
func test01() {

    rand := suite.Cipher([]byte("example"))
    x := suite.Secret().Pick(rand)
    y := suite.Point().Mul(nil, x)

    cs := Struct01{
        Secret: x,
        Point: y,
    }

    buf := new(bytes.Buffer)
    err := abstract.Write(buf, cs, suite)
    if err != nil {
        fmt.Printf("Error Writing : %v\n", err)
    }

    buf2 := bytes.NewBuffer(buf.Bytes())
    cs2 := Struct01{}
    err = abstract.Read(buf2, &cs2, suite)
    if err != nil {
        fmt.Printf("Error Reading : %v\n", err)
    }

    fmt.Printf("Test01: ")
    if reflect.DeepEqual(cs,cs2) {
        fmt.Print("Hurray!\n")
    } else {
        fmt.Printf("Nope!\n")
        fmt.Println(cs)
        fmt.Println(cs2)
    }
}


type Struct02 struct {
    Index int32
    Arr []byte
    Secret abstract.Secret
    Point abstract.Point
}

// this test crashes since the to-be-(un)packed struct includes elements with
// types other than abstract.Secret and abstract.Point
func test02() {

    rand := suite.Cipher([]byte("example"))
    x := suite.Secret().Pick(rand)
    y := suite.Point().Mul(nil, x)

    cs := Struct02{
        Index: 4, // it seems this value is not read at all by abstract.Read
        Arr: []byte{5,6}, // during abstract.Read the two bytes are merged into the Secret below which in turn corrupts reading of Secret and Point
        Secret: x,
        Point: y,
    }

    buf := new(bytes.Buffer)
    err := abstract.Write(buf, cs, suite)
    if err != nil {
        fmt.Printf("Error Writing : %v\n", err)
    }

    buf2 := bytes.NewBuffer(buf.Bytes())
    cs2 := Struct02{}
    err = abstract.Read(buf2, &cs2, suite)
    if err != nil {
        fmt.Printf("Error Reading : %v\n", err)
    }

    fmt.Printf("Test02: ")
    if reflect.DeepEqual(cs,cs2) {
        fmt.Printf("Hurray!\n")
    } else {
        fmt.Printf("Nope!\n")
        fmt.Println(cs)
        fmt.Println(cs2)
    }
}


func main() {
    test01()
    test02()
}
```

nikkolasg: Found it !
We just have to change the line in abstract.Read (~182) 

```
binary.Read(de.r,binary.BigEndian,obj)
```
TO
```
binary.Read(de.r,binary.BigEndian,v.Addr().Interface())
```
binary.Read needs a pointer to the type he is going to read,and not the value itself.

Now my example works, and your example Philipp just need one arrangement line 83:
```
cs2 := Struct02{}
```
TO
```
cs2 := Struct02{Arr: make([]byte,2)}
```
Indeed, the struct needs to know how many bytes to read, otherwise the slice is just an empty slice and won't read anything (so the ints are "going" into the secret field after).

NOTE: We should always use pointer when call abstract.Read / Write (at least read), as it is in the "binary" package.

Daeinar: Great, now the test passes. Makes of course sense that the function needs to know how many bytes it should read.

Daeinar: Got it. I think there is still some debug output in there though. Or is this on purpose? See:
https://github.com/DeDiS/crypto/blob/master/abstract/encoding.go#L141
https://github.com/DeDiS/crypto/blob/master/abstract/encoding.go#L215

nikkolasg: I've found another bug/issue at line 161
```
if v.IsNil() {
    panic("...")
}
```
That means that when using structs with pointers inside, you had to initialize the pointers before calling abstract.Read. That's not really scalable nor usable. Instead i've changed to:
```
if v.IsNil() {
    v.Set(reflect.New(v.Type().Elem()))
}
```
The effect of this is that it is initializing the pointer before reading the underlying struct, something like 
```
ptr := new(MyStruct)
```
You can now call abstract.Read with basic types + (unitialized) pointers inside your structs !

**NOTE**: again I've pushed this change into the master branch.

nikkolasg: Forgot to remove the prindent calls. Pushed to main branch again. 
Which tests are you talking about ?

Daeinar: I was only talking about the two tests from my example above.

Daeinar: And thanks for removing the debug output. :)

nikkolasg: I've added one more piece of code that i find useful is the ability to read/write simple "int". The code in the master branch only allow to use specific fixed-length type such as int32 or uint64. In my structs and many others, we use the generic type "int". What I did is simply writing the value as an int64 and reading it as an int64. So the int value is read transparently to the user and sent across the wire as a int64. 
Would you be interested if I push that to the master branch or not ? 
As always there are some tradeoff, here we must create a new variable of type int64, but I don't think that will make things slower than using the reflect library at all ;)

bford: Sorry for not being able to keep up or reply earlier yesterday or today, due to the Zurich trip and my first lecture this afternoon…

Please do NOT directly push changes to the master branch, even when they seem like trivial or obvious bug fixes.  Commit them to an alternate branch and open a pull request for discussion, peer review, and signed.  You can keep your own persistent working branch if you don’t like creating temporary branches each time, and if anyone else needs to use your fixes right away, they can ‘git checkout’ your branch or merge the changes from your branch into theirs.

In particular, what might seem like a trivial or obvious “bug fix” to someone might seem more subtle, problematic, or multifaceted to someone else. :)  For example, you decided unilaterally that the right semantics for what happens when the reader encounters a null pointer is that it should silently create an instance of the appropriate type.  It’s not obvious to me that those are necessarily the right or appropriate semantics.  

For example, these “auto-construct” semantics for pointers are inconsistent with the behavior for slices, which as you observed, require the slice to be created with the appropriate length before the Read so that Read will know how many elements to decode.  This basic binary encoding obviously cannot instantiate a nil slice on Read, because the encoding doesn’t transmit metadata on how long the slice should be.  (Unlike protobuf or another more sophisticated encoding, for example).  Thus for this encoding the only two feasible semantic behaviors for reading a nil slice are to fail/panic or to treat a nil slice like an empty slice.  I’m not entirely sure which is better - treating it like a nil slice might be slightly more convenient when that’s what the developer wants, but it also might risk leading to harder-to-find bugs when the developer actually intended to initialize the slice to a non-nil length but just forgot to initialize it.  If the “forgot to initialize it” case is more common, that might actually argue for a fail/panic to be the right semantic behavior, so the error will be caught more immediately.  I’m actually not sure which of these two behaviors the code currently exhibits - forgetting to test, think about, decide on, and document that corner case is actually probably a meta-bug that needs to be fixed one way or another.  But certainly the behavior isn’t and can’t be just “create a non-nil slice of the ‘right’ length” because the ‘right’ length is unknown by Read.

So back to the semantics for nil pointers: should the right semantics be to create an instance?  Maybe, but I’m not sure.  That seems necessarily inconsistent with slice handling, at least.  Why was a pointer type used there in the first place in the type (usually a struct) containing it, and not just a direct non-pointer instance of the pointed-to type?  The protobuf encoding/decoding code (dedis/protobuf) takes a pointer-type struct element as indicating an “optional” protobuf field - which may or may not be the right thing, but the point is it’s semantically meaningful: a pointer-to-X is different from X and there might thus be a reason or need for the encoding/decoding code to treat them differently.  The simple fixed-size binary format, unlike protobuf, doesn’t have metadata with which to indicate the presence or absence of a field, so that particular interpretation of the meaning of a pointer-type struct element doesn’t make sense.  What exactly does make sense?  I’m not sure, but I’m hesitant to say the right behavior is just to create an instance automatically.  If the developer “always” wanted an instance (i.e., a mandatory, non-optional field), then shouldn’t he have just used a non-pointer embedded instance of the type in the struct instead of a pointer type?  If the fixed-size binary encoding requires developers to instantiate slices before Read, why shouldn’t it also require developers to instantiate pointers before Read?

At any rate, again the meta-point is that these kinds of things need to be discussed - or at least given the opportunity for discussion if needed - by using branches and pull requests before immediately committing to the master branch.

Bryan

nikkolasg: All right. Sorry for the "automatic" push. Please delete the last two commits then (easily done on the github interface). I'll keep the changes on my local branch. I'll post my thoughts on this next week.

nikkolasg:
Some comments:
1. Abstract/encoding.go : "XXX now this code could/should be moved into a separate package". Indeed, we should but what would be our options :
  * In the top level (since it will probably be used by every other packages at some point) but we would run into the problem of package dependency we talked about.
  * In a separate package. That would enable us to extend at some point on the encoding part of the library. One future development I could see is the use of compression for example (big topic)
  
Further notes:
2. For all 'Suites' defined in various packages (edwards/{ed25519/},nist/{qrsuite} ...) :
Instead of calling 
   ```
      func (s *suiteEd25519) Read(r io.Reader, objs ...interface{}) error {

          return abstract.SuiteRead(s, r, objs)
      }
   ```
We could create a field 'bc BinaryEncoding' and initialize it at the creation of the suite ? So there is not a new instantiation of a BinaryEncoding struct every time it is called. But it seems only a micro optimization and is  maybe not so relevant for now ;)
3. Abstract/encoding_test.go : We should include at least one pointer in the struct being serialized, and test for its automatic inialization. 
4. Abstract/encoding_test.go : Should we integrate also some serialization of Point + Secret of different suites or is it already tested somewhere ? (since it was already provided by abstract.Read/Write before)

bford:
1. I was thinking of moving the BinaryEncoding type into another package called (for example) crypto/binary or crypto/encoding.  Or perhaps we should be more descriptive about what kind of encoding it is: for example, I thought of the term "rigid encoding" meaning that all encodable objects have well-defined lengths dependent only on type and configuration information (e.g., cipher suite).  So this could go into a package crypto/encoding/rigid, alongside any other different encoding packages we might add later.  But I haven't decided for sure if this is really worth doing (yet); it's only one type and hence doesn't clutter the abstract namespace too much.  Perhaps we should just leave it as-is for now and mull it a bit?

2. I don't think "creating" a new BinaryEncoding object each time matters in terms of performance because it's purely by-value (we never take a pointer to this created BinaryEncoding object) and hence the compiler will treat it as a trivial inline wrapper for the Constructor interface-pointer it contains.  I'm pretty sure it doesn't/shouldn't actually do any kind of object-allocation on each SuiteRead/SuiteWrite call.

3. Agreed; can you add that test?

4. That would be nice but I don't think we can add concrete Point/Secret marshaling tests in this module without creating circular dependencies.  But the general ciphersuite-testing code in crypto/test/test.go does test encoding and decoding of every cipher suite; hopefully that should be enough (or if it isn't let's add the tests there).

nikkolasg:
Done.

nikkolasg:
Does someone has something else to comment on this shamir branch ?
It'd be quite nice if we could merge it to the master branch soon as some PHD's are starting their work using the crypto library (and not necessarily wanting to do a git checkout shamir ;)

poly/constants.go
```diff
@@ -0,0 +1,15 @@
+package poly
+
+import (
+	"github.com/dedis/crypto/edwards"
+)
+
+const (
```
bford: What are these consts for exactly, and how do you intend them to be used?  It looks like they're only used in testing?  Anything that is only used in testing should be in a _test.go file.  And in any case, I don't really see the benefit of using vague terms like MINIMUM (minimum of what set of possibilities) or MODERATE for picking test configurations; why not just list a few actual configurations to test (e.g., specific t and n thresholds, specific ciphersuites, etc.)?

poly/constants.go
```diff
@@ -0,0 +1,15 @@
...
+	MINIMUM = iota
+	MODERATE
+	MAXIMUM
+)
+
+var SECURITY int = MAXIMUM
+
+var SUITE = edwards.NewAES128SHA256Ed25519(true)
```
bford: This *definitely* needs to be fixed before merge.  It's OK to have a global var declaring a particular suite to be used for testing purposes, but that should be only in a _test.go file, and I would call it 'testSuite' or just lowercase 'suite' to be consistent with Go conventions and other package.  And it's definitely *not* OK to have the non-test code dependent on a global SUITE variable like this: all code that depends on the ciphersuite needs to be parameterized properly.

poly/helper_test.go
```diff
@@ -0,0 +1,106 @@
+package poly
+
+import (
+	"fmt"
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/config"
+	"github.com/dedis/crypto/random"
+)
+
+var edward = SUITE
```
bford: Unnecessary: just declare a standard test suite like 'suite' or 'testSuite' and refer to that.

poly/helper_test.go
```diff
@@ -0,0 +1,106 @@
...
+	d := make([]*Dealer, n)
+	for i := 0; i < n; i++ {
+		d[i] = NewDealer(info, generateKeyPair(), generateKeyPair(), receiverList)
+	}
+	return d
+}
+
+// Returns N dealers with M receivers with the right keys / public keys ...
+func generateNDealerMReceiver(info PolyInfo, n, m int) ([]*Dealer, []*Receiver) {
+	receiverKeys := generateKeyPairList(m)
```
bford: Are all the above four functions called only from here?  Looks like it, and if so, having all those separate functions just seems to be creating unnecessary code-bloat.  Just inline the four functions here.  And while you're at it, just have one loop to initialize all four arrays together.  The code will be way more concise and simpler.

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/dedis/crypto/config"
+	"io"
+)
+
+// This package provides  a dealer-less distributed verifiable secret sharing
+// using Pedersen VSS scheme as explained in "Provably Secure Distributed
+// Schnorr Signatures and a (t, n) Threshold Scheme for Implicit Certificates"
+// This file is only responsible for the setup of a shared secret among n peers.
+// The output is a global public polynomial (pubPoly) and a secret share for
+// each peers.
```
bford: You declare a lot of public types below in this module.  Is it actually appropriate for all of them to be public?  Are any of them unlikely to be conceivably useful to anyone outside this module?  I'm not sure yet myself, but should be considered carefully in any case, to avoid unnecessary type clutter that will make it difficult for potential users to penetrate and understand this package's API.

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+
+// PolyInfo describe the information needed to construct (and verify) a matrixShare
+type PolyInfo struct {
```
bford: PolyInfo is vague.  Rename "Threshold" perhaps?

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+// for a peer
+type SharedSecret struct {
+
+	// The shared public polynomial
+	Pub *PubPoly
+
+	// The share of the shared secret
+	Share *abstract.Secret
+
+	// The index of this share regarding the secret private poly / pub poly
```
bford: Vague comment: please clarify.  Index in *what* array or list?

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	T int
+	// How many peers do we need to verify
+	R int
+	// How many peers are collaborating into constructing the shared secret ( i.e. MatrixShare is of size NxN)
+	N int
+}
+
+// Represent the output of a VSS Pedersen scheme : a global public polynomial and a share of its related priv poly
+// for a peer
+type SharedSecret struct {
```
bford: Please clarify further in the comment how this type is used, not just what it represents.  e.g., calling foozlebob generates it, and it is needed as input to function blahdeblah.

bford: Also, should the fields be public or private?

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	// The share of the shared secret
+	Share *abstract.Secret
+
+	// The index of this share regarding the secret private poly / pub poly
+	// i.e. it is the same as the receiver's index
+	Index int
+}
+
+// Dealer is a peer that will create a promise and distribute it to each receivers needed
+type Dealer struct {
```
bford: Similarly, please clarify the comment a bit further, and should these fields be public or private?

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	// info is just the info about the polynomials we're gonna use
+	info PolyInfo
+
+	// This index is the index used by the dealers to make the share for this receiver
+	// For a given receiver, It should be the same for every dealers /!!\
+	index int
+
+	// the Receiver private / public key combination
+	// it may or may not have to be the long term key of the node
+	Key *config.KeyPair
```
bford: 'info' and 'index' are private, but 'Key' and 'Dealers' are public: intensional?  justification?

poly/joint.go
bford: Coming to think of it further, it's not clear to me exactly what added value the Dealer type is contributing beyond the Promise and/or State types.  Do we really need an additional public type in the API just to encapsulate a Promise and a State?

When I suggested the Dealer name, my intent was that the existing Promise type be renamed to Dealer, not that a new type called Dealer be added.  Is that not workable?

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	r.Key = key
+	r.Dealers = make([]*Dealer, 0, info.N)
+	return r
+}
+
+// AddDealer adds a dealer to the array of dealers the receiver already has.
+// You must give the index of the receiver in the promise of the dealer,
+// i.e. index is generally the index of the receiver in the matrix, and
+// is usually fixed.
+// Most of the time it will be like , for peer i :
```
bford: shouldn't be a space before a comma.

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	// This index is the index used by the dealers to make the share for this receiver
+	// For a given receiver, It should be the same for every dealers /!!\
+	index int
+
+	// the Receiver private / public key combination
+	// it may or may not have to be the long term key of the node
+	Key *config.KeyPair
+
+	// List of Dealers. Be careful : this receiver should have the SAME index for all the Dealer's promises !!
+	// otherwise we wouldn't know which index to chose from the shared public polynomial
```
bford: Please clarify this comment.  I don't understand what "for all the Dealer's promises" means.  Doesn't one Dealer have only one Promise?  Perhaps you instead mean "should have the SAME index in each of these Dealers"?

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+// is usually fixed.
+// Most of the time it will be like , for peer i :
+// receiver(i).AddDealer(i,dealer(j)) for all j
+// It will return a Response to be sent back to the Dealer so he can verify its promise
+func (r *Receiver) AddDealer(index int, dealer *Dealer) (*Response, error) {
+	if r.index == -1 {
+		r.index = index
+	}
+	if r.index != index {
+		return nil, errors.New(fmt.Sprintf("Wrong index received for receiver : %d instead of %d", index, r.index))
```
bford: Is this an error that should ever be able to happen outside the control of this program (due to external circumstances such as some other node's action or an incorrect message being received)?  If not - i.e., if this error can only happen because of a program bug in the caller of this function - then this should be a panic, not an error return.  And if there's *no* way this function should ever be able to fail due to external, uncontrollable circumstances (only program bugs), then it probably shouldn't have an error return at all.

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+// ProduceSharedSecret will generate the sharedsecret relative to this receiver
+// it will throw an error if something is wrong such as not enough Dealers received
+func (r *Receiver) ProduceSharedSecret() (*SharedSecret, error) {
+	if len(r.Dealers) < 1 {
+		return nil, errors.New("Receiver has 0 Dealers in its data.Can't produce SharedSecret.")
+	}
+	pub := new(PubPoly)
+	//pub.InitNull(r.info.Suite, r.info.T, r.Dealers[0].Promise.PubPoly().GetB())
+	pub.InitNull(SUITE, r.info.T, SUITE.Point().Base())
+	share := SUITE.Secret()
```
bford: Non-parameterized suite: not OK

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	goodShare := 0
+	for index, _ := range r.Dealers {
```
bford: you don't need the ", _" here do you?

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+		// Only need T shares
+		if goodShare >= r.info.T {
+			break
+		}
+		// Compute secret shares of the shared secret = sum of the respectives shares of peer i
+		// For peer i , s = SUM fj(i)
+		// TODO WARNING : THIS IS TEST IMPLEMENTATION
```
bford: As a matter of principle I think we should fix "TEST IMPLEMENTATION" things like this before merging into a mainline branch (even the 'development' branch).  In particular, any TODO that will leave a gaping security hole if the code gets used without being fixed should really be taken care of before going into a mainline of any kind.

poly/joint.go
```diff
@@ -0,0 +1,284 @@
...
+	return p.N == p2.N && p.R == p2.R && p.T == p2.T
+}
+
+// Dealer must implement Marshaling interface (abstract/encoding.go)
+func (d *Dealer) UnmarshalInit(info PolyInfo) *Dealer {
+	d.Promise = new(Promise).UnmarshalInit(info.T, info.R, info.N, SUITE)
+	return d
+}
+
+func (d *Dealer) MarshalBinary() ([]byte, error) {
```
bford: Reduce duplication: e.g., either have MarshalBinary by a simple wrapper around MarshalTo, or vice-versa.  No need to have two complete sets of marshal/unmarshal code just because there are two entrypoints to it.

poly/schnorr.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// The partials signature of each other peers (i.e. receiver)
+	partials []*PartialSchnorrSig
+
+	/////////////////////////////////////////////////////
+}
+
+// Partial Schnorr Sig represents the partial signatures that each peers must generate in order to
+// generate the "global" signature. This struct must be sent across each peers for each peers
+type PartialSchnorrSig struct {
```
bford: Maybe rename 'SchnorrPart' or something similar like that?  (and starting with 'Schnorr' so that all the 'Schnorr*' types will appear together in the godoc API docs)

poly/schnorr.go
```diff
@@ -0,0 +1,270 @@
...
+	// The partial signature itself
+	Part *abstract.Secret
+}
+
+// SchnorrSig represents the final signature of a distribtued threshold schnorr signature
+// which can be verified against a message
+// This struct is not intended to be constructed manually but can be:
+//  - produce by the Schnorr struct
+//  - verified against a Schnorr struct
+type SchnorrSig struct {
```
bford: If this is only supposed to be treated as an opaque blob by clients of this package, shouldn't it be private?  Are the fields public because it's used for marshaling/unmarshaling?  If so, it might be better for the clients just to see []byte-slices, and have this be a private struct used only internally for marshaling.  (Note that private structs can have public fields and can be used for mashaling/unmarshaling.)

poly/schnorr.go
```diff
@@ -0,0 +1,270 @@
...
+	s.info = info
+	s.Longterm = longterm
+	return s
+}
+
+// NewRound set the random key for the d.schnorr algo + set the msg to be signed
+// You call this function when you want a new signature to be issued on a specific message
+// The security of the d. schnorr signature protocol is the same as the regular :
+// The random secret MUST BE FRESH for EACH signature / signed message (hence the 'NewRound')
+func (s *Schnorr) NewRound(random *SharedSecret, msg []byte) error {
```
bford: Any particular reason we need a NewRound() function, instead of the API simply to be that the client needs to re-initialize for each signature?

bford: Also, this should not take a message as a byte-slice, but probably a hash.Hash object.  That way the caller can handle large messages efficiently via streaming before calling this.

poly/schnorr.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	if nsig < s.info.T {
+		return errors.New(fmt.Sprintf("Received to few Partial Signatures (%d vs %d) to complete a global schnorr signature", len(s.partials), s.info.T))
+	}
+	return nil
+}
+
+// verifyPartialSig will verify if a given partial signature can be checked against the longterm and random secrets
+// of the schnorr structures.
+func (s *Schnorr) verifyPartialSig(ps *PartialSchnorrSig) error {
```
bford: verifyPart?

poly/schnorr.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// RevealPartialSig reveals the partial signature for this peer
+// Si = Ri + H(m || V) * Pi
+// with :
+// 	- Ri = share of the random secret for peer i
+//  - V  = public commitment of the random secret (i.e. Public random poly evaluated at point 0 )
+//  - Pi = share of the longterm secret for peer i
+// This signature is to be sent to each others peers
+func (s *Schnorr) RevealPartialSig() *PartialSchnorrSig {
```
bford: RevealPart?

poly/schnorr.go
```diff
@@ -0,0 +1,270 @@
...
+	if err := s.verifyPartialSig(ps); err != nil {
+		return errors.New(fmt.Sprintf("Partial signature to add is not valid : %v", err))
+	}
+	s.partials[ps.Index] = ps
+	return nil
+}
+
+// SchnorrSig  will generate the global schnorr signature
+// By reconstructing the secret that the partial responses contains
+func (s *Schnorr) SchnorrSig() (*SchnorrSig, error) {
```
bford: Rename 'Sign()' perhaps?

poly/schnorr.go
```diff
@@ -0,0 +1,270 @@
...
+	return sig, nil
+}
+
+// VerifySchnorrSig will verify if a given signature is correct regarding the message
+// NOTE: This belongs to the schnorr structs however it can be called at any time you want
+// This check is static, meaning it only needs the longterm shared secret, and the signature to
+// check. Think of the schnorr signature as a black box having two inputs
+//  - a message to be signed + a random secret ==> NewRound
+//  - a message + a signature to check on ==> VerifySchnorrSig
+func (s *Schnorr) VerifySchnorrSig(sig *SchnorrSig, msg []byte) error {
```
bford: Similarly, just 'Verify()'?  And take msg as a hash or something like that, not a byte-slice.

util/encoding.go
bford: Do you still think Int2Buf etc are needed?

bford:
Yes, I agree it would be nice to get this merged soon, but it's not ready to merge yet; see my comments. ;)  

And @ineiti have you reviewed in detail as well yet?  If not, can you please when you have a chance?

Thanks.

poly/constants.go
nikkolasg: This SECURITY variable address the concern we talked about the promise.go code. In the current version, a State has to have received all Responses before revealing any share. In this case, by setting SECURITY to MINIMUM, the State does NOT check if it has received all responses before revealing the share. THis behavior is needed in order to easily establish a shared secret as done in joint.go. Moderate is just extra, and can be removed without further notice.

poly/helper_test.go
nikkolasg: The first functions are called inside joint_test.go and the latest are used inside schnorr_test.go and are dependant on the first one so... Think of these functions as helpers because both test suite (joint / schnorr) use them.  These functions are separate because they are used independently in each test functions in joint.go. If i do make them united, it will reduce the coverage of my tests greatly. And if i do remove this helper file, I'll just have to write the same functions in both joint.go + schnorr.go.

poly/joint.go
nikkolasg: I know this is duplication but the only solution to get around that is in ReadFrom, I should read ALL the buffer (that is what is done in promise.go) and I am not particularly fan of this approach. Indeed, the io.Reader given may contain additional data not related to this structure and we just wipe them out by reading the whole "reader" interface. The abstract.Encoding interface contains too much method to my taste (no need to get both marshaling capability to buf + writer, encoding only uses writer method anyway). I just made those two functions have similar code in order to not discard the rest of the buffer in the "reader" method. Does that make sense ?

poly/schnorr.go
nikkolasg: Are you talking about the PartialSchnorrSig ?  If yes, you are right it's better to put it a private struct.

poly/schnorr.go
nikkolasg: > Any particular reason we need a NewRound() function, instead of the API simply to be that the client needs to re-initialize for each signature? (i.e., just combine NewRound into Init?)

Yes, because we don't want to recreate a new long term shared key each time we want to sign something, it's only a one-time operation. But we want a new random shared key each time we want to sign something

Right about the hash.Hash, didn't know this one in the lib.

poly/schnorr.go
nikkolasg: Absolutely.

poly/constants.go
bford: I see.  But then there should be parameterized configuration state variables that actually describe what they do, rather than global constants with vague names like "minimum/moderate/maximum security." :)

Also, now might actually be a good time to do away with that whole "Certified" process and threshold once and for all anyway.  Any roadblocks to doing that?  Or is that more tricky than it seems for any reason, which might warrant doing it separately?

poly/schnorr.go
nikkolasg: For this one, since it is a **public** function I prefer the more explicit terms instead of the shorter one...

poly/schnorr.go
nikkolasg: SInce it is a private one, it's fine for me "verifyPart" ! but it has to be consistent with the "RevealPartialSIg" discussion below !

poly/schnorr.go
nikkolasg: Agreed.

util/encoding.go
nikkolasg: Definitely not !

poly/helper_test.go
bford: I see.  But it doesn't look like generateKeyPair() is ever called independently of generateKeyPairList(), so those could readily be merged into one, right?  (just have it return two slices.)  Similarly, looks like generateReceivers and generateDealers are always called together so they could be merged too, right?  (and it makes sense to me intuitively anyway: you always need to set up a dealer/receiver matrix of some kind!)

poly/joint.go
bford: If that's what the promise code does, that's probably wrong: for this encoding, ReadFrom is supposed to read only "what it needs" and leave any extra data unread.  

And I agree that abstract.Encoding isn't as minimal as it should be; I want to work on fixing that, but probably as a separate subproject.

poly/schnorr.go
bford: But the Init function takes a 'longterm *' as a parameter anyway: the caller doesn't have to "re-create" the longterm secret each time it initializes or re-initializes a Schnorr instance.  It just has to pass the same longterm * parameter each time, no?

poly/constants.go
ineiti: I'd prefer getting this branch in shape and merged before putting in other things, mostly because the ```cothority``` also depends on it.

+1 for doing the threshold in a new branch, once this is accepted.

poly/helper_test.go
nikkolasg: generateKeyPair() is called in the generateDealers() method. 
It's Ok for generateReceivers / Dealers.
I will not argue more here since this is only for a helper test file but I will still say one last thing ;) From a global point of view, I think having separate functions for separate functionalities is always better, even if you create a function of 4 lines, than having a bigger functions that does many things. (But of course if your goal is to have a minimum l.o.c number, then this is not the thing to do !) 
I've merged the dealers / receivers part in the matrix setup.

poly/joint.go
nikkolasg: Agreed.

poly/joint.go
nikkolasg: On one hand, I think they should be public since the owner of this SharedSecret has the "right" to know its share, since it is its own. This struct is only a collection / holder of different pieces of information to represent the share of a node, and, in contrast, is not a fully oop-class-like struct with its own methods set and stuff, where private fields make sense. At least to me ;)
On the other hand, it is necessary that the pub poly, share & index stay constant and are not modified in any way by the holder of the struct otherwise the share would become invalid. In that way, it is preferable that they become private fields.
One trade off for these two ideas would be to make the fields private and make some Getter on it. Would that be ok for you ?

poly/joint.go
nikkolasg: Yes it is absolutely workable. We should talk about wether we need the verification of the signature (validProof / blameProof) to happen when we generate the SharedSecret. If we do not **need** that, then a simple promise (so maybe renamed to Dealer) can do the trick and it is all good ! If we **do** need that verification (and if it is in promise, it must be for a good reason ;) then we need the dealer to send back its State to each peer after having received each signatures. I see two possible problems with that. **First**, the returned state would contain every shares of every one, and that is definitely not needed. **Secondly**, the terminology from the Receiver point of view would not be so elegant as we would have a :
* AddPromise (or AddDealer) (index,dealer)
* AddState (index, state) **MUST** be called at least T times ( for each dealer) before ProduceSharedSecret()

The first problem can be changed in a separate branch in the promise.go and do not impact this code. However, we must find a suitable "algorithm flow" for the second problem. Since it is closely related to how the promise works, we must talk in details about how are we going to restructure promise.go so this code can already be adapted to the future changes.

poly/joint.go
nikkolasg: Not really intentional. In fact, they moved from being public to private just now !

poly/joint.go
nikkolasg: Corrected.

poly/joint.go
nikkolasg: >Is this an error that should ever be able to happen outside the control of this program (due to external circumstances such as some other node's action or an incorrect message being received)?
Totally, for example a dealer sending the wrong share to a receiver. It is totally external to the local control of the program. So in my understanding of your questions, it should have a return error.

poly/joint.go
nikkolasg: Again, we should talk about the underlying promise code verification mechanism.

poly/joint.go
nikkolasg: Corrected.

poly/joint.go
nikkolasg: Corrected

poly/schnorr.go
nikkolasg: Indeed. We have two different opinions on how we see this algorithm-flow. My approach looks more oriented towards a kind of distributed-schnorr-signature-as-a-service, where the matrix is already done, the longterm secret already computed, and each time we want to sign something, we just launch a new round. We expect this "service" to be used many times, so we don't have to recreate a Schnorr struct for each signing request. 
Yours looks more one-time fashion where we may or may not need to do the setup for each signing request, and we can sign something as long as we have both the longterm and the random shared secret. It is more useful when we know we don't have to accept many signing requests but rather on a one-time only manner.

Which approach is better than the other ? Linus, comments ? ;)

poly/schnorr.go
nikkolasg: moved to SchnorrPartialSig

poly/schnorr.go
ineiti: As Nicolas said: it's about style, and as the rest of the library we're using is Bryan-styled, I propose we follow here. It's not a big initialization that is done in ```Init``` or such, so we don't loose anything if we do it every time we want to sign something.

For fun we could write down preferred minimum-sizes of functions and whether comments should start with the name of the function or not ;)

nikkolasg: Good call !

nikkolasg:
Ok so' I've removed the certification process, removed the Dealer struct, change promise to Deal. All tests working and it is a lot cleaner !

bford:
Excellent!

poly/deal.go
```diff
@@ -234,10 +234,10 @@ type Promise struct {
  *    secretPair.Suite == longPair.Suite
  *
  * Returns
- *   A newly constructed Promise
+ *   A newly constructed Deal
  */
-func (p *Promise) ConstructPromise(secretPair *config.KeyPair,
-	longPair *config.KeyPair, t, r int, insurers []abstract.Point) *Promise {
+func (p *Deal) ConstructDeal(secretPair *config.KeyPair,
```
bford: I think this method should just be called Construct or Create; no need to include Deal in it since it's a Deal method.

poly/deal.go
```diff
@@ -276,68 +276,74 @@ func (p *Promise) ConstructPromise(secretPair *config.KeyPair,
...
+ *                  deal
  *    n           = the total number of insurers.
- *    suite       = the suite used within the Promise
+ *    suite       = the suite used within the Deal
  *
  * Returns
- *   An initialized Promise ready to be unmarshalled
+ *   An initialized Deal ready to be unmarshalled
  */
-func (p *Promise) UnmarshalInit(t, r, n int, suite abstract.Suite) *Promise {
```
bford: I think this and other methods called UnmarshalInit should just be called Init.  That's the standard name we've been using to initialize otherwise-"empty" objects; doesn't matter if they subsequently get used for unmarshmaling or something else.

poly/deal.go
ineiti: Isn’t it possible to make these constructors/initialisers more unified? I think we have (mixed from different structures/interfaces):

* Init
* Create
* Construct
* NewDealer

According to

https://golang.org/doc/effective_go.html#composite_literals <https://golang.org/doc/effective_go.html#composite_literals>
and
http://stackoverflow.com/questions/4498998/how-to-initialize-members-in-go-struct <http://stackoverflow.com/questions/4498998/how-to-initialize-members-in-go-struct>

it should all be ```New*``` anyway, no ```Init```, no ```Create``` and no ```Construct```.

What do you think?

nikkolasg: I think we should talk about this more in depth when we start refactoring the promise / deal code ...?  Like we said yesterday, there are others issues in this code and i think It deserve its proper pull request.

PR #45, opened 2015-09-18
Int type case + pointer initialization

nikkolasg:
Here we are, the branch regarding the abstract/encoding.go file.
I've added a case 
```
reflect.Int
```
where it encodes / decodes that value as an int64 and set the value back into the field. 
I think the cost of creating an int64 is fairly low compared to using the reflect library, and the advantages are quite high, as we often use simple "int" in our structs. By "we", i mean the coders of this library AND the users of this library too, that may not want to use fixed-type length.

Also I've added the automatic initialization of the underlying struct when we got a nil pointer. That was the subject of the discussion in the "shamir" pull request, that should continue here ;)


bford:
Thanks Nicolas.

I'm coming around to agreeing with you regarding what the nil-pointer behavior should be.  It's probably more important for it to be consistent with the semantics of interfaces than that of slices, and for interfaces the auto-construct behavior is already well-established and fairly important in making the encoding system convenient for objects like Point and Secret.  In fact I've been meaning to make the interface-constructor system more general in the same way the protobuf implementation already is - i.e., using a Constructors or similar parameterized map from reflective interface types to constructor functions.  At any rate, if we consider pointers and interfaces to be comparable (other than concrete vs abstract) in Go's semantics, it probably makes sense to have auto-instantiate semantics for both.  Thanks for bringing up this issue.

I'm less convinced (yet anyway) that treating int types as int64 is the right thing.  The whole idea of int types in Go's language semantics is that it's size-indefinite; hence it's less clear that it makes sense to "pretend" it has a definite size when it doesn't.  I do certainly see the appeal in terms of convenience though.  Let me think about that one a bit more.

Regardless, one thing we should probably do is add panic() calls with more clear messages for any situations we decide just aren't allowed/supported by the library.  (e.g., if we decide size-indefinite int shouldn't be supported by this encoding, add an explicit detection and panic saying so, rather than adding confusion for everyone new who comes along and uses it.)  

Another situation where I think the library might currently have less-than-friendly behavior is with the marshaling of nil slices.  Does Write() of a nil slice panic or just treat it as a zero-length slice?  For this kind of thing I'm somewhat inclined to say the code should detect this and panic, because it most likely means the developer just forgot to initialize the slice to the right length - and in the rare case the developer really did want a zero-length slice, it's easy to create one.

bford:
Further thinking out loud…

I’m coming around to thinking that we can simply specify that the “standard integer size” for this fixed-size binary encoding is 64-bits, and marshal/unmarshal Go’s ‘int’ type as an int64.  

However, my remaining concern is that in corner cases this could lead to subtle security bugs, e.g., if an attacker marshals an int that’s too large for platforms for which int==int32 but representable on platforms for which int==int64, perhaps (for example) causing different platforms’ behaviors to diverge in maliciously-controlled ways.  I could envision this being used to create quasi-equivocation attacks across different platforms for example.  It's hard to think of specific scenarios, but the generic security lesson we keep learning over and over is "you'd be surprised"...

I think we could head off this class of risks simply by ensuring that Read’s unmarshmaling code for ‘int’ types explicitly checks that the unmarshaled int64 is in fact representable in the local platform’s int, and throw an error (e.g., “decoded int is too large for this platform”) if the incoming int64 isn’t representable in an int.  Just unmarshal the int64 into the destination int, then convert the int back to an int64 and check for equality.

If you agree with the reasoning, could you change your branch/pull-request to include this check, as well as to ensure that trying to Write or Read a nil slice causes a panic with a comprehensible message?  And maybe include corner-case Test cases that exercise both of these failure scenarios?

Finally, we should probably start documenting this encoding properly somewhere, perhaps in the godoc documentation for Read or Write - i.e., add some notes saying how this encoding works, how int is int64, how pointers and interface types are handled, how slices are handled (i.e., must be initialized at both ends to correct length), etc.  It doesn’t have to be beautiful, but we should probably at least quickly document these behaviors so they don’t keep confusing newcomers to the crypto library (as they have multiple times before). :)

nikkolasg:
>"t's probably more important for it to be consistent with the semantics of interfaces than that of slices, and for interfaces the auto-construct behavior is already well-established and fairly important in making the encoding system convenient for objects like Point and Secret" 

I totally agree, slice is a specifc construct of Golang has its own kind of "initializer"  (e.g. make). 
>I think we could head off this class of risks simply by ensuring that Read’s unmarshmaling code for ‘int’ types explicitly checks that the unmarshaled int64 is in fact representable in the local platform’s int, and throw an error (e.g., “decoded int is too large for this platform”) if the incoming int64 isn’t representable in an int. Just unmarshal the int64 into the destination int, then convert the int back to an int64 and check for equality.

Indeed, that should be the perfect balance between usability and encoding format/security. 
> as well as to ensure that trying to Write or Read a nil slice causes a panic with a comprehensible message? And maybe include corner-case Test cases that exercise both of these failure scenarios?

Yes, it should definitly panic when trying to encode a nil slice.
Where should I put the few test ? in abstract/encoding_test.go ? or in test/ ?

bford:
Mind if I hack on this temporary encoding branch a bit too? :)

Building on your changes, I did a few things (still experimental) on a train ride to Paris this afternoon:

- The most important: Integrated the abstract.Read, abstract.Write API into the Suite interface, so that every Suite has a Read and Write method that no longer needs to take a Group as a separate parameter.

- Added the beginnings of an encoding_test.go, based on the similar test suite I already had in the dedis/protobuf repo.  It doesn't yet do everything it should, but maybe you can add the tests we discussed to that?

- Changed the implementation of the Read/Write methods to be more generic; they now could and perhaps at some point should be moved into a separate utility package somewhere else, since they no longer depend directly on any crypto-specific stuff.  In particular, the Read implementation just takes a 'Constructor' object, which just provides a New method to construct an instance of a given reflect.Type if possible.  And this Constructor.New method now happens to be a standard method of any Suite, so a Suite object itself can be used as that parameter in the crypto case.  

Again this is still very experimental; please comment and let me know what you think - is it a good idea or not so good?  any improvements you can think of?  Certainly various things could still be cleaned up better...

bford:
(Oh yes and I probably forgot to gofmt the code and such - certainly not ready to merge just yet even if you agree with the direction.)

ineiti:
Le 19/09/2015 11:31, Bryan Ford a écrit :
> I think we could head off this class of risks simply by ensuring that
> Read’s unmarshmaling code for ‘int’ types explicitly checks that the
> unmarshaled int64 is in fact representable in the local platform’s int,
> and throw an error (e.g., “decoded int is too large for this platform”)
> if the incoming int64 isn’t representable in an int. Just unmarshal the
> int64 into the destination int, then convert the int back to an int64
> and check for equality.

Or you could only decode len(int) bytes if it is an int - but you could 
still fall short if the sending platform is 64-bit and the receiving one 
is 32-bit.

I'm in with Bryan on "anything unforseeable might just happen" with 
regard to int - throw a panic if you detect one in the code.

nikkolasg:
One problem I had with the current implementation is regarding the unexported fields of a struct,i.e. a private field. It paniced when trying to write it. Should we let it panic (but with a more detailled error message) or should we just ignore the field and go to the next one (if possible) ?
The former solution looks better, maybe with a simple "CanAddr()" check.

bford:
A panic with a more appropriate error message would be great.  I don't think it should just ignore private fields; that could easily lead to very surprising and unexpected wrong behavior.

I think I might not have put detection code in this before because I might have been worried that doing a CanAddr() check on every field might be fairly slow.  But that's probably not the most important consideration here, and I don't even have a good idea (yet) of the performance considerations, so let's not worry about that for now.

nikkolasg:
By the way, this "unexported field" issue brings a lot of trouble for the poly/promise since the structs in it have a lot of private fields. I am currently refactoring to the **minimum** so I can use suite.Read/Write for the joint + schnorr , but keep in minde that we have to go through that at a future point in time, when refactoring promise.go.

abstract/encoding.go
```diff
@@ -88,49 +100,63 @@ var aPoint Point
...
 	fmt.Printf(format, a...)
 }
 
 type decoder struct {
-	g Group
+	c Constructor
 	r io.Reader
 }
 
 // XXX should this perhaps become a Suite method?
```
ineiti: This is done now, so the comment can go, no?

abstract/encoding.go
```diff
@@ -170,13 +193,31 @@ func (de *decoder) value(v reflect.Value, depth int) error {
...
+		fallthrough
+	case reflect.Array:
 		l := v.Len()
 		for i := 0; i < l; i++ {
 			if err = de.value(v.Index(i), depth+1); err != nil {
 				return err
 			}
 		}
+
+	case reflect.Int:
```
ineiti: Some possibilities:
  * always treat int as int32, but fail if we cut off the value -> faster on 32-bit architectures, but eventually failing when trying to transmit too big values
  * transmit as int64, try to cast to an int and fail if it's not possible -> faster on 64-bits, and if everybody is 64-bit, you can put bigger values, but might break once the app is rolled-out
  * log.Fail -> enforces usage of int32 or int64, but is complicated with lots of other libraries that might use int

A link for why we should allow for ints:
  * http://stackoverflow.com/questions/16427416/what-are-the-advantages-of-the-general-types-int-uint-over-specific-types-i

abstract/encoding.go
bford: Yep, thanks

abstract/encoding.go
bford: Since 64-bit architectures are increasingly the norm, I think we should define 64-bit as the standard "wire-format size" of an int, and just produce a decoding error on 32-bit platforms if the decoded int is too big.  I'm committing a change to do this.

bford:
Ready to merge?  Go ahead if you agree.

abstract/encoding.go
ineiti: I'm afraid hunting down a bug in a client (or a server, when it has to receive) will be a REAL headache - a real heisenbug. Imagine your app working fine in test development, and when you deploy it, all of a sudden the thing crashes on the clients and you have no log?

But then I don't know how many 'int's are currently in a structure that is being (Un)Marshalled by the Suite.Write, which will also be difficult to change.

Can't a basic crypto-suite make some enforcements about what type of int to use, if it's a possible problem?

bford: I'm not sure I quite understand what kinds of [heisen]bugs you're worried about here, or what you want to enforce about "what type of int to use" - can you clarify?  Avoiding silent heisenbugs is the reason I added the specific overflow check and error return if a 64-bit int arrives that's too large for a 32-bit platform to represent as an int internally.  Are there other important failure conditions I'm missing?

ineiti: Suppose you have a 64-bit server and a 32-client. A structure that holds the number of milliseconds since the first round is integrated as 'int' and sent from the server to the client. Now when testing things on your 64-bit computer, everything is going well. Deploying goes well, the first week you're happy.

Unfortunately, after one month your clients start failing, because now the 32-bit can't hold anymore the number of milliseconds. While debugging you restart the server, and everything is OK again, so you're happy for the next month, before things start going bad again.

bford: Right.  It sounds here like you're arguing for not supporting the size-unspecified 'int' type in the binary/rigid encoding at all, right?  At least that's the only solution I can see here that would avoid any risk of those types of bugs.  Or are you proposing something else?

On the other hand, I might argue that there's nothing specific to network wire-encodings in the general class of bug risk you're pointing out.  I might just as well have an 'int' variable that's used only purely locally in my program, which counts the same thing (e.g., number of milliseconds since the first something), and again works fine on a 64-bit machine but dies on a 32-bit machine.  This is a risk we run all the time when we use limited-size int types whose size depends on the platform, and it's certainly a problem, but not one that Go tries to deal with.  

In fact I'm not sure I really like the fact that Go even has a size-unspecified int type at all - that's one way in which I think Java was perhaps better (where 'int' is always 32-bits and 'long' is always 64-bits).  But I do kinda understand their motivations for sticking with this somewhat unfortunate C-ism.  Tastes and philosophies differ here.  But again that doesn't seem like anything specific to a wire format; these considerations and tradeoffs apply to all code written in the language.

ineiti: Yes, I liked my third argument: log.Fail if you encounter an int.
Fail early, fail hard.
But now I know I'd better put int64 on these milliseconds anyway. Or on the clicks for Youtube-videos...

nikkolasg:
For me it's fine to merge.

abstract/encoding.go
ineiti: Last attempt to save my log.Fail, else I'll also OK for the 64-bit:
- standard version fails on int
- 'unsafe' version does the 64-bit conversion - where 'unsafe' might be another name or a flag in the code

bford: I think I'm still agreeing with Nicolas that from a language/programmability perspective it's just pretty inconvenient for the common 'int' to be not supported at all by the encoding.  

Making it a selectable, perhaps non-default option might be a reasonable alternative.  (I would pick something more specific than "unsafe", since there are many possible things that can be done that are "unsafe" and many are probably more unsafe than int conversions. :) )  Perhaps "AllowInt" or something like that.  Maybe even an "IntSize" enum-type with options "NoInt" (disallowed, default), "Int32", or "Int64".  I haven't yet convinced myself this is worth the extra complexity though.

Yet another thought is to support int encoding but make it go the other way, defining the wire encoding of an int as being always 32 bits.  This would ensure that a protocol that gets adequately tested on 64-bit platform will not break because of smaller ints if later run on a 32-bit platform.  Defining int as 32-bit would also be consistent with Java/Scala etc, where int is always 32 bits.

nikkolasg: That is what we discussed also yesterday. Going with the Int32 approach, that would raise error only on the "sender" part, and not on the "receiver" part. If a server is using a big number and wants to encode it as an Int (so implicitly as an Int32), it will fail **before** it has a chance to send it. Better send nothing than send something that will potentially fail, I'd say.

bford: Sure, I'm fine with the int32 approach.  Linus, does that seem reasonable to you?

ineiti: Yes - I preserve the right for a "I told you so" ;)

bford:
Oops - didn't we kinda decide to change from a 64-bit to a 32-bit wire format for 'int'?  Looks like that change didn't get in before the merge?

PR #46, opened 2015-09-23
Beginnings of more general signature-handling framework

bford:
This branch introduces the beginnings of a more general framework for signing and handling signatures.  In particular:
- the 'sig' package defines abstract interfaces for signing schemes
- the 'sig/schnorr' package providing an initial concrete implementation for simple Schnorr signatures.  (More to come later, e.g., ring signatures in the same framework, legacy-algorithm signatures such as RSA/DSA/ECDSA, etc.)
- the Reader and Writer functions in the sig package allow a client to "wrap" any io.Reader or io.Writer with another one that validates or appends signatures, respectively.  My hope is that this will make it more convenient to build protocols that use signed messages on a routine basis, since the signing and verification can now just be composed directly into the reading/writing process between encoding/decoding and whatever the source/sink of the encoded messages are.  But let's see how useful this turns out to be.

Anyone interested, please review and comment.  Also, this branch builds on the encoding branch, which should be reviewed and merged first.


abstract/group.go
```diff
@@ -14,17 +15,27 @@ in which security is based on the Discrete Logarithm assumption,
...
+	String() string
 
 	// Equality test for two Secrets derived from the same Group
 	Equal(s2 Secret) bool
 
 	// Set equal to another Secret a
 	Set(a Secret) Secret
 
+	// Set to a big integer value
+	SetInt(v *big.Int) Secret
```
nikkolasg: I'm not necessarily a fan of overloading the `Secret` interface with this method, when you already added the `SetBytes` method which for me has the same semantic idea: a big.Int is *only* a slice of bytes, both are equivalent. Moreover, this idea of equivalence prevails with the `Secret` becoming a `Scalar`. 
Also, here it makes explicit that a secret *must* be a `big.Int` or at least must deal with `big.Int` where I don't see any reason it should be enforced. Maybe for some weird encryption/signing mechanisms, a scalar won't just *be* a scalar but something more. 
Finally, you don't seem to use it for now ...

sig/key.go
```diff
@@ -0,0 +1,56 @@
...
+	Verify(sig []byte, hash hash.Hash) error
+}
+
+// SecretKey is the abstract interface for a secret signing key
+// that can be used for both signing and verification.
+// This interface may be used with a variety of signing schemes,
+// but it assumes that the signing scheme produces and consumes
+// fixed-size signatures.
+type SecretKey interface {
+	PublicKey
```
nikkolasg: I would stick with `PublicKey() PublicKey` method. Having both seems to me that it only clutters things. Moreover, I have trouble to see the reasoning behind that: why should a `SecretKey` should be considered as a `PublicKey` ?

sig/key.go
```diff
@@ -0,0 +1,56 @@
...
+	"crypto/cipher"
+	"github.com/dedis/crypto/abstract"
+	"hash"
+)
+
+// PublicKey is the abstract interface for a public signing key
+// that can be used for signature verification.
+// This interface may be used with a variety of signing schemes,
+// but it assumes that the signing scheme produces and consumes
+// fixed-size signatures.
```
nikkolasg: Why must we assume that it produces and consume fixed length signature ? I see below you hash all the content using your writer, so the message signature is fixed size regarding the hash. 
Is it something 99% of signing scheme do ? I assume it has some reasoning regarding side-channel attacks ?

sig/key.go
```diff
@@ -0,0 +1,56 @@
...
+type PublicKey interface {
+	abstract.Marshaling
```
nikkolasg: Could we use `RigitMarshaling` here ?

sig/key.go
```diff
@@ -0,0 +1,56 @@
...
+	Sign(sig []byte, hash hash.Hash, rand cipher.Stream) ([]byte, error)
+
+	// Return the PublicKey for this SecretKey.
+	PublicKey() PublicKey
+}
+
+// Scheme is an interface defining an abstract digital signature scheme.
+// An instance provides constructors for PublicKey and SecretKey objects
+// for a particular digital signature scheme.
+type Scheme interface {
```
nikkolasg: Is there a reason why you put that as a separate package ?
I'm asking because it seems it would be a good addition to the `Suite` having a `SigningScheme()` method, so we use the signing scheme associated with the suite we are using. With your example it works fine like this, but some signing algorithms, standards one, are designed with a particular suite in mind, and maybe be completely unsafe if used somewhere else. 
What do you think ?

sig/writer.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// Create a Writer that interposes on underlying Writer wr,
+// hashing all bytes written and appending a trailing signature on Close.
+//
+// The signature is assumed to be fixed-size at least for a given public key,
+// and is appended to the end of the written data with no framing metadata.
+// Thus we can sign arbitrary-size streaming messages efficiently,
+// and the message's total size does not need to be known in advance.
+func Writer(wr io.Writer, key SecretKey, rand cipher.Stream) io.WriteCloser {
```
nikkolasg: I like the idea of automatically appending a signature on messages :) 
One small detail though: it seems more ... *natural* to pass a `Scheme` here instead of a `SecretKey`, as some signing mechanisms also need the public key for signing. Of course you can do that only having the `SecretKey` and calling `PublicKey()` on it, but the way I would like to read this method's signature is : "Give me a writer that appends signature  using this signing scheme", vs "Give me a writer that appends signature using this secret key".
Well, that's only semantics.. :)

nikkolasg:
DO we want that for v1 ?

bford:
Hmmm, I guess I'd forgotten about this since last year. :/  No, it probably shouldn't be for v1, if it's not actually needed currently by onet or anything else.  But if you think it would significantly improve crypto v1 and/or onet to have something like this, we could push on it a bit.  Either way is ok with me.

jeffallen:
I do not plan to put this in v1. Closing for now.

PR #49, opened 2015-10-01
Encoding

nikkolasg:
Changed the default encoding of Int to Int32.

### DIFF
--- a/abstract/encoding.go
+++ b/abstract/encoding.go
@@ -214,10 +214,10 @@ func (de *decoder) value(v reflect.Value, depth int) error {
 	case reflect.Int:
 		var i int32
 		err := binary.Read(de.r, binary.BigEndian, &i)
-		if err != nil || int32(int(i)) != i {
-			return errors.New("int too large for this platform")
+		if err != nil {
+			return errors.New(fmt.Sprintf("Error converting int to int32 ( %v )", err))
 		}
-		v.SetInt(i)
+		v.SetInt(int64(i))
 		return err
 
 	case reflect.Bool:
--- a/abstract/encoding.go
+++ b/abstract/encoding.go
@@ -208,9 +208,9 @@ func (de *decoder) value(v reflect.Value, depth int) error {
 		}
 
 	case reflect.Int:
-		var i int64
+		var i int32
 		err := binary.Read(de.r, binary.BigEndian, &i)
-		if int64(int(i)) != i {
+		if err != nil || int32(int(i)) != i {
 			return errors.New("int too large for this platform")
 		}
 		v.SetInt(i)
@@ -288,8 +288,10 @@ func (en *encoder) value(obj interface{}, depth int) error {
 		}
 
 	case reflect.Int:
-		t := reflect.TypeOf(int64(0))
-		return binary.Write(en.w, binary.BigEndian, v.Convert(t).Interface())
+		t := reflect.TypeOf(int32(0))
+		converted := v.Convert(t) // should pass if the int is smaller than 2^32 -1
+		// otherwise it will panic
+		return binary.Write(en.w, binary.BigEndian, converted.Interface())
 
 	case reflect.Bool:
 		b := uint8(0)
--- a/abstract/encoding.go
+++ b/abstract/encoding.go
@@ -143,6 +143,8 @@ type decoder struct {
 	r io.Reader
 }
 
+var int32Type reflect.Type = reflect.TypeOf(int32(0))
+
 // Read a series of binary objects from an io.Reader.
 // The objs must be a list of pointers.
 func (e BinaryEncoding) Read(r io.Reader, objs ...interface{}) error {
@@ -292,10 +294,11 @@ func (en *encoder) value(obj interface{}, depth int) error {
 		}
 
 	case reflect.Int:
-		t := reflect.TypeOf(int32(0))
-		converted := v.Convert(t) // should pass if the int is smaller than 2^32 -1
-		// otherwise it will panic
-		return binary.Write(en.w, binary.BigEndian, converted.Interface())
+		i := int32(obj.(int))
+		if int(i) != obj.(int) {
+			panic("Int does not fit into int32")
+		}
+		return binary.Write(en.w, binary.BigEndian, i)
 
 	case reflect.Bool:
 		b := uint8(0)
--- a/abstract/encoding_test.go
+++ b/abstract/encoding_test.go
@@ -117,4 +117,16 @@ func TestEncoding(t *testing.T) {
 	err = BinaryEncoding{}.Read(&buf, &t2)
 	assert.NoError(t, err)
 	assert.Equal(t, t2, t1)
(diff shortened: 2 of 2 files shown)


abstract/encoding.go
```diff
@@ -292,8 +292,10 @@ func (en *encoder) value(obj interface{}, depth int) error {
 		}
 
 	case reflect.Int:
-		i := int64(v.Int())
-		return binary.Write(en.w, binary.BigEndian, i)
+		t := reflect.TypeOf(int32(0))
```
bford: Maybe do this as "var int32Type = ..." in package scope, so the reflect.Typeof call only needs to be done once and not for every int unmarshaled.

Also, why do you want to use reflection (reflect.Convert) instead of just marshaling it from an int32 variable and doing a simple non-reflective test to make sure the int32 variable correctly represents the originally passed int?  Reflection calls are definitely expensive (though I don't have a good sense for just how expensive), and using reflect.Convert for this just seems like one definitely unnecessary reflection call.

Finally, are we sure panic is what we want to happen if the written int is too big?  I see that panic is indeed consistent with the way some other encoding errors are handled in the encoder side, but I'm not sure that's the right way given that the encoding API actually has a graceful error-return path (which might currently never be used but users of this module might understandably *expect* it to be used to report encoding errors).

bford: On second thought, I think panic is indeed what we want or encoding errors like this where the code calling the encoder incorrectly formed the object to be encoded, so never mind that part of my comment above.  (The need for the err return is mainly to pass back errors from Write() calls cleanly, which can happen due to external conditions, e.g., disk full or network connection dropped, and not just due to internal bugs.  So there's still a legit, and different, need, for that error return path.)

But the considerations about unnecessarily using expensive reflection remain. :)

nikkolasg:
I've switched to a type assertion + added a test for the overflow. 
For me it's OK to merge !

nikkolasg:
Oups. Will have to make the change to use this function to know if we can cast to int32. Much nicer !
>func (v Value) OverflowInt(x int64) bool
>OverflowInt reports whether the int64 x cannot be represented by v's type. It panics >if v's Kind is not Int, Int8, int16, Int32, or Int64.

ineiti:
Is there also a OverflowUint?

nikkolasg:
Yes.

ineiti:
I didn’t even look whether there is also a Uint: -path. Or does Uint without a bit-specification not exist?

bford:
Didn't know there was an OverflowInt function - fine to use that if you like, but I still think it would be way more efficient and nearly as clear just to do the conversion-back-and-forth and test. :)

ineiti:
I just did a very simple test:

```
package comparison
import (
   "reflect"
   "fmt"
)

func cast(input int) {
   i32 := int32(input)
   if int(i32) == input {
      fmt.Println("is the same")
   } else {
      fmt.Println("is too big")
   }
}

func oint(input int) {
   var i32 int32
   if reflect.ValueOf(i32).OverflowInt(int64(input)) {
      fmt.Println("is the same")
   } else {
      fmt.Println("is too big")
   }
}
```

and then compiled it using

```go tool compile -S comparison.go```

The relevant part (the if) for the ```cast```-function:

```
0x0027 00039 (comparison.go:8)	MOVQL	CX, BX
0x0029 00041 (comparison.go:9)	MOVLQSX	BX, BX
0x002c 00044 (comparison.go:9)	CMPQ	BX, CX
0x002f 00047 (comparison.go:9)	JNE	275
```

And for the ```oint```-function:

```
0x001f 00031 (comparison.go:17)	MOVL	$0, AX
0x0021 00033 (comparison.go:18)	MOVL	AX, "".autotmp_0012+52(SP)
0x0025 00037 (comparison.go:18)	LEAQ	type.int32(SB), BX
0x002c 00044 (comparison.go:18)	MOVQ	BX, (SP)
0x0030 00048 (comparison.go:18)	LEAQ	"".autotmp_0012+52(SP), BX
0x0035 00053 (comparison.go:18)	MOVQ	BX, 8(SP)
0x003a 00058 (comparison.go:18)	MOVQ	$0, 16(SP)
0x0043 00067 (comparison.go:18)	PCDATA	$0, $0
0x0043 00067 (comparison.go:18)	CALL	runtime.convT2E(SB)
0x0048 00072 (comparison.go:18)	LEAQ	24(SP), BX
0x004d 00077 (comparison.go:18)	MOVQ	(BX), CX
0x0050 00080 (comparison.go:18)	MOVQ	CX, (SP)
0x0054 00084 (comparison.go:18)	MOVQ	8(BX), CX
0x0058 00088 (comparison.go:18)	MOVQ	CX, 8(SP)
0x005d 00093 (comparison.go:18)	PCDATA	$0, $0
0x005d 00093 (comparison.go:18)	CALL	reflect.ValueOf(SB)
0x0062 00098 (comparison.go:18)	MOVQ	16(SP), DX
0x0067 00103 (comparison.go:18)	MOVQ	24(SP), CX
0x006c 00108 (comparison.go:18)	MOVQ	32(SP), AX
0x0071 00113 (comparison.go:18)	MOVQ	DX, "".autotmp_0013+144(SP)
0x0079 00121 (comparison.go:18)	MOVQ	DX, (SP)
0x007d 00125 (comparison.go:18)	MOVQ	CX, "".autotmp_0013+152(SP)
0x0085 00133 (comparison.go:18)	MOVQ	CX, 8(SP)
0x008a 00138 (comparison.go:18)	MOVQ	AX, "".autotmp_0013+160(SP)
0x0092 00146 (comparison.go:18)	MOVQ	AX, 16(SP)
0x0097 00151 (comparison.go:18)	MOVQ	"".input+176(FP), BX
0x009f 00159 (comparison.go:18)	MOVQ	BX, 24(SP)
0x00a4 00164 (comparison.go:18)	PCDATA	$0, $0
0x00a4 00164 (comparison.go:18)	CALL	reflect.Value.OverflowInt(SB)
0x00a9 00169 (comparison.go:18)	MOVBQZX	32(SP), BX
0x00ae 00174 (comparison.go:18)	CMPB	BL, $0
0x00b1 00177 (comparison.go:18)	JEQ	411
```

Pretending my assembly-understanding is correct, I’d say that the simple conversion wins it hands down. Whereas I think the ```OverflowInt``` is nicer ;)

Linus

nikkolasg:
You need to remove the ```reflect.valueof(i32)```  since we already have the value in encoding, So It Does not add a lot of overhead. Look at the change.
Look at the src of overflowint, it does some unnecessary check  for the type but the check for overflow is pretty good (bit shifting) maybe we could only take that part.

ineiti:
Well, add a _benchmark.go where you do it for 10^6 times, once overflowint, once simple casting.

nikkolasg:
Ahah. Agreed ! Let's do some microbenchmarking ! (After the important stuff of course ;)

ineiti:
Yes, put it to December, or so ;)

PR #51, opened 2015-10-26
added functions that add padding which leaks only loglog(l) bits. 

LegoShrimp:
To use you just call PadGeneric(message, overhead_from_encryption), which returns a byte slice.
Where message is a byte slice, and overhead_from_encryption is the number of bytes added to message after encryption.
To remove the padding just call UnPadGeneric(padded_message) which returns a byte slice.

The testing isn't extensive enough currently.

### DIFF
--- a/ed25519/const.go
+++ b/ed25519/const.go
@@ -17,7 +17,7 @@ var prime, _ = new(big.Int).SetString("57896044618658097711785492504343953926634
 var primeOrder, _ = new(nist.Int).SetString("7237005577332262213973186563042994240857116359379907606001950938285454250989", "", 10)
 
 // cofactor of the curve, as a ModInt
-var cofactor = nist.NewInt(8, &primeOrder.V)
+var cofactor = nist.NewInt64(8, &primeOrder.V)
 
 // order of the full curve including the cofactor
 var fullOrder = new(big.Int).Mul(&primeOrder.V, &cofactor.V)
--- a/ed25519/point.go
+++ b/ed25519/point.go
@@ -260,7 +260,7 @@ func (c *Curve) Scalar() abstract.Scalar {
 	//	if c.FullGroup {
 	//		return nist.NewInt(0, fullOrder)
 	//	} else {
-	i := nist.NewInt(0, &primeOrder.V)
+	i := nist.NewInt64(0, &primeOrder.V)
 	i.BO = nist.LittleEndian
 	return i
 	//	}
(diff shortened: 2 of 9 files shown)


padding/padding.go
```diff
@@ -0,0 +1,114 @@
...
+
+}
+
+//generates padding, can be later modified if something else is better.
+//There is also probably a more efficient way to generate it.
+func GeneratePadding(p uint64) []byte {
+	b := make([]byte, p)
+	var i uint64
+	for i = 0; i < p; i++ {
+		b[i] = 51 //random choice
```
ineiti: Couldn't resist: https://xkcd.com/221/

padding/padding_test.go
```diff
@@ -0,0 +1,82 @@
...
+	z := GetZeroBits(len(pt), 16+8)
+	p := GetPaddingLen(len(pt), 16+8)
+	fmt.Println("Msg w/ static overhead: ", len(pt)+16+8, " Bits to store len: ",
+		y, " Leak Bits: ", x, " ZeroBits: ", z)
+
+	fmt.Println("Padding needed: ", p)
+	fmt.Printf("Increase in message size caused by the padding (padding+len+oh)/len: %.4f\n", ((float64(p+24) + float64(len(pt))) / float64(len(pt))))
+
+	pt1 := pt
+	//PadGeneric takes in a pt, and the overhead that encryption scheme will add. It returns a p' with the format [Amount of padding(8 bytes)][original pt][padding]
```
ineiti: I think it's more readable if you keep the original in pt (you could call it 'message'), then work with another variable down. Else it can be difficult to see the difference between 'pt' after and before here.

ineiti:
Cool - looks nice. In order to faster get an overview over your code, it is useful to:

- Comments
   - add a small introduction at the start of ```padding.go``` - goal - how to use
   - lines should be cut at 80 characters
   - add comment about the input-parameters (funny, usually I forget about the return-parameter comment, but put the input-parameter comment)
- Testing
   - add some 'small' testing of functions or group of functions. For example
      - One test over all of Get(Msg|Leak|Zero)Bits with 4 pre-defined values
      - One test over PadGeneric / UnpadGeneric
- Code
   - don't put empty lines where they don't belong (after the final return, before a closing bracket)
   - use more than 1 character for input-parameters (I prefer having words, Bryan is OK with 3-5 characters, IIRC)
   - put 'magical' parameters as constants in the beginning (like the 51 in GeneratePadding)

Small nitpicks, but it helps to dive into a code, even a starting one ;) Such small things help a lot - last time for me to understand the certificate transparency code from Google...

ineiti:
I like to do a branch first that tells what you are doing

```git checkout -b padding```

Then even if you have multiple projects at the same time, you know what pull-request does what.

PR #52, opened 2015-11-26
updated examples to use nist-library

ineiti:
The `/proof`-directory uses the `openssl`-library for tests. Replaced all `openssl` by `nist` and updated the results of the Examples.

Please verify if the output is sensible or not. Performed only copy/paste.


Daeinar:
@bford: Fyi, I think this branch is ready for merging. Thanks!

ineiti:
As discussed, please make the Travis work again.

PR #53, opened 2015-12-01
added a StringToSuite method in suites for easier retrieval of suites

ineiti:
For easy encapsulation it is useful to not only have the `.String()`-method, but to also be able to do the reverse.

### DIFF
--- a/nist/curve.go
+++ b/nist/curve.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"io"
 	"math/big"
-	//"encoding/hex"
+//"encoding/hex"
 	"crypto/cipher"
 	"crypto/elliptic"
 
@@ -54,7 +54,7 @@ func (p *curvePoint) Valid() bool {
 	// The IsOnCurve function in Go's elliptic curve package
 	// doesn't consider the point-at-infinity to be "on the curve"
 	return p.c.IsOnCurve(p.x, p.y) ||
-		(p.x.Sign() == 0 && p.y.Sign() == 0)
+	(p.x.Sign() == 0 && p.y.Sign() == 0)
 }
 
 // Try to generate a point on this curve from a chosen x-coordinate,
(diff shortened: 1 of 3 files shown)


ineiti:
Also added correct Unmarshalling of the 'infinite'-point in P256 and thus enabling that suite again.

suites/list.go
```diff
@@ -3,24 +3,38 @@
...
 type Suites map[string]abstract.Suite
 
-func (s Suites) add(suite abstract.Suite) {
-	s[suite.String()] = suite
-}
-
+// Returns a map of all suites
 func All() Suites {
 	s := make(Suites)
+	// The P256-nist implementation doesn't work for the moment
```
Daeinar: @ineiti: According to your latest note, P256 does work again and thus the above comment is redundant?

nist/curve.go
```diff
@@ -179,10 +179,20 @@ func (p *curvePoint) MarshalBinary() ([]byte, error) {
 }
 
 func (p *curvePoint) UnmarshalBinary(buf []byte) error {
-	p.x, p.y = elliptic.Unmarshal(p.c, buf)
-	if p.x == nil || !p.Valid() {
-		return errors.New("invalid elliptic curve point")
+	// Check whether all bytes after first one are 0, so we
+	// just return the initial point.
```
Daeinar: I don't think that the current code below leads to a timing attack (which might happen with byte-by-byte comparisons where secret values are involved) but why take a risk? Maybe do something like the following?
```
var c byte = 0
for _, b := range buf[1:] {
		c |= b
}
if c != 0 {
	p.x, p.y = elliptic.Unmarshal(p.c, buf)
	if p.x == nil || !p.Valid() {
		return errors.New("invalid elliptic curve point")
	}
	return nil
}
...
```
I am probably just nitpicking...

Btw. are there somewhere utility functions in dedis/crypto which do constant time comparisons, like those on https://cryptocoding.net/index.php/Coding_rules#Compare_secret_strings_in_constant_time ? Might come in handy at various points.

Daeinar:
Except for two small things, I think the changes are ok.

PR #54, opened 2015-12-04
Work-in-progress iteration on the cipher APIs

bford:
In particular:
- Point and Secret (now called Scalar) become front-end types to a single underlying Element interface.
- Suite becomes a front-end type to the semi-Go-standard 'context.Context' interface, which can maintain arbitrary key/value pairs.
- Many other smaller changes...


ineiti:
There is a git-pull hook in https://github.com/DeDiS/Coding that will automatically run goftm on the files...

ineiti:
Isn't `encoding` an old branch that has been merged and should be deleted?

bford:
Oops, indeed looks like encoding was already merged into master and should probably just be deleted.  It was the wrong base for this pull request, sorry.

ineiti:
Juste reading up on contexts - https://godoc.org/golang.org/x/net/context says that you should not put contexts in a structure, but rather pass them around. So what about https://github.com/DeDiS/crypto/blob/cipher/abstract/suite.go#L22 ?

bford:
Yes I see your point - but I think what they mean is not to put contexts in long-lived stateful structs. This Suite struct is just a "convenience front-end" to the Context it contains that unpacks and keeps cached the most important context key/value pairs used throughout the crypto library. The Suite struct is thus just a different, temporary "representation" of a Context, which can be trivially converted to and from a plain Context. 

We might even consider whether the Suite struct should actually be called Context, just distinguished by virtue of being in a different package (i.e. In the crypto library, hence a "crypto Context").  My only worry is about that causing additional confusion, especially due to the fact that one is a struct and the other is an interface. 

B

bford:
But on further thought, your question made realize that I defined the new version of the Suite type in the wrong way, which unnecessarily created a need to change all 'abstract.Suite' parameters throughout the crypto library to ave '*abstract.Suite' pointer type instead.  I just now changed it so that abstract.Suite is instead a front-end type representing a pointer to a purely private stuff that actually contains the cached info.  This way a Suite variable or parameter really is isomorphic to a Context and can be converted back and forth trivially, and callers don't need to remember to use a pointer-to-Suite in parameters instead of just a Suite.

Does this make sense and seem reasonable?

bford:
I just finished bringing this branch up-to-date with the current master branch.  Doing so left a few unresolved issues that probably need a bit more work.

nikkolasg:
Is this still revelant ?
If not, then I guess we can close it ?
If yes, then I figure this need more discussion. I'm not convinced of the usefulness of the `Context/Suite` in the crypto library. 
The `Element` abstraction can potentially be useful (already in vss / pvss) but it needs to be closely looked at, and this PR seems to fix way to much things to be adequatly reviewed ;)

ineiti:
Yes, I think we killed that `Context`-thingie. Not sure if Bryan wants to revive it...

bford:
I think the Context thingie can stay dead. :)

PR #55, opened 2015-12-05
Rename abstract.Secret to abstract.Scalar

bford:
This pull request makes only one change, but one that touches a lot of files: it renames abstract.Secret to abstract.Scalar.  I've been meaning to do this for a while, to be more consistent with standard elliptic curve crypto terminology, and in recognition of the fact that scalars are not *always* used as secrets in elliptic curve crypto (e.g., a challenge/response pair comprising a Schnorr signature is a pair of scalars neither of which is secret!).

This separates out from my 'cipher' pull request one of the more diff-noise-inducing parts of that branch, hopefully making the rest of the 'cipher' branch diffs a bit more readable.


ineiti:
Don't forget to use the git-pre-commit-hook from https://github.com/DeDiS/Coding/tree/master/git-hooks , so that all your code gets `gofmt`.

ineiti:
What is the time-frame for this pull-request to be merged? Just so that we know how we should change `cothority`, too. Else it will break on all new installations...

Is there any recommendations about how we should treat new versions? As far as I see, Go doesn't have any possibility to say *use version 1.0 of DeDiS/crypto*... So what is best practices?

bford:
No particular time-frame; assuming you agree it should be done at all, let’s plan to do it at some point when it hopefully won’t cause too much pain or delay with dependent work.

B

nikkolasg:
Closed and duplicated here : https://github.com/dedis/crypto/pull/82

PR #56, opened 2015-12-08
Suite marshal

ineiti:
This is a proof-of-concept of an auto-unmarshaling point and secret structure. 
It solves two problems with regard to unmarshaling:
- we don't know beforehand what suite we're using, so we have to write a Binary(Un)marshaler for every struct that contains abstract.Point or abstract.Secret
- as `abstract.Point` and `abstract.Secret` are interfaces, we can't use them as-is with `gob`, `protobuf`, `json` or any other marshaler, as they can't unmarshal into an interface

So I added a structure containing the interface of `abstract.Point` and `abstract.Secret` and wrote a Binary(Un)marshaler for it, so that all encoders can directly work on the structures that contain `abstract.Point` or `abstract.Secret`.

Now if you have a structure like

```
type Message struct{
   X abstract.Point
   v abstract.Secret
}
```

You don't need to write any Binary(Un)marshaler for it anymore that adds the suite and prepares the interfaces, you can just give it to any en/decoder available in Go.

This has been tested with `gob` and `json` en- and decoders.

### DIFF
--- a/abstract/group.go
+++ b/abstract/group.go
@@ -1,9 +1,7 @@
 package abstract
 
 import (
-	"bytes"
 	"crypto/cipher"
-	"fmt"
 )
 
 // XXX consider renaming Secret to Scalar?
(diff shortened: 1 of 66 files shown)


bford:
Ah, now I understand the problem you're trying to solve, instantiating appropriate Point and Scalar types when unmarshaling.  But that's a problem that's already solved by the Constructor mechanism, which both protobuf and the binary marshaling code already support, and the ciphersuites already implement.  

When marshaling or unmarshaling, you just pass the Suite as the Constructor parameter to the unmarshaler: e.g., BinaryEncoding{Constructor: suite}.Read(...)  Then when the unmarshaler encounters an uninstantiated interface variable, it invokes constructor.New(interface_type), which (when constructor is actually a Suite) will instantiate the correct Point or Scalar type.

The current Constructor mechanism isn't perfectly implemented or even entirely consistent unfortunately, but that's fixable.  There's a pull request in protobuf that's been outstanding for a while (https://github.com/DeDiS/protobuf/pull/8/files) that attempts to clean up the Constructor mechanism and make it more consistent with the binary marshaler.  And even the cleaned-up, more consistent version isn't necessarily done the right way; since then I've thought of a perhaps even better way to handle the Constructor mechanism using context.Context.  That's one of the developments my new (admittedly overambitious) "cipher" branch was intended to move toward, although even it's not quite there yet.

At any rate, I *really* don't want to marshal a variable-length ciphersuite name string with every Point or Scalar; that seems both inefficient space-wise and potentially vulnerability-prone in the way it adds new failure modes on the receive side (i.e., not just "I received an invalid Point or Scalar" but also now "I received a Point or Scalar from ciphersuite A when I was expecting one of ciphersuite B"), and the potentially subtle impacts of that when an adversary might intentionally send wrong-suite Points or Scalars kinda scare me.

ineiti:
OK, I see it's a big change, going from an interface to a structure. It's my best shot at "(Un)marshaling should just work", which can also go a long way to have the library being used in other projects. 
Some comments:

- Suite-name length: the suite can also be embedded as an index to a fixed list, or a hash of it's name, so it doesn't have to be variable-length
- BinaryEncoding using Constructor: didn't know about that one, looks nice, but still needs special treatment of the Encoding/Decoding for every structure
- Protobuf: doesn't do all types, e.g. can't send `map[string]abstract.Point`, whereas `gob` and `json` can do it
- Mixing of suites with wrong input: it's easy to test for compatible suite-types when doing operations

I have this 'it should just work' idea with regard to code (coming from Ruby), which is not always compatible with security and/or performance, I know. But seeing how many hours we spent understanding how to pass that suite to the structures, clean the code from global suites, Binary(Un)marshaling structures that should be just easy to pass I thought it's worth a try.
Also we're working with PhD students and I'd like to make it as easy as possible for them.

PR #57, opened 2015-12-10
Purbgp

LegoShrimp:
Basic functionality is done. 
There is a bug that I think is from the crypto library where it fails to authenticate data after every other attempt, but I could be wrong and I am doing something wrong that is causing it.

It can create and write a purb to a file.


nikkolasg:
Does this PR still makes sense ? Or would a separate repo for PURB and Purbgp would be a better fit ? I'm not sure it belongs to this cothority repo but rather as a standalone library that cothority could use...

LegoShrimp:
I am not sure, this is out dated a bit though. I will start a new pull request from the purbtls branch, which has everything. 
Also this is the crypto, not cothority, repo so I think it makes sense to be in this repo.

nikkolasg:
@LegoShrimp Right I just saw that, I was confused since Zenhub showed me all issues for all repos at once... 
Cool then feel free to close this one if you open another one.

PR #59, opened 2016-01-15
added NewKeyPair method

ineiti:
Useful function to get directly a new keypair without having to create the structure first. Also added a test.

### DIFF
--- a/config/key.go
+++ b/config/key.go
@@ -19,6 +19,13 @@ type KeyPair struct {
 	Secret abstract.Secret // Secret key
 }
 
+// NewKeyPair directly creates a secret/public key pair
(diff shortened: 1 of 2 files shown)


config/key.go
```diff
@@ -19,6 +19,13 @@ type KeyPair struct {
 	Secret abstract.Secret // Secret key
 }
 
+// NewKeyPair directly creates a secret/public key pair
+func NewKeyPair(suite abstract.Suite) (abstract.Secret, abstract.Point) {
```
nikkolasg: Maybe NewRandomKeyPair would be better suited, but I know that there's some minimalistic-struct-name coder out there so ...
Otherwise, looks fine for me and useful !

Issue #60, opened 2016-01-19
NewKeyPair() should return a KeyPair struct


Daeinar:
I think the recently added method `NewKeyPair()` should return a `KeyPair` struct instead of separate private and public keys, see [1]. One example where this is already realised is in the following test: [2].

In order to use things like `ConstructDeal`, see [3], one would otherwise need to generate a `KeyPair` yet again from the above private and public keys which can be clearly avoided.

[1] https://github.com/dedis/crypto/blob/master/config/key.go#L23
[2] https://github.com/dedis/crypto/blob/master/poly/deal_test.go#L32
[3] https://github.com/dedis/crypto/blob/master/poly/deal.go#L239

Daeinar:
PR: #61

Daeinar:
I just saw that this fix has consequences over here: https://github.com/dedis/cothority/blob/sda_1/lib/sda/host.go#L114

But that should be easy to fix.

Daeinar:
Done.

Liamsi:
As this is a minor change (without side effects, besides https://github.com/dedis/cothority/pull/171) I merged it right away.
Note to @nikkolasg, @ineiti, myself : Don't forget to merge https://github.com/dedis/cothority/pull/171 if you pull crypto.

PR #61, opened 2016-01-19
In NewKeyPair() return KeyPair struct instead of separate private and public keys. Addresses issue #60.

Daeinar:


Liamsi:
I think this makes more sense and is ready to be merged.

ineiti:
yes, looks better than NOT returning a `KeyPair`...

PR #62, opened 2016-02-12
BugFix : HOME folder on Windows

lbarman:
Please review line 42, hopefully this is not hindering security too much


ineiti:
Is there a way to check `Everybody`-permission in Windows? That would be the most appropriate in this case.

ineiti:
And you should use `gofmt` to make Travis happy.

PR #63, opened 2016-03-16
Make hashing a bit more flexible

Daeinar:
As the title mentions, this PR aims to make hashing via `Sum` a bit more flexible to allow concatenating data from multiple sources and run it through the hash function.


nikkolasg:
Ok for me.

bford:
Looks fine!

PR #66, opened 2016-05-19
Nist.Int.LittleEndian panics

nikkolasg:
In the LittleEndian(min,max) function, there can be a case where it panics whereas it may not have to.
Here is an example (with hex-encoded value):
```
Trying buf[:32] vs len(V.Bytes()) = 31
i.MarshalSize = 32
i.V.Bytes() :  9e86dbc411ddeab7515cfadacc4516d00d6858a3c1ec4084c05ed27c36ada6
i.M.Bytes() : 1000000000000000000000000000000014def9dea2f79cd65812631a5cf5d3ed
```
Here it panics because it tries to call `util.Reverse(buf[:32],i.V.Bytes())` where `len(i.V.Bytes()) == 31`.
The thing is that the function checks against the Modulo size and assume it is the same size for the value, but it's not always the case.

### DIFF
--- a/nist/group_test.go
+++ b/nist/group_test.go
@@ -1,8 +1,9 @@
 package nist
 
 import (
-	"github.com/dedis/crypto/test"
 	"testing"
+
+	"github.com/dedis/crypto/test"
 )
 
 var testQR512 = NewAES128SHA256QR512()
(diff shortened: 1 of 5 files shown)


bford:
I see that the LittleEndian function is already proving to be a source of complexity and subtle bugs.  It needs to die. :)

PR #67, opened 2016-06-03
Lib cosi integration

nikkolasg:
This PR is to integrate cosi as a additional component of the crypto library in order to use cosigning in a friendly way !


cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+		return errors.New("No random secret computed in this cosi")
+	}
+	if c.challenge == nil {
+		return errors.New("No challenge computed in this cosi")
+	}
+
+	// hash the private key
+	hash := sha512.New()
+	privKeyBuff := secretToSlice(c.private)
+	//privKeyBuff := c.private.(*nist.Int).LittleEndian(32, 32)
```
ineiti: Is this still used?

ineiti:
```
$ golint
cosi.go:318:2: don't use leading k in Go names; var kBuff should be buff
cosi.go:389:6: type name will be used as cosi.CosiMask by other packages, and that stutters; consider calling this Mask
cosi.go:410:1: exported method CosiMask.AllEnabled should have comment or be unexported
cosi.go:417:1: comment on exported method CosiMask.Set should be of the form "Set ..."
cosi.go:490:1: exported method CosiMask.MarshalBinary should have comment or be unexported
cosi.go:496:1: exported method CosiMask.UnmarshalBinary should have comment or be unexported
cosi.go:500:1: exported method CosiMask.Aggregate should have comment or be unexported
```

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+}
+
+// NewCosiMask returns a new mask to use with the cosigning with all cosigners enabled
+func NewCosiMask(suite abstract.Suite, publics []abstract.Point) *CosiMask {
+	// Start with an all-disabled participation mask, then set it correctly
+	cm := &CosiMask{
+		publics: publics,
+		suite:   suite,
+	}
+	cm.mask = make([]byte, cm.MaskLen()) //(len(publics)+7)>>3)
```
ineiti: What is that comment saying?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	"github.com/dedis/crypto/abstract"
+	//own "github.com/nikkolasg/learning/crypto/util"
+)
+
+// Cosi is the struct that implements the basic cosi.
+type Cosi struct {
+	// Suite used
+	suite abstract.Suite
+	// publics is the list of public keys used for signing
+	publics []abstract.Point
```
ineiti: I think this is not used - if that's true, remove it here, as it is available in CosiMask.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	// or not. All code regarding the mask is directly inspired from
+	// github.com/bford/golang-x-crypto/ed25519/cosi code.
+	mask *CosiMask
+	// the longterm private key we use during the rounds
+	private abstract.Secret
+	// the message being co-signed
+	message []byte
+	// timestamp of when the announcement is done (i.e. timestamp of the four
+	// phases)
+	timestamp int64
```
ineiti: What is timestamp used for? I can't find it anywhere used...

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+
+	return nil
+}
+
+// VerifySignature is the method to call to verify a signature issued by a Cosi
+// struct. Publics is the WHOLE list of publics keys, the mask at the end of the
+// signature will take care of removing the indivual public keys that did not
+// participate
+func VerifySignature(suite abstract.Suite, publics []abstract.Point, message, sig []byte) error {
```
ineiti: I propose the following structure of the file:
- structures
- New-methods for the structures
- public methods on the structues
- private methods on the structures
- public library-methods
- private library-methods

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+	if c.random == nil {
+		return errors.New("No random secret computed in this cosi")
+	}
+	if c.challenge == nil {
+		return errors.New("No challenge computed in this cosi")
+	}
+
+	// hash the private key
+	hash := sha512.New()
```
ineiti: Can you use `Ed25519ToPublic` here?

cosi/crypto.go
```diff
@@ -0,0 +1,76 @@
...
+// NOTE: Similar to secretToSlice, it will modulo'd the secret if not already.
+func sliceToSecret(suite abstract.Suite, buffer []byte) abstract.Secret {
+	s := suite.Secret().(*nist.Int)
+	s.SetLittleEndian(buffer)
+	return s
+}
+
+// Ed25519ToPublic will transform a ed25519 scalar to a ed25519 EDDSA formated public key using
+// the digest + prune transofrmation
+func Ed25519ToPublic(suite abstract.Suite, s abstract.Secret) abstract.Point {
```
ineiti: Public methods on top, private on bottom.

cosi/cosi_test.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	// pass them up to the root
+	_, err := root.Response(responses)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Response phase failed:%v", err)
+	}
+	return root, children, nil
+}
+
+/*func TestCosiEd25519(t *testing.T) {*/
```
ineiti: This has to go away...

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	response abstract.Secret
+	// aggregateResponses is the aggregated response from the children + our own
+	aggregateResponse abstract.Secret
+}
+
+// NewCosi returns a new Cosi struct given the suite, the longterm secret, and
+// the list of public keys. If some signers were not to be participating, you
+// have to set the mask using `SetMask` method. By default, all participants are
+// designated as participating.
+func NewCosi(suite abstract.Suite, private abstract.Secret, publics []abstract.Point) *Cosi {
```
ineiti: Some thoughts about `public`: I think it's only used in the root-node for creating the challenge. But when calling `NewCosi`, we don't know yet who will be signing or not, and we can't change the mask afterwards, as `publics` is private. And CoSi shouldn't care how this mask is generated, anyway, this is the protocol's or service's task.

I propose we remove `publics` from `NewCosi` and give a `CosiMask` to `CreateChallenge`.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/dedis/crypto/abstract"
+	//own "github.com/nikkolasg/learning/crypto/util"
+)
+
+// Cosi is the struct that implements the basic cosi.
+type Cosi struct {
```
bford: Cosi -> CoSi, to be consistent with the capitalization in the paper?

bford: Also, please make the comment more descriptive.  Yes, it's obviously a struct, but what's "the basic cosi"?  The struct is one-per-what?  One per protocol instance?  One per protocol run?  What?  Also summarize briefly how to use it: the way(s) to create one, the life-cycle of the struct, etc.  If the caller is supposed to invoke methods in sequence for the four stages mentioned in the module godoc, then no need to duplicate all that info, but at least list the 4 specific function names the caller is expected to call for each of the 4 stages, etc.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	// Suite used
+	suite abstract.Suite
+	// publics is the list of public keys used for signing
+	publics []abstract.Point
+	// mask is the mask used to select which signers participated in this round
+	// or not. All code regarding the mask is directly inspired from
+	// github.com/bford/golang-x-crypto/ed25519/cosi code.
+	mask *CosiMask
+	// the longterm private key we use during the rounds
+	private abstract.Secret
```
bford: Which private key?  i.e., who is "we"?  Is the identity of who "we" are (e.g., an index within a list of public keys) indicated in the cosi struct in anyway, or does it implicitly come from somewhere else?

cosi/cosi.go
bford: Why is this needed?  Isn't this specific to the timestamp protocol, which should be higher-level and independent of the basic cosi signing mechanism?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	// globalComitment is the global commit of all participants
+	//passed down in the challenge phase
+	globalCommitment abstract.Point
+	// challenge holds the challenge for this round
+	challenge abstract.Secret
+	// response is our own computed response
+	response abstract.Secret
+	// aggregateResponses is the aggregated response from the children + our own
+	aggregateResponse abstract.Secret
+}
```
bford: In the above, I think the fields should be organized so it's clear which fields are "public global information" - i.e., common information known to everyone, and which represent fields known only to a particular signer (e.g., private key, random, commitment).

cosi/cosi.go
bford: rename private -> priKey or secKey, and publics -> pubKeys?  Or perhaps just pubKey[], letting the type system make it clear that it's an *array* of public keys?  I personally find code that says 'pubKey[i]' more clear than 'pubKeys[i]', because the former obviously means "public key [number] i" whereas the English transliteration of the latter is "public keys i" and it's not as obvious what that means.  So I tend to avoid "pluralizing" variable names just because they happen to be an array.  But I admit that's personal taste.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+		private: private,
+		publics: publics,
+	}
+	// Start with an all-disabled participation mask, then set it correctly
+	cos.mask = NewCosiMask(suite, publics)
+	return cos
+}
+
+// Announcement holds only the timestamp for that round
+type Announcement struct {
```
bford: Make the comment docs more descriptive.  What is this and the other structs actually used for?  Are these protobuf-encodable structs, for example?  How are they supposed to be used?  Do they need to be public?  (i.e., what is the caller of this module supposed to do with them?)

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	Timestamp int64
```
bford: again not sure what a timestamp is doing here; shouldn't that be an application-specific optional component of a message?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+}
+
+// Commitment sends it's own commit Vi and the aggregate children's
+// commit V^i
+type Commitment struct {
+	Commitment     abstract.Point
```
bford: Add (brief) clarifying comments to each struct field.  e.g., is this purely the *local* commitment from this node itself, or is it an aggregate of any kind?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	ChildrenCommit abstract.Point
+}
```
bford: Also, does each node really need to send up its children's aggregate commitment separately from its own?  At least in the current protocol where we have to restart if anyone drops out in the middle of a round, it's not clear to me why we can't just pass up the aggregate commit alone (and aggregate response later).

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+// CreateAnnouncement creates an Announcement message with the timestamp set
+// to the current time.
+func (c *Cosi) CreateAnnouncement() *Announcement {
+	now := time.Now().Unix()
+	c.timestamp = now
+	return &Announcement{now}
+}
+
+// Announce stores the timestamp and relays the message.
+func (c *Cosi) Announce(in *Announcement) *Announcement {
```
bford: Relays the message to whom?  And it doesn't look like this method actually relays it to anyone; who actually does the relaying?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+func (c *Cosi) CreateCommitment(r io.Reader) *Commitment {
+	c.genCommit(r)
+	return &Commitment{
+		Commitment: c.commitment,
+	}
+}
+
+// Commit creates the commitment / secret + aggregate children commitments from
+// the children's messages.
+func (c *Cosi) Commit(r io.Reader, comms []*Commitment) *Commitment {
```
bford: comms[] -> subComms[] or subCom[]?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+
+// Announce stores the timestamp and relays the message.
+func (c *Cosi) Announce(in *Announcement) *Announcement {
+	c.timestamp = in.Timestamp
+	return in
+}
+
+// CreateCommitment creates the commitment out of the random secret and returns
+// the message to pass up in the tree. This is typically called by the leaves.
+func (c *Cosi) CreateCommitment(r io.Reader) *Commitment {
```
bford: This method looks redundant with Commit(): why is a special function needed that works only for the leaves?  Why not just use Commit(), with the second parameter nil?  (I think that'll even "just work" in that the for loop over "range comms" should just do nothing since len(nil) == 0.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+		// Add commitment of one child
+		childVHat = childVHat.Add(childVHat, com.Commitment)
+		// add commitment of it's children if there is one (i.e. if it is not a
+		// leaf)
+		if com.ChildrenCommit != nil {
+			childVHat = childVHat.Add(childVHat, com.ChildrenCommit)
+		}
+	}
+	// add our own commitment to the global V_hat
+	c.aggregateCommitment = c.suite.Point().Add(childVHat, c.commitment)
```
bford: Why not initialize childVHat to c.commitment (our local commitment) in the first place up above, rather than initializing it to null and then having to Add it separately later?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+		Commitment:     c.commitment,
+	}
+
+}
+
+// CreateChallenge creates the challenge out of the message it has been given.
+// This is typically called by Root.
+func (c *Cosi) CreateChallenge(msg []byte) (*Challenge, error) {
+	// H( Commit || AggPublic || M)
+	hash := sha512.New()
```
bford: The hash function needs to be parameterized, just like the suite is.  Maybe that should be part of a "signing scheme" interface providing signing-scheme-specific methods for creating challenges etc.?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+
+	pb, err := c.aggregateCommitment.MarshalBinary()
```
bford: Why not just use MarshalTo(hash)?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+func (c *Cosi) Challenge(ch *Challenge) *Challenge {
+	c.challenge = ch.Challenge
+	c.globalCommitment = ch.GlobalCommitment
+	return ch
+}
+
+// CreateResponse is called by a leaf to create its own response from the
+// challenge + commitment + private key. It returns the response to send up to
+// the tree.
+func (c *Cosi) CreateResponse() (*Response, error) {
```
bford: Again, why a separate function just for leaves?  Should be unnecessary.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	return &Response{
+		Response:     c.response,
+		ChildrenResp: aggregateResponse,
+	}, nil
+}
+
+// Signature returns a signature in the like the EdDSA signature format plus the
+// bitmask at the end.
+// Sig = AggCommit || AggResponse || bitmask
+func (c *Cosi) Signature() []byte {
```
bford: Is this callable at any node or only at the root?

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	final := make([]byte, 64+c.mask.MaskLen())
+	copy(final[:], sigR)
+	copy(final[32:64], sigS)
+	copy(final[64:], c.mask.mask)
+	return final
+}
+
+// GetAggregateResponse returns the aggregated response that this cosi has
+// accumulated.
+func (c *Cosi) GetAggregateResponse() abstract.Secret {
```
bford: Remember, the Go convention is not to prefix the names of getters with "Get" - just name it AggregateResponse() etc.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+
+// GetCommitment returns the commitment generated by this CoSi (not aggregated).
+func (c *Cosi) GetCommitment() abstract.Point {
+	return c.commitment
+}
+
+// VerifyResponses verifies the response this CoSi has against the aggregated
+// public key the tree is using.
+// Reconstruct the AggCommit
+func (c *Cosi) VerifyResponses(aggregatedPublic abstract.Point) error {
```
bford: This can be called at any node, not just the root, correct?  Clarify in the documentation.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	left := suite.Point().Add(kA, sB)
+
+	if !left.Equal(aggCommit) {
+		return errors.New("Signature invalid")
+	}
+
+	return nil
+}
+
+// genCommit generates a random secret vi and computes it's individual commit
```
bford: it's -> its

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	expandedPrivKey := h[0:32]
+	expandedPrivKey[0] &= 248
+	expandedPrivKey[31] &= 127
+	expandedPrivKey[31] |= 64
+	expandedPrivKeyInt := sliceToSecret(c.suite, expandedPrivKey)
+
+	// resp = random - challenge * privatekey
+	// i.e. ri = vi + c * xi
+	resp := c.suite.Secret().Mul(expandedPrivKeyInt, c.challenge)
+	c.response = resp.Add(c.random, resp)
```
bford: You should set c.random to nil here, after using it!  Just to make absolutely sure it only ever gets used once (security paranoia).

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	// hash the private key
+	hash := sha512.New()
+	privKeyBuff := secretToSlice(c.private)
+	//privKeyBuff := c.private.(*nist.Int).LittleEndian(32, 32)
+
+	hash.Write(privKeyBuff)
+	h := hash.Sum(nil)
+
+	// prune it up
+	expandedPrivKey := h[0:32]
```
bford: Again these details need to be parameterized or moved to ed25519-specific stuff somewhere; I assume you already know that.

cosi/cosi.go
```diff
@@ -0,0 +1,502 @@
...
+	// i.e. ri = vi + c * xi
+	resp := c.suite.Secret().Mul(expandedPrivKeyInt, c.challenge)
+	c.response = resp.Add(c.random, resp)
+	// no aggregation here
+	c.aggregateResponse = c.response
+	return nil
+}
+
+// CosiMask holds the mask utilities
+type CosiMask struct {
```
bford: No need to prefix type names (or anything!) with "Cosi" in a module named "cosi".  Just call it Mask.

cosi/cosi.go
nikkolasg: Yes, I removed it.

cosi/cosi.go
nikkolasg: It's not really about failure but about mischief, if a node does not generate correctly its response, we can check only its individual response (by having its individual commit). Instead of checking for the whole subtree if the response is correct, we can check also for direct children's reponse too. But if you think that's not a good use case, I'm all for removing it also, it's not so important because if a wrong response has been generated, we need to deal with it anyway by restarting I guess so..

cosi/cosi.go
nikkolasg: Removed the announcement part totally.

cosi/cosi.go
nikkolasg: 1) I wanted to make clear by the methods in this API what should you be doing, (should you be *starting* a new phase or should you just pass up/down). There's practically no code duplication since the real logic is the the `genCommit` function.
2) I'm always reluctant of these magic parameters where you can give *nil* or not: I always prefer explicit rather than implicit. Golang does not have optional parameters, and I don't think it's a good idea to try to emulate that.

I'm doing the change for using the new ed25519 for the moment. If you still think that makes sense to get only one function, then I'll do it right afterward :) (so please comment if you have time !)

cosi/cosi.go
nikkolasg: @ineiti :
> CoSi shouldn't care how this mask is generated, anyway, this is the protocol's or service's task.

Well, not really in my opinion because CoSi *is* the one generating the mask and the mask *is part of* its protocol... But it does not change anything for CoSi which cosigners are missing and so on.
Actually, in the comments it says, 
```
// If some signers were not to be participating, you
// have to set the mask using `SetMask` method. By default, all participants are
// designated as participating.
```
Then I forgot to implement the `SetMask` method , which I'm doing right now. I find this solution more elegant than having to specify a mask explicitely *if you don't need it*.
(1)
```
// No missing cosigners
c := NewCosi(suite,secret,NewCosiMask(suite,publics))
...
// Missing cosigners
mask := NewCosiMask(suite,publics)
mask.SetBitMask(4,false)
c := NewCosi(suite,secret,mask)
```
vs (2)
```
// No missing cosigners
cosi := NewCosi(suite,secret,publics)
...
// Missing cosigners
mask := NewCosiMask(suite,publics)
mask.SetBitMask(4,false)
c := NewCosi(suite,secret,publics)
c.SetMask(mask)
``` 
Or (2- bis)
```
// No missing cosigners
cosi := NewCosi(suite,secret,publics)
...
// Missing cosigners
mask := NewCosiMask(suite,publics)
mask.SetBitMask(4,false)
c := NewCosiWithMask(suite,secret,mask)
```

One approach is shorter for the standard behavior (2), and the other one is reducing the code for the exception behavior (1), but if you use (2-bis), then it's the same amount. I would go for the (2-bis) solution, but as democracy imposes us a majority, let the majority decide :)

cosi/cosi.go
nikkolasg: Well, I've tested Cosi using edwards for example, it did not work, sometimes it throws a panic saying the aggregateCommitment is not valid. Using the (new formatted) ed25519 suite, it works all the time... I'll have to debug this into more details...
But regarding the rest, why should allow the user to use a different hash function than the one intended for the algorithm ? Allowing more choice than necessary is more hazardous, more prone to being less secure. For example, a user could give a md5 hash function... See [crypto coding's rule](https://cryptocoding.net/index.php/Coding_rules#Prevent_confusion_between_secure_and_insecure_APIs) of Aumasson for an example. 
But having a "signing scheme" that only gives safe (and standards) choice could be a good idea yes. Since it's not clear yet how to do it, I prefer to leave it as-is for this PR.

cosi/cosi.go
nikkolasg: nice shorthand indeed !

cosi/cosi.go
nikkolasg: see my response for commits.

cosi/cosi.go
nikkolasg: Added a note in the comments.

cosi/cosi.go
nikkolasg: Yes, I know the convention but for the case of `GetChallenge()` and `GetCommitment()` I can't remove the -Get- part because a function already exists without the -Get-.  I wanted to stay consistent with the other getters here also and not having a -Get-.  Maybe the name of `Challenge` and `Commitment` is not adequate then, do any of you have other ideas ?
`ChallengePhase()` / `CommitmentPhase` ? Not elegant but explicit ...

cosi/cosi.go
nikkolasg: Ok since this is the second argument regarding whether we should send the individual response or not in the message, I'll remove the *individual response* in the packets and then I'll be able to put the response to nil here also.

cosi/cosi.go
nikkolasg: @bford Regarding the taste, I tend to go on the opposite side and I support my choice like this:
By *reading* the function signature in the API go doc, it seems more natural to read that `functionXXX(publics []YYY)` expects **many** YYY,i.e. a *list* of public keys YYY; rather than `functionXXX(public []YYY)` where it *can* (not necessarily) be interpreted as:  the function needs **one**  public key which *consists of* a list of YYY. 
But inside the function, it's true that when you access a public key with an index, it's easier to read as you said...

cosi/cosi.go
nikkolasg: I ordered them from the global common information to *personal* informations.

cosi/cosi.go
nikkolasg: removed all structs.

PR #68, opened 2016-06-07
(1/3) Formatting ed25519 to the standards -> Add new suite to ed25519

nikkolasg:
This PR is to make `edwards/ed25519` scalars being formatted to the standards. More specifically, the secret now stores the hashed private key with bit twiddling, as NaCL is doing (see [here](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/)).
This PR make part of our crypto library adhering to the standards and also being able to generate and verify signature in the EdDSA way !
As a bonus, there's now also a `EdDSASign` and `EdDSAVerify` function included. Thoses are tested against the tests vectors included in the [RFC](https://tools.ietf.org/html/draft-josefsson-eddsa-ed25519-02).

This PR has to be merged / or refused (!) before we can go on with #67 .


edwards/ed25519/const.go
```diff
@@ -4,22 +4,20 @@
...
-var primeOrder, _ = new(nist.Int).SetString("7237005577332262213973186563042994240857116359379907606001950938285454250989", "", 10)
+var primeOrder = newSecretFromString("7237005577332262213973186563042994240857116359379907606001950938285454250989", 10)
 
 // cofactor of the curve, as a ModInt
-var cofactor = nist.NewInt(8, &primeOrder.V)
+var cofactor = newSecretFromInt64(8)
 
 // order of the full curve including the cofactor
-var fullOrder = new(big.Int).Mul(&primeOrder.V, &cofactor.V)
+// XXX NON USED
```
ineiti: If it's really unused, remove it.

edwards/ed25519/eddsa.go
```diff
@@ -0,0 +1,107 @@
...
+// Picked() so it can use the prefix, the right part of the hashing done in
+// Pick.
+// Takes
+// - suite is the suite to use
+// - a secret to use to sign the message
+// - A corresponding public key
+// - msg message to sign
+// NOTE: Code taken from the Python implementation from the RFC
+// https://tools.ietf.org/html/draft-josefsson-eddsa-ed25519-02
+func EdDSASign(suite abstract.Suite, a abstract.Secret, A abstract.Point, msg []byte) ([]byte, error) {
```
ineiti: Should we take a generic suite? Or does this only work with the eddsa-suite?
What about taking the `suite.Hash` for hashing?

nikkolasg: Right. it only works with EdDSA suite (which I should create now). And I will change the hashing function sha512.

edwards/ed25519/eddsa.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"crypto/sha512"
+	"errors"
+	"fmt"
+
+	"github.com/dedis/crypto/abstract"
+)
+
+// EdDSASign will return a EdDSA signature using Ed25519. The secret must be
```
ineiti: // EdDSASign returns a EdDSA signature using Ed25519

edwards/ed25519/eddsa.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// return R || s
+	var sig [64]byte
+	copy(sig[:], Rbuff)
+	copy(sig[32:], sBuff)
+
+	return sig[:], nil
+}
+
+// EdDSAVerify verify a signature issued by EdDSASign
```
ineiti: EdDSAVerify verifies a signature issued by EdDSASign

edwards/ed25519/secret.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (s *secret) Pick(rand cipher.Stream) abstract.Secret {
+	pre := random.Bytes(32, rand)
+	expandedSecretKey := sha512.Sum512(pre)
+	expandedSecretKey[0] &= 0xf8
+	expandedSecretKey[31] &= 0x3f
+	expandedSecretKey[31] |= 0x40
+
+	/* base := big.NewInt(2)*/
```
ineiti: Unused comment

edwards/ed25519/secret_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if !D.Equal(AMB) {
+		t.Error("Wrong for sub")
+	}
+}
+
+// EdDSATestVectors taken from the RFC with Ed25519 test vectors with compressed
+// public keys
+// https://tools.ietf.org/html/draft-josefsson-eddsa-ed25519-02#section-6
+var EdDSATestVectors = []struct {
```
ineiti: +1

edwards/ed25519/eddsa.go
ineiti: So you can take away the `suite` here?

PR #69, opened 2016-06-07
(2/3) Formatting ed25519 to the standards -> Replace ed25519 suite

nikkolasg:
This PR is similar to #68 EXCEPT that it does NOT try to save the existing suite, but rather CHANGE it to use the standards. 
For me this would be the best approach as we would only have *one* suite for Ed25519 that is following the standards formats (instead of 2 in `ed25519` where only one follows the standards OR an external struct to use the standards, which would still not be the defaults, as in #71 ).
Other pros:
+ we can exchange keys with NaCL at least, and others!) 
+ Additional security benefits with the Hashing (uniform looking output even with broken PRNG)
+ Additional security benefits with the bit tweaklings (avoid subgroup attacks)

Sorry for the mess of having two PR for the almost the same thing but since Bryan's not there, I prefer to propose both options and then choose one and remove the other.


ineiti:
Will you add the `struct`-approach, or do you want me to do that?

PR #71, opened 2016-06-08
(3/3) Formatting ed25519 to the standards -> New struct generating correct format

nikkolasg:
Same goal as #68 and #69 but here we have a `EdDSA` struct that can create secret/point of the correct format. The original suite is kept as-is.

### DIFF
--- a/edwards/ed25519/eddsa.go
+++ b/edwards/ed25519/eddsa.go
@@ -1,0 +1,154 @@
+package ed25519
+
+import (
+	"crypto/cipher"
+	"crypto/sha512"
+	"errors"
+	"fmt"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/nist"
+	"github.com/dedis/crypto/random"
+	"github.com/dedis/crypto/util"
+)
+
+type EdDSA struct {
+	Secret abstract.Secret
+	prefix []byte
+	Public abstract.Point
+}
+
+func NewEdDSA(randStream cipher.Stream) *EdDSA {
+	// XXX if this PR is accepted, then change this line to iterate as long as
+	// we have 0x00...00 slice buffer. See issue #70
+	randSlice := random.Bytes(32, randStream)
+	hash := sha512.New()
+	hash.Write(randSlice)
+	// XXX Should we copy it to s and prefix ?
+	expanded := hash.Sum(nil)
+	s := expanded[0:32]
+	prefix := expanded[32:]
+	s[0] &= 0xf8
+	s[31] &= 0x7f
+	s[31] |= 0x40
+	// giving the reverse as nist.Int expect big endian
+	var reverse [32]byte
+	util.Reverse(reverse[:], s)
+	suite := NewAES128SHA256Ed25519(false)
+	// We can't use Secret().Pick() because it uses random.int which tests if
+	// the results is < Mod, which is never true (because of the bit tweaks)
+	sec := suite.Secret().(*nist.Int)
+	sec.SetLittleEndian(s)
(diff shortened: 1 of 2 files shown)


ineiti:
Will probably not pass `golint` - needs documentation on exported methods.

edwards/ed25519/eddsa.go
```diff
@@ -0,0 +1,154 @@
...
+	Public abstract.Point
+}
+
+func NewEdDSA(randStream cipher.Stream) *EdDSA {
+	// XXX if this PR is accepted, then change this line to iterate as long as
+	// we have 0x00...00 slice buffer. See issue #70
+	randSlice := random.Bytes(32, randStream)
+	hash := sha512.New()
+	hash.Write(randSlice)
+	// XXX Should we copy it to s and prefix ?
```
ineiti: Why copy?

edwards/ed25519/eddsa.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/nist"
+	"github.com/dedis/crypto/random"
+	"github.com/dedis/crypto/util"
+)
+
+type EdDSA struct {
+	Secret abstract.Secret
+	prefix []byte
```
ineiti: Shouldn't the `prefix` be public? How and where do we use it?

edwards/ed25519/eddsa.go
```diff
@@ -0,0 +1,154 @@
...
+func (ed *EdDSA) Sign(msg []byte) ([]byte, error) {
+	sec := ed.Secret
+	hash := sha512.New()
+	hash.Write(ed.prefix)
+	hash.Write(msg)
+
+	suite := NewAES128SHA256Ed25519(false)
+
+	// deterministic random secret and its commit
+	r := suite.Secret().(*nist.Int)
```
ineiti: `SetBytes` would be a nice addition for `Secret`.

edwards/ed25519/eddsa.go
```diff
@@ -0,0 +1,154 @@
...
+	hA := suite.Point().Mul(public, h)
+	RhA := suite.Point().Add(R, hA)
+
+	if !RhA.Equal(S) {
+		return errors.New("Recontructed S is not equal to signature")
+	}
+	return nil
+}
+
+type fixedStream struct {
```
ineiti: As this is only used in the testing-file, I'll put it there.

ineiti:
Looks good - definitively my favourite of the three solutions. If this will be chosen in the end, it might be good to add a `SetBytes` to `Secret`.

With regard to standards: the ietf-document this is taken from is still in 'draft' - so not really a standard (yet). I didn't find a final one on EdDSA-signatures.

edwards/ed25519/eddsa.go
nikkolasg: Not really. It's only used in EdDSA sign, because it's the algorithm definition.

edwards/ed25519/eddsa.go
nikkolasg: Well we would still have to put them in big endian before, so either 
```
buffBE := util.Reverse(buffBE,buff)
secret.SetBytes(buffBE)
```
or
```
sNist := secret.(*nist.Int)
sNist.SetLittleEndian(buffer)
```
Same amount of code and for me it would be confusing to have `UnmarshalBinary(b []byte)` and at the same time `SetBytes(b []byte)`. There's no clear separate logic between the two...

For me one useful addition would be to specify the endianess on a nist.Int and having `Marshal/UnmarshalBinary` return the correct endianess.

Issue #70, opened 2016-06-08
sha3.Shake128 with abstract.NoKey NOT random


nikkolasg:
Found a unusual behavior when debugging edwards/ed25519 suites: 
If you create a `rand := cipher.Stream` from those suites (using sha3.Shake128), and then retrieve a slice of bytes from it using `random.Bytes(32,rand)`, the slice of bytes are always zero the first 6 times. Having discussed with @Daeinar , it may be related to the size of the state of sha3.Shake128 but as I'm really not familiar with those sponge cipher (yet!), I can't say.
It's definitely not a correct behavior as every output should look completely random whatever the input is.
This faulty behavior is kinda _hidden_ in `nist.Int.Pick(rand)` where this function calls `random.Int` which iterates as long as the number returned fulfills `n != 0`.

jeffallen:
This is no longer an issue since XOF replaced Cipher.

PR #72, opened 2016-06-09
Final version: Ed25519 with EdDSA compatibility

nikkolasg:
Finally after some talks, this is the right PR where we do the compatibility with EdDSA. I will delete #68 #69 #71 .


nist/int.go
```diff
@@ -37,8 +39,9 @@ var two = big.NewInt(2)
 // whose target is assumed never to change.
 //
 type Int struct {
-	V big.Int  // Integer value from 0 through M-1
-	M *big.Int // Modulus for finite field arithmetic
+	V          big.Int          // Integer value from 0 through M-1
+	M          *big.Int         // Modulus for finite field arithmetic
+	endianness binary.ByteOrder // Endianness considered for this int
```
bford: Maybe rename the element 'bo' or 'byteOrder' since that seems more self-explanatory (and more like a word) than "endianness"?

And perhaps just make it public (e.g., 'BO'), since the other elements are already public and I don't see any obvious harm/risk if this one is public too?

Finally, I'm not totally sure 'binary.ByteOrder' is the appropriate type to use here, since that's a whole interface-pointer that could in principle point to any object implementation that interface, but here the semantics are that we *need* it to be 'BigEndian' or 'LittleEndian' and nothing else.  i.e., if the caller creates their own 'myLittleEndian' object satisfying the ByteOrder interface and behaving exactly like LittleEndian, but a different pointer from LittleEndian, things will break, which violates the way interfaces are "supposed" to work.  So here I think the cleaner approach would actually be to declare a local ByteOrder type, e.g.:

type ByteOrder bool
const (
   LittleEndian ByteOrder = false
   BigEndian ByteOrder = true
)

...and just use that.

nikkolasg: All right for the public. I also hesitated creating our own type but then I chose the *standard* way. But since as you noted it might break the thing, I now think that's also best to declare our own...

nist/int.go
```diff
@@ -314,21 +313,17 @@ func (i *Int) UnmarshalBinary(buf []byte) error {
 	if len(buf) != i.MarshalSize() {
 		return errors.New("Int.Decode: wrong size buffer")
 	}
-	i.V.SetBytes(buf)
+	if i.endianness == binary.LittleEndian {
+		i.SetLittleEndian(buf)
+	} else {
+		i.V.SetBytes(buf)
```
bford: SetBytes should be sensitive to the current byte-order, so this if-else shouldn't be needed here; it should happen in SetBytes instead.

nist/int.go
```diff
@@ -346,6 +341,21 @@ func (i *Int) BigEndian(min, max int) []byte {
...
 
+// Set value to a number represented by a byte string.
+// Endianness depends on the endianess set in i.
+func (i *Int) SetBytes(a []byte) *Int {
+	i.V.SetBytes(a).Mod(&i.V, i.M)
+	return i
+}
+
+// Set value to a number represented in a little-endian byte string.
+func (i *Int) SetLittleEndian(a []byte) *Int {
```
bford: This shouldn't be needed anymore, or at least doesn't need to be public; it should get merged into SetBytes right?

nist/int.go
```diff
@@ -416,3 +434,14 @@ func (i *Int) HideDecode(buf []byte) {
 	i.V.SetBytes(buf)
 	i.V.Mod(&i.V, i.M)
 }
+
+// SetEndianess will set the endianness of the value to end.
+func (i *Int) SetEndianness(end binary.ByteOrder) {
```
bford: Rename to SetByteOrder?

nist/int.go
nikkolasg: I also think that this should not be needed anymore but when I asked you you said "let's keep it" .. :) I'm all for removing that !

nist/int.go
nikkolasg: Since I'm going for a Public field as suggested above, we can even remove that method

nist/int.go
nikkolasg: Well, if `MarshalBinary` returns a slice in LittleEndian format, then `UnmarshalBinary` on the same `Int` should expect a slice in LittleEndian format ... I thought it was the whole point of the thing,  to be able to use  `Secret` which `Marshal/Unmarshal` in LittleEndian format... ?

random/rand.go
```diff
@@ -86,6 +86,24 @@ func Bytes(n int, rand cipher.Stream) []byte {
 	return b
 }
 
+// NonZeroBytes calls Bytes as long as it gets a slice full of '0's.
+// This is needed when using suite.Cipher(abstract.NoKey)
+// because the first 6 iterations returns 0000...000 as
+// bytes for edwards & ed25519 cipher.
+// XXX Issue reported in https://github.com/dedis/crypto/issues/70
+func NonZeroBytes(n int, rand cipher.Stream) []byte {
```
nikkolasg: I've put it here but I'm still hesitating whether this should go into `util` or stg like that . At least two packages will need this function.

nist/int.go
bford: I probably meant that I want to keep the *ability* to get little-endian behavior out of nist.Int - but that doesn't mean we need to keep the SetLittleEndian *method*. :)

nist/int.go
bford: Fine with me.  No real need for getters/setters if there's no important internal abstraction to be protected.

random/rand.go
ineiti: So there is no way of having the ciphers behaving in a non-zero way?

abstract/group.go
```diff
@@ -51,6 +51,9 @@ type Secret interface {
 
 	// Set to a fresh random or pseudo-random secret
 	Pick(rand cipher.Stream) Secret
+
+	// SetBytes will take bytes and create a secret out of it
+	SetBytes([]byte) Secret
```
ineiti: Will we also have a `Bytes()`-method?

ed25519/suite.go
```diff
@@ -37,8 +42,32 @@ func (s *suiteEd25519) New(t reflect.Type) interface{} {
...
+
 // Ciphersuite based on AES-128, SHA-256, and the Ed25519 curve.
 func NewAES128SHA256Ed25519(fullGroup bool) abstract.Suite {
 	suite := new(suiteEd25519)
 	return suite
 }
+
+// NewEd25519Scalar returns the scalar derived from the stream and the right
+// half of the hash during the derivation of the scalar.
+func NewEd25519Scalar(stream cipher.Stream) ([]byte, []byte) {
```
ineiti: Will this also be used elsewhere? If no, I'd include it in `NewKey`.

abstract/group.go
nikkolasg: 1) There's no immediate reason to put it in this PR since it won't be used (But still can!)
2) Do you have the exact functionality of this function in mind ? 
 * `Marshal/Unmarshal` will give/read the representation of a secret correctly formatted and everything
 * `SetBytes` is supposed to accept any bytes input and squeeze that into a scalar. So I guess `Bytes()` is supposed to give back the original slice (1) ? Or does it returns the same thing as `MarshalBinary` (2) ? 
IF (1), then we need to keep track of an original slice alongside with the scalar (nist.Int).
IF (2), then there's absolutely no difference between `MarshalBinary` and `Bytes` 

I'm not so sure about both options, so what do you have in mind for the `Bytes()` function...

random/rand.go
nikkolasg: Yes there's surely a way since that is supposed to be their default behavior, return always random looking string, but for the moment there's a bug ( #70  ) so ...

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
+	Secret abstract.Secret
+	// Public is the corresponding public key
+	Public abstract.Point
+}
+
+// NewEdDSAKey will return a freshly generated key pair to use for generating
+// EdDSA signatures.
+// If stream == nil, it will take the random.Stream.
+func NewEdDSA(stream cipher.Stream) *EdDSA {
+	if stream == nil {
```
ineiti: Correcting myself: here you could use the `NewEd25519Scalar`

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
+	secret := suite.Secret().SetBytes(scalar[:32])
+	public := suite.Point().Mul(nil, secret)
+
+	return &EdDSA{buffer, scalar[32:], secret, public}
+}
+
+// Prefix returns the Prefix as being the right part of
+// the hashed seed
+func (e *EdDSA) Prefix() []byte {
+	c := make([]byte, len(e.seed))
```
ineiti: That would be `e.prefix`, no?

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
+	buffer := random.NonZeroBytes(32, stream)
+	scalar := sha512.Sum512(buffer)
+	scalar[0] &= 0xf8
+	scalar[31] &= 0x3f
+	scalar[31] |= 0x40
+
+	secret := suite.Secret().SetBytes(scalar[:32])
+	public := suite.Point().Mul(nil, secret)
+
+	return &EdDSA{buffer, scalar[32:], secret, public}
```
ineiti: as the variable-names are different from the field-names of `EdDSA`, I'd do the `Seed: scalar[32:]`-type initialisation of the structure.

eddsa/eddsa.go
ineiti: Oups - just saw that `NewEd25519Scalar` doesn't return the seed.

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
+	copy(eddsa, e.seed)
+	copy(eddsa[32:], pBuff)
+	return eddsa, nil
+}
+
+// EdDSASign will return a EdDSA signature using Ed25519. The secret must be
+// Picked() so it can use the prefix, the right part of the hashing done in
+// Pick.
+// Takes
+// - suite is the suite to use
```
ineiti: Update comment.

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
+	hash.Write(msg)
+
+	// deterministic random secret and its commit
+	r := suite.Secret().SetBytes(hash.Sum(nil))
+	R := suite.Point().Mul(nil, r)
+
+	// challenge
+	// H( R || Public || Msg)
+	hash.Reset()
+	Rbuff, err := R.MarshalBinary()
```
ineiti: Shouldn't we replace this with `R.Bytes()`? I thought `MarshalBinary` might compress the integer if the size of it is lower, no?

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, err
+	}
+	Abuff, err := e.Public.MarshalBinary()
```
ineiti: Question: does `MarshalBinary` follow the draft?

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// return R || s
+	var sig [64]byte
+	copy(sig[:], Rbuff)
+	copy(sig[32:], sBuff)
+
+	return sig[:], nil
+}
+
+// Verify verify a signature issued by EdDSA.Sign
```
ineiti: Typo

eddsa/eddsa.go
nikkolasg: R is a point so `MarshalBinary` is the correct representation of a point.

eddsa/eddsa.go
nikkolasg: Yes

eddsa/eddsa_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Comparing our implementation with the test vectors of the RFC
+func TestEdDSASigning(t *testing.T) {
+	for _, vec := range EdDSATestVectors {
+		seed, err := hex.DecodeString(vec.private)
+		assert.Nil(t, err)
+
+		stream := util.NewConstantStream(seed)
+
+		ed := NewEdDSA(stream)
```
ineiti: I think about what we discussed with Bryan and whether it's a good idea to use a stream for initialization when in reality we want it to be set to a given byte-slice. In crypto, `stream` is mostly used for random-streams.

eddsa/eddsa.go
nikkolasg: You just write comment faster than I can code, I was removing it when you commented out

abstract/group.go
ineiti: Isn't `MarshalBinary` compressing the integer if it's value allows it? I thought that:
- (Un)MarshalBinary are used to store things in a file - variable length, might have indexes and other stuff
- (Set)Bytes work directly on the value of the secret and always take/return the same length

eddsa/eddsa.go
ineiti: Sorry...

eddsa/eddsa_test.go
nikkolasg: > a stream for initialization when in reality we want it to be set to a given byte-slice

In **reality** (*not* in testing), we want that to be a *stream* so we get different bytes each time we call it. In **testing**, we want that to be a fixed slice of *bytes* so we know the expected result. For EdDSA, in **reality** we don't want to rely on a slice of bytes !

eddsa/eddsa.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	hash.Write(Rbuff)
+	hash.Write(Abuff)
+	hash.Write(msg)
+
+	h := suite.Secret().SetBytes(hash.Sum(nil))
+
+	// response
+	// s = r  h * secret
```
ineiti: Typo: `s = r + h * s`

eddsa/eddsa.go
```diff
@@ -0,0 +1,174 @@
...
+	s := suite.Secret().Mul(e.Secret, h)
+	s.Add(r, s)
+
+	sBuff, err := s.MarshalBinary()
```
ineiti: *The return of MarshalBinary/Bytes* - if it turns out to be correct that `MarshalBinary` on a secret might return fewer bytes, we need to use `Bytes()` here.

abstract/group.go
ineiti: Just if somebody else thinks like me: you're wrong ;) `(Un)MarshalBinary` is fixed-size for `secret`s and `point`s.

bford: Wait, what?  From what I'm reading, it sounds like you're saying the exact opposite of what I thought the semantics should be. :)

- MarshalBinary() should ALWAYS produce exactly MarshalSize() bytes, independent of the value of the integer being marshaled.  That's why it has the code above to pad the return from i.V.Bytes() if it is too short.

- Bytes() should probably behave the same way as big.Int.Bytes(), namely returning a byte-slice just large enough to hold the actual integer being encoded.  

In other words, SetBytes() takes a variable-length integer of any size as input, and Bytes() produces a variable-length return.  UnmarshalBinary() always consumes *exactly* MarshalSize() bytes, and MarshalBinary() always produces *exactly* MarshalSize() bytes.

So the implementation of Bytes() in this module should just call i.v.Bytes(), then reverse the slice if the caller wants little-endian.

Make sense?

ed25519/suite.go
bford: Agreed (or just make it private to the module).

Also, no need for function names in this module to contain "Ed25519" when they're in a module called "ed25519"!

nikkolasg: It's already removed since > 10mn now , reload your page! :D

random/rand.go
ineiti: I'd leave it here, as that method is only used till the bug will (eventually) be fixed.

eddsa/eddsa.go
```diff
@@ -0,0 +1,163 @@
...
```
bford: No, exactly the opposite, as discussed above.  MarshalBinary() always returns a fixed-size slice, and I'm pretty sure is what you want here.

abstract/group.go
nikkolasg: > In other words, SetBytes() takes a variable-length integer of any size as input, and Bytes() produces a variable-length return

So you want to keep a reference to the original slice then in nist.Int ... ? For me it does not make sense to keep twice the information where it's not needed. If a user called `SetBytes` it can still keep the original reference to the slice, we should not need to duplicate the information (the original slice and the *squeezed* version)

nist/int.go
```diff
@@ -37,8 +45,9 @@ var two = big.NewInt(2)
 // whose target is assumed never to change.
 //
 type Int struct {
-	V big.Int  // Integer value from 0 through M-1
-	M *big.Int // Modulus for finite field arithmetic
+	V  big.Int   // Integer value from 0 through M-1
+	M  *big.Int  // Modulus for finite field arithmetic
+	Bo ByteOrder // Endianness considered for this int
```
bford: Shouldn't it be BO rather than Bo, since it's an abbreviation and not a single word "bo"?

abstract/group.go
ineiti: No need to keep the original around. You can just squeeze whatever int you have to remove leading zeroes.

nist/int.go
```diff
@@ -299,10 +298,16 @@ func (i *Int) MarshalSize() int {
 func (i *Int) MarshalBinary() ([]byte, error) {
 	l := i.MarshalSize()
 	b := i.V.Bytes() // may be shorter than l
-	if ofs := l - len(b); ofs != 0 {
+	offset := l - len(b)
+
+	if i.Bo == LittleEndian {
+		return i.LittleEndian(l, l), nil
```
bford: Ooh, what is the reason for the LittleEndian() method to exist on this object.  Looks unnecessary and kinda hairy and gross to me.  Why not just delete it and instead:

1. Encode into a big-endian byte-slice, handling the offset correctly for big-endian as the code below does.
2. *after* that (before line 307/312 below), test for LittleEndian and conditionally to a util.Reverse on the encoded slice.

Wouldn't that be way simpler?  Am I missing anything?

nikkolasg: I also think it's not necessary now, but I understood you wanted to keep them, I again may have misunderstood. For me `LittleEndian` &  `BigEndian`  should be removed, does that seem OK ?

nist/int.go
```diff
@@ -346,6 +343,18 @@ func (i *Int) BigEndian(min, max int) []byte {
 	return buf
 }
 
+// Set value to a number represented by a byte string.
+// Endianness depends on the endianess set in i.
+func (i *Int) SetBytes(a []byte) abstract.Secret {
+	var buff = a
+	if i.Bo == LittleEndian {
+		buff = make([]byte, len(a))
+		util.Reverse(buff, a)
```
bford: I just realized that util.Reverse() should really have append()-like or Hash()-like semantics, allocating or extending the destination slice automatically if it is too short.  That simplify this and probably other calling code - e.g., here, this would become simply 'a = util.Reverse(nil, a)', and you could delete the 'buff' variable entirely.

bford: Note that the trivial way to fix util.Reverse() to have this behavior is for it simply to call util.Grow() on its first argument.

nikkolasg: Yep, could simplify things. Doing it here..

nikkolasg: > Note that the trivial way to fix util.Reverse() to have this behavior is for it simply to call util.Grow() on its first argument.

Well not quite because `util.Grow` still expect a non-nil slice ...

random/rand.go
bford: Sounds good.  In the meantime you could at least make it non-public (assuming it's only used internally within this module).

util/stream.go
```diff
@@ -0,0 +1,19 @@
+package util
+
+import "crypto/cipher"
+
+// ConstantStream is a cipher.Stream which always returns
+// the same value.
+type ConstantStream struct {
```
bford: Why should this type be public?

random/rand.go
nikkolasg: No it's not, it's used by `ed25519` and `edwards`

util/stream.go
```diff
@@ -0,0 +1,19 @@
...
+	seed []byte
+}
+
+// NewConstantStream returns a ConstantStream seeded with buff
+func NewConstantStream(buff []byte) cipher.Stream {
```
bford: Just rename this function to ConstantStream() and make it return a value of a hidden type, since there's no reason for the caller to be aware of anything but the cipher.Stream interface to it.

util/stream.go
nikkolasg: Because since users are going to have a reference on it, they may have to understand what's it about and for the doc to show the comments it must be public ... ?

abstract/group.go
nikkolasg: Well I don't get it then how you can return the original variable length slice byte without storing a reference to it ...

util/stream.go
bford: Just put the relevant comments on the function that produces the object of the hidden type!  No need to have an unnecessary public type just in order to hang comments off of!

nikkolasg: It's already done, but github still shows it .. .><

abstract/group.go
bford: No, I'm pretty sure big.Int.Bytes() just returns a *new* slice into which it copies a representation of the integer - and that's what nist.Int.Bytes() should do as well.  Note that big.Int has a Words() function that actually returns the internal-representation slice, but that's a slice of (32-bit?) words instead of bytes for efficiency, and it's only intended for use in performance-critical extensions to big.Int that really need to manipulate the internal state efficiently.  That's definitely not what Bytes() is for.  Just create a new slice, encode into it, and return it.

nikkolasg: Yes ok,  I get it this way. For me it still does not make much sense to for `SetBytes` and `Bytes` to not use the same slice. Example with `Un/Marshal` we can do :
```
a := sec.MarshalBinary()
sec.UnmarshalBinary(a)
a2 := sec.MarshalBinary()
a == a2
```
with `Bytes`:
```
secret.SetBytes(a)
a2 := secret.Bytes()
a != a2
```
But ok, if that makes sense for most of you then I abide :)

bford: Oooh - I'm pretty sure you really don't want to depend on being able to do a == a2 even in the MarshalBinary case!  Is there code that is doing that?  Sounds like a really bad idea to me - whether it works would definitely depend on which particular implementation of Secret or Point we're looking at.  For Point.MarshalBinary(), for example, I think special encoding is always going to be required, which will always allocate a new byte-slice.  If that's not happening for ed25519.Secret, then that might be a bug that needs to be fixed.

nikkolasg: Well, for the `(Un)Marshal` case, the snippet I posted above seems very normal and required to me, otherwise we get different representations of a secret ? 
And I've made the changes for the `Bytes()` now.

bford: Oh wait - by "a == a2" you meant a comparison of the *contents* of the slices, not a comparison of the slice-pointers?  (Because syntactically, "a == a2" in Go does *not* do a slice contents-comparison, if that's what you meant.)

If you did mean slice content-comparison, then yes, Marshal/Unmarshal can and should be expected to differ from SetBytes/Bytes in certain semantic subtleties.  

Marshal then Unmarshal then Marshal again should always produce the same thing the second time, since its purpose is to encode the internal state into its appropriate fixed-length "native" representation.  Similarly, if you do:

a := sec.Bytes()
sec.SetBytes(a)
a2 := sec.Bytes()

...then a byte wise comparison of a and a2 should always return true, even if the length of a (and a2) may vary depending on the specific integer contained in 'sec'.

However, UnmarshalBinary definitely should differ from SetBytes in that there are some inputs to SetBytes (e.g., 512-bit integers) for which a subsequent Bytes() won't yield the same thing.  Whereas if you do sec.UnmarshalBinary(a), then sec.MarshalBinary(), you should always get back a *if* a was actually a valid encoding of something.  (Note that sec.UnmarshalBinary() might return an error if not.)

nikkolasg: Yes sorry I meant in my *pseudo code* a byte content checking. And what you wrote is clear in my head but I'm just not so a fan of having this behavior of  `(Set)Bytes()`. Using `big.Int` it stays consistent: https://play.golang.org/p/pm3eAfoKYg
But anyway, as i'm clearly alone in my opinion, I already did the changes and that's fine, I'll still be able to sleep at night ;)

bford: But big.Int has quite different semantics from nist.Int.  big.Int is supposed to store an *arbitrary* big-integer.  nist.Int is supposed to store a *modular* integer (i.e., an element of a modular integer group or a scalar secret/exponent).  Because of this difference, big.Int by definition never needs to "decide" what to do if an integer supplied to SetBytes is "too big" because no integer is "too big".  nist.Int by contrast does need to decide what to do with a too-big integer, and the most convenient/useful semantics seems to apply the appropriate modulus.

Actually, perhaps some of your discomfort comes from the fact that "nist.Int" is probably misnamed.  For one thing, it should probably not be part of the 'nist' package at all (especially since that package probably should not even be part of the v1.0-stable version of crypto we're preparing), but probably elsewhere - I just haven't quite figured out where yet.  For another thing, perhaps it should not just be named 'Int' but instead perhaps 'ModInt' or something like that, to make it clear that it's not just a "different version of" big.Int but in fact does something else (which is true).

bford: Or when we rename abstract.Secret to abstract.Scalar, maybe nist.Int should become abstract.IntScalar, representing a "standard/default implementation" of the abstract.Scalar interface?

nikkolasg: I totally understand all the **semantics** and **practical** differences between nist.Int and big.Int but even though, from an **API perspective**, I should have said, it still *feels wrong*.  `(Set)Bytes` feels like it's a getter/setter interface while it's not because it introduces change (modulo) in the middle. Moreover, the `Bytes()` functionality can be questioned regarding its utility because it can be emulated using `MarshalBinary` and then cut off the padding.

Well, anyway, it's an endless loop and I think we should move on, so I accept this solution without any regret ;)

Regarding the naming, yes it would be a good idea to rename it `ModInt` or `IntMod` stg along those lines...

random/rand.go
Liamsi: If you still wish to hide it from outside, you could create a `internal` package on the top-level (which can safely be deleted later)

abstract/suite.go
```diff
@@ -39,6 +40,10 @@ type Suite interface {
 	// Generic constructor to instantiate any abstract interface type
 	// supported by this suite: at least Cipher, Hash, Point, Secret.
 	Constructor
+
+	// NewKey returns a freshly generated private key from the cipher stream.
+	// If cipher == nil, it uses random.Stream.
+	NewKey(cipher.Stream) Secret
 }
```
Liamsi: Question: The idea to have a `GenKeyPair` method in each suite is obsolete?

eddsa/eddsa.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	hash.Write(Rbuff)
+	hash.Write(Abuff)
+	hash.Write(msg)
+
+	h := suite.Secret().SetBytes(hash.Sum(nil))
+
+	// response
+	// s = r  h * s
```
Liamsi: Isn't sth. missing in the comment? (A `+`)?

eddsa/eddsa.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return err
+	}
+	hash := sha512.New()
+	hash.Write(sig[:32])
+	hash.Write(Pbuff)
+	hash.Write(msg)
+
+	h := suite.Secret().SetBytes(hash.Sum(nil))
+	// reconstruct S == k*A  R
```
Liamsi: Do you skip `+` on purpose? It's only an internal comment.

abstract/group.go
Liamsi: The `Bytes()` function is implemented but not in the interface... (like `SetBytes(...)` which is)

Liamsi:
Besides some mini-glitches: looks OK to me.

abstract/suite.go
nikkolasg: I went for the simplest. We could also have done `NewKeyPair(cipher.Stream) config.KeyPair` but config package already uses abstract, so that would mean to move the `config.KeyPair` into abstract which do not seem something that has to be done in this PR... :)

PR #75, opened 2016-06-14
Lib cosi integration

nikkolasg:
This PR is for integrating CoSi into the crypto library *into* development branch.
See old PR https://github.com/dedis/crypto/pull/67


ineiti:
Instead of making a PR against development, you can make a PR against #74, so we only see what you changed with regard to that other PR.

nikkolasg:
Yeah it was to avoid confusion as we did last time ... Best thing would to have #74 already merged into dev so I can PR against dev.

ineiti:
Now we learnt, no? It's difficult to look at the PR with all changes from ed being in here, too.

PR #80, opened 2016-06-14
anon test fixed -> duplicating fields

nikkolasg:
Using unexported fiels don't work with the current abstract.Read/Write methods. Either putting `anon/uSig` as public or duplicating the fields of `anon/uSig` into `anon/lSig`.
I chose to duplicate them as to keep the API clean, even if the code is less (~ still less than 10 lines of codes)
It's a hotfix  for an issue that should be scrutinized further ! See https://github.com/dedis/crypto/issues/81

### DIFF
--- a/anon/sig.go
+++ b/anon/sig.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/cipher"
 	"errors"
+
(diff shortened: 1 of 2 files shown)


nikkolasg:
Merged with #79

PR #83, opened 2016-06-17
Have GetResponse in cosi 

nikkolasg:
In BFTCosi we need to be able to retrieve the individual response of a node, here's why this little addition.


ineiti:
Was not really worth a PR - probably with regard to the discussion about not doing more than 1 thing in a PR?

Issue #84, opened 2016-06-30
Implement a Clone method for abstract.Point 


Liamsi:
Due to popular demand, we want to add a `Clone(p abstract.Point)` method to the `abstract.Point` interface. Currently, there are three ways to do create a copy of an existing point: using internal representation, using (un)marshal, with `Add()` or `Mul()` with `Zero()` or `One()` respectively. For this issue we want:

- [x] benchmarks (on ed25519?) of the 3 mentioned methods to be sure which method is the most efficient
- [x] add a `Clone` method to the interface and implement it according to the findings above
- [x] add a `Set` method to `abstract.Point` to be consistent
- [x] add a `Clone` method to `abstract.Scalar` to be consistent

ineiti:
Wouldn't it be better to have `Clone() abstract.Point` and `Clone() abstract.Scalar`?

Liamsi:
Right I didn't even mention which interface (I meant abstract.Point). 
`Scalar` has a `Set` method. Not sure if it's good to add another method with (almost) the same functionality.

Liamsi:
The only difference between `Clone() abstract.Point` and `Clone(abstract.Point)` would be that the first would always create a new point, while the latter would allow you to reuse an existing instance (but then `Set` would be a better name, right?). Should we use the `Set` approach for the sake of consistency? Or just add a `Clone` method to either both or only to `Point`?

ineiti:
So you would propose to have

```
type Scalar interface {
    Set(s Scalar) Scalar
    Clone() Scalar
}

type Point interface{
    Set(p Point) Point
    Clone() Point
}
```

?

Liamsi:
Yes, or only `Set` (as it already exists for `Scalar`). But the above is also OK, I guess. Let's (briefly) discuss with @nikkolasg and @Daeinar about it (next week), OK?

bford:
I have considered at various points adding a Clone() method to these interfaces, but never quite got around to deciding whether or not to do it.  

[...]

PR #86, opened 2016-07-05
Consistent Clone() and Set() for abstract.Point and abstract.Scalar

Liamsi:
*Work in progress* and related to #84


ed25519/suite_test.go
```diff
@@ -0,0 +1,63 @@
...
+	assert.True(t, P1.Equal(P2))
+}
+
+func TestCurvePointCloneRandom(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Randomized test skipped in short mode")
+	}
+	iters := 1000
+	null := edSuite.Point().Null()
+	for iters > 0 {
```
ineiti: `for iters := 0; iters < 1000; iters++` ?

Liamsi: Doesn't make a difference, but yes the old-school is probably more readable.

ineiti:
Is it possible to put all tests in `test/` so that they are done for all suites at the same time?

ed25519/suite_test.go
ineiti: yes, that was about readability.

Liamsi:
> Is it possible to put all tests in test/ so that they are done for all suites at the same time?

Also thought about that. I think, yes it's possible. This is pretty much work in progress and I worked myself through each suite individually. I'll clean the test before I set it to Ready4Merge again.

ed25519/suite_test.go
Liamsi: OK, will change it together with putting all tests for each suite in `test/` at the end.

Liamsi:
Can be reviewed if travis passes (see the discussion here #84 )

nist/curve.go
```diff
@@ -252,6 +252,18 @@ func (c *curve) Point() abstract.Point {
...
 
+func (p *curvePoint) Set(P abstract.Point) abstract.Point {
+	p.x = P.(*curvePoint).x
+	p.y = P.(*curvePoint).y
+	return p
+}
+
+func (p *curvePoint) Clone() abstract.Point {
+	// TODO test if we really created a new Point
+	// changing original point doesn't change cloned
```
nikkolasg: Apparently the tests are already verifying that invariant so the comment can be removed i guess.

Liamsi: Thanks, I forgot to remove this one.

test/test.go
```diff
@@ -28,6 +28,80 @@ func testEmbed(g abstract.Group, rand cipher.Stream, points *[]abstract.Point,
 	*points = append(*points, p)
 }
 
+func testPointSet(g abstract.Group, rand cipher.Stream) {
+	N := 1000
+	null := g.Point().Null()
+	for i := 0; i < N; i++ {
```
nikkolasg: I don't really get why we do this a thousand time ... If it's "just to be safe", I'm not sure I would keep it, because it only makes the test bigger and slower (it's different than the case with SDA with network communications and concurrency issues and such). If there's a specific reason, then I'd love to hear it.

test/test.go
```diff
@@ -28,6 +28,80 @@ func testEmbed(g abstract.Group, rand cipher.Stream, points *[]abstract.Point,
...
+func testScalarSet(g abstract.Group, rand cipher.Stream) {
+	N := 1000
+	one := g.Scalar().One()
+	for i := 0; i < N; i++ {
+		s1 := g.Scalar().Pick(rand)
+		s2 := g.Scalar().Set(s1)
+		if !s1.Equal(s2) {
+			panic("Clone didn't create a scalar s2 with same value as s1's.")
+		}
+		if !s1.Equal(one) {
```
nikkolasg: You tested against `Add` with the tests regarding the `Point` and here you do with `Mul`. If you stick to `Add` here, you can remove that `if` (because *one + one = two*)

test/test.go
Liamsi: I first added some extensive "randomized" tests which were only ran, if not in "short mode" (otherwise thy will be skipped): https://github.com/dedis/crypto/pull/86/files/9b4da9e2cc655606f348ec6b2c1394e90102d8ad?diff=unified#diff-287a11fd3cbb07e3e4f6b496c93fa64fR46
This isn't possible the way we do the tests now. @ineiti and me agreed to leave this tests if they don't really take much time.

test/test.go
```diff
@@ -28,6 +28,80 @@ func testEmbed(g abstract.Group, rand cipher.Stream, points *[]abstract.Point,
...
+	N := 1000
+	null := g.Point().Null()
+	for i := 0; i < N; i++ {
+		P1, _ := g.Point().Pick(nil, rand)
+		P2 := g.Point()
+		P2.Set(P1)
+		if !P1.Equal(P2) {
+			panic("Set() set to a different point.")
+		}
+		if !P1.Equal(null) {
```
nikkolasg: Stylish issue: prefer doing the *regular* code outside an `if`
```
if P1.Equal(null) {
    continue
}
... // regular test
```

test/test.go
nikkolasg: Ok you agreed but that does not say why you agreed to keep it ? What is the benefit of doing it a thousand time ?
There's no concurrency here, no external actor, no statistics involved and so on ...

test/test.go
ineiti: usually yes, here I wouldn't like the `continue`, as it could lead to errors if you overlook it.

test/test.go
Liamsi: Good catch, but than there would still be an `if` checking agains zero or not?

test/test.go
nikkolasg: ? Well, in every other situation I can still overlook the `if` also, that does not mean we don't prefer doing the regular code outside. I don't get why the `continue` keyword is an exception ... it's part of the language as well.
But ok, I'm only back a few hours from my vacation here, so I don't really mind :D

test/test.go
Liamsi: Just to be sure (that we didn't just pick a point for which the test passes). Adding randomized tests is in fact good practice. If you the implementations (sth the test shouldn't really care/should be agnostic about) it's impossible that for one point/scalar the test passes but not for other. 
But if you have strong reason to remove it (besides adding a rather negligible amount of extra time, especially compared to the overall test time of the crypto package), I'm happy to remove it.

test/test.go
nikkolasg: Right... There's some solutions and we can dive in this problem to have a perfect test I don't think that's a reasonable good thing to do now :D
So be the judge of that if you want to stick with `Mul` or `Add` .

test/test.go
nikkolasg: > Adding randomized tests is in fact good practice

Well, since I'm a rookie I've done a quick search on unit test and I did not find anything about randomized tests. On the contrary, I found reasons to NOT do that: [SO1](https://stackoverflow.com/questions/256065/testing-with-random-inputs-best-practices) and [SO2](https://stackoverflow.com/questions/3441686/what-are-the-downsides-using-random-values-in-unit-testing)
So I still have to disagree with you on that, because I still don't see any good reason to keep the tests like that...

nikkolasg: But again, majority wins so I abide :)

Liamsi: Hmm, I agree that reproducibility is indeed a *very* good argument against random tests. On the other hand we increase the coverage (which is a treadeoff between testing against only one totally deterministically created point and testing against all). 
@ineiti if you are still convinced that we should keep it like this I would still prefer to leave it as is.

test/test.go
Liamsi: I'm OK with both. But as this is only a test I would also just leave it as is.

ineiti:
OK for me.

nikkolasg:
Majority has decided: OK to merge.

PR #87, opened 2016-09-30
Fixing nist.Int.Clone()

nikkolasg:
Value was copied but value is a big.Int which contains the real value which is a slice (i.e. pointer).

### DIFF
--- a/nist/int.go
+++ b/nist/int.go
@@ -140,7 +140,10 @@ func (i *Int) Set(a abstract.Scalar) abstract.Scalar {
 }
 
 func (i *Int) Clone() abstract.Scalar {
-	return &Int{V: i.V, M: i.M, BO: i.BO}
+	v := new(big.Int).Set(&i.V)
+	m := new(big.Int).Set(i.M)
+	return &Int{V: *v, M: m, BO: i.BO}
+	//return &Int{V: I->V, M: i.M, BO: i.BO}
 }
 
 // Set to the value 0.  The modulus must already be initialized.
--- a/nist/int.go
+++ b/nist/int.go
@@ -143,7 +143,6 @@ func (i *Int) Clone() abstract.Scalar {
 	v := new(big.Int).Set(&i.V)
 	m := new(big.Int).Set(i.M)
 	return &Int{V: *v, M: m, BO: i.BO}
-	//return &Int{V: I->V, M: i.M, BO: i.BO}
 }
 
 // Set to the value 0.  The modulus must already be initialized.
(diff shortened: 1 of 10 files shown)


nist/int.go
```diff
@@ -140,7 +140,9 @@ func (i *Int) Set(a abstract.Scalar) abstract.Scalar {
 }
 
 func (i *Int) Clone() abstract.Scalar {
-	return &Int{V: i.V, M: i.M, BO: i.BO}
+	v := new(big.Int).Set(&i.V)
+	m := new(big.Int).Set(i.M)
+	return &Int{V: *v, M: m, BO: i.BO}
```
ineiti: `return new(Int).Init(&i.V, i.M)`

nikkolasg: But then the ByteOrder may not be correct.

nist/int_test.go
```diff
@@ -63,3 +63,22 @@ func TestIntEndianBytes(t *testing.T) {
 	assert.Equal(t, 2, i.MarshalSize())
 	assert.NotPanics(t, func() { i.LittleEndian(2, 2) })
 }
+
+func TestIntClone(t *testing.T) {
+	modulo, err := hex.DecodeString("1000")
+	moduloI := new(big.Int).SetBytes(modulo)
```
ineiti: Does
`moduloI := new(big.Int).SetBytes(0x1000)`
not work?

nist/int_test.go
```diff
@@ -63,3 +63,22 @@ func TestIntEndianBytes(t *testing.T) {
...
+	assert.Nil(t, err)
+	v, err := hex.DecodeString("10")
+	assert.Nil(t, err)
+
+	base := new(Int).InitBytes(v, moduloI)
```
ineiti: Same thing - `0x10` instead of `hex.DecodeString`?

nist/int.go
ineiti: yes, so we have to add that, too... I saw it does not correctly marshal/unmarshal 'my' way.

ineiti: So the complete is:
```
	ni := new(Int).Init(&i.V, i.M)
	ni.BO = i.BO
	return ni
```

So either we take your solution, or we re-write `Int.Init` to take `BO` as a parameter, which might makes some sense, given that the `V` -argument to `Init` could be in the wrong endianness. But I don't know if `BO` is really relevant in `i.V.Set(V).Mod(&i.V, M)` - probably not.

Your call, as Ismail is not here to vote ;)

nikkolasg:
Can we go back to the simpler version without changing nist.Int  and let #90 decide on that matter ?

ineiti:
Replaced by #87

PR #92, opened 2016-10-09
Public Verifiable Secret Sharing (PVSS)

Daeinar:
Extract PVSS from the cothority project and put it in here. Also see https://github.com/dedis/cothority/issues/624.

This PR still has some dependencies to `dedis/cothority` which I wanted to get rid of ideally. The `log` dependencies in `pvss_test.go` are trivial to remove. `pvss.go` has some dependencies to `dedis/cothority/crypto` (in particular `hash.go`) and while looking at some of the files there I realised that they might better fit into the crypto library. Thus we might want to consider refactoring `dedis/cothority/crypto` and move some / all of its functionality into the crypto library.

Thoughts?

### DIFF
--- a/poly/pvss.go
+++ b/poly/pvss.go
@@ -1,0 +1,298 @@
+package poly
+
+import (
+	"errors"
+
+	"github.com/dedis/cothority/crypto"
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/random"
+)
+
+// This package implements public verifiable secret sharing (PVSS) using
+// non-interactive zero-knowledge (NIZK) proofs to show equality of discrete
+// logarithms (DLE).
+
+// DLEProof resembles a NIZK dlog-equality proof. Allows to handle multiple proofs.
+type DLEProof struct {
+	suite abstract.Suite
+	base  []DLEProofBase
+	core  []DLEProofCore
+}
+
+// DLEProofBase contains the base points against which the core proof is created.
+type DLEProofBase struct {
+	g abstract.Point
+	h abstract.Point
+}
+
+// DLEProofCore contains the core elements of the NIZK dlog-equality proof.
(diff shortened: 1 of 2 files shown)


nikkolasg:
> Thus we might want to consider refactoring dedis/cothority/crypto and move some / all of its functionality into the crypto library.

We originally made this design decision because we were not sure if we had "full power" over the crypto repository or if we always had to wait for Bryan's OK for each PR. We prefered having our own code that can be merged quickly. Now that assumption can be removed because from a global agreeing we can merge stuff into master.  SO let's go for that. We'll need to decide what needs to be inside and maybe bring some modification but it's a  good idea.

poly/pvss.go
```diff
@@ -0,0 +1,298 @@
...
+	"errors"
+
+	"github.com/dedis/cothority/crypto"
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/random"
+)
+
+// This package implements public verifiable secret sharing (PVSS) using
+// non-interactive zero-knowledge (NIZK) proofs to show equality of discrete
+// logarithms (DLE).
```
Liamsi: If you want the documentation to display nicely, you could either add a `doc.go` file with a concise explanation or (if you are short in time, which I guess never happens ;-) you could simply move this comment to the top. Then it would need to start with `// Package poly ...`. See: https://blog.golang.org/godoc-documenting-go-code.
Also, I think a link to (or at least name of) the paper where the exact algo is described would be useful.

Liamsi: OK, I see there already is such documentation in this package... We could create a `proof` or `pvss` sub-package for this, but I'm not sure this a good idea.

nikkolasg:
Closing because it's being taken care of in https://github.com/dedis/crypto/issues/96

PR #94, opened 2016-11-16
Purbtls

LegoShrimp:
Here is what I have now. I am not sure if it should just be left in my repo or merged.
Purb primitive is in /purb, but currently only purbtls uses it.
Purbgp is in its own folder. Currently it doesn't use the separate /purb.


nikkolasg:
> I am not sure if it should just be left in my repo or merged.

Well, I think once it reaches a good state, that'd be a perfect target for the crypto repo !
However, to reach this "good state" there's several things that need to be done in priority:
* separate the pull requests into *at least* 3: one for PURB, one for purbTLS, and one for purbPGP. If  you think that you could reduce more the size of the PR then go with it. Currently, 26 files with > 98,000 "additions" is not realistically review-able.
* Making  purbpgp using the PURB package *if possible*. 
* merge it with master
* make golint pass

Then after we'll be able to comment on this PR and within some iterations of the comment/correction process, it should be integrated :)

LegoShrimp:
Alright makes sense. It isn't as bad as it looks 94,000 of the lines are from a dummy file I made to test the padding scheme. There are probably some other files like that as well. Which should probably also be added as a separate pull request. Any simple way to split up the pull requests? Just create a new branch and remove the folders not related to it? Although I will clean up the files and see if separate pull requests still seem required.

nikkolasg:
> It isn't as bad as it looks 94,000 of the lines are from a dummy file I made to test the padding scheme. 

ok good to know. In the future, try to avoid putting "dump" files like this on Git because every user that uses dedis/crypto will have to download these 90,000 "additions" if it ever gets merged to master ... :s Don't worry, I'm the first to have made this mistake ;)
Anyway, I think it's possible to rebase the history pretty easily since it's only in your branch and not in master.

> Although I will clean up the files and see if separate pull requests still seem required.

Sure, sounds fine. I'm not familiar with the code so it may work as one PR but keep in mind that if you start implementing different protocols/strategy in one PR, you can probably split it up regarding those different protocols/strategy.

bford:
Indeed, please don't commit large data files to git source repositories in general - especially automatically-generated test files and such.  Instead just commit the programs/scripts necessary to regenerate them.  If we eventually have large test suites we need to store, we can keep them in separate dedicated repositories if necessary.

ineiti:
@LegoShrimp - There is a `.gitignore`-file that holds files that don't need to be tracked, I use it quite a lot to avoid adding files when committing with git.

ineiti:
I also see you're using nego (https://godoc.org/github.com/dyv/crypto/nego) - did you change their package, or would it be possible to use their package without changing it? Then you wouldn't have to include their file, which would make the PR smaller.

nikkolasg:
@ineiti good call. Moreover, why do you use the one from dyvan instead of the one present in the crypto library ? There's a `nego` package in the master branch at least...

LegoShrimp:
If I am, it is not intentional, it looks like dyv/crypto/nego is unmodified from when the library was forked. I originally started by modifying the files in nego to make purb, but decided to create a new folder and leave that unchanged, but apparently I forgot to revert some of the files, or accidentally changed them later. 
I will try to create a few new pull requests later one that just contains: 
1. the padding scheme 
2. PURB and PURBgp, unless splitting them is simpler then I think it will be. 
3. purbTLS. 
While I do this i will make sure not to add any unnecessary files and revert files that aren't actually used.

jeffallen:
I propose to close this because network stuff does not belong in kyber, and because we will be adding real TLS to Onet next month.

PR #95, opened 2016-11-18
Purb

LegoShrimp:
This just contains the padding code, and PURB code.


purb/skip.go
```diff
@@ -0,0 +1,208 @@
...
+// Pick a random height for a new skip-list node from a suitable distribution.
+func skipHeight() int {
+	height := 1
+	for v := randUint32() | (1 << 31); v&1 == 0; v >>= 1 {
+		height++
+	}
+	return height
+}
+
+type skipNode struct {
```
nikkolasg: If I understood correctly, this is taken from @bford 's code somewhere's else in the library ? If so, maybe we could come up with something that does not require duplicate the code ?

padding/padding.go
```diff
@@ -0,0 +1,157 @@
...
+		padding[i] = PADBYTE //random choice
+	}
+	return padding
+}
+
+//Inputs:
+//overhead int, is the number of bytes the encryption will add to the plaintext
+//msg []byte, is the plaintext message that needs to be padded.
+//Returns a properly padded message.
+func PadGeneric(msg []byte, overhead uint64) []byte {
```
nikkolasg: Moreover, from what I understood, this "overhead" only works when `len(pt) == len(ct) + overhead`. this conditions is known to be true for most notable stream ciphers, so maybe add a comment specifiying this condition ?

purb/skip.go
LegoShrimp: It is possible, but most (all?) of the methods are private. So to use it we would have to modify the source code, which might be a better solution then copying it. Although possibly we should consider removing nego, since PURB and PURB TLS is the evolution of what the nego code did/was meant to do.

padding/padding.go
```diff
@@ -0,0 +1,157 @@
...
+	padAmount := GetPaddingLen(uint64(len(msg)), overhead+PADLEN)
+	padding := GeneratePadding(padAmount)
+	buf := new(bytes.Buffer)
+	binary.Write(buf, binary.BigEndian, int64(padAmount))
```
LegoShrimp: Almost certainly not, but I am not sure what value would make the most sense. possibly just shrink it to int32? Although if this is used for a file bigger then 8 GB it is possible it would need more padding. Although it could be an application dependent thing probably.

padding/padding.go
```diff
@@ -0,0 +1,157 @@
...
+	"encoding/binary"
+	//	"fmt"
+	"math"
+)
+
+//overhead to encode the amount of padding,
+const PADLEN = 8
+
+//the byte used to pad messages.
+//0 because it makes things easier elsewhere, might need to be changed
```
LegoShrimp: Since we are encrypting it anyways I don't think it needs to be random, although it is possible there are some benefits to doing random padding. Like protecting against use of bad symmetric/block ciphers that leak when encrypting the same byte over and over again. But I am not sure that really should be something we worry about.

jeffallen:
Are we planning on integrating this? I propose to close it because I think there is other more up to date work happen on PURBs.

Issue #96, opened 2017-01-20
[Refactoring] Secret Sharing


Daeinar:
This will address (at least) the following issues:

- Deletion of the entire secret sharing code base found in `dedis/crypto/poly`
- New framework for secret sharing now in `dedis/crypto/share`
  + Rewrite the core of the library for polynomials
  + Integration of *public verifiable secret sharing* found in [randhound](https://github.com/dedis/cothority/blob/master/randhound/proof.go)
  + Rewrite of the old `poly/deal.go` into `share/vss.go`
  + Rewrite of the `poly/joint.go` and `poly/schnorr.go`
  + Adapt/rename the (JVSS)[https://github.com/dedis/cothority/tree/master/jvss]
  
By side effect, this will also include
  + `sign` package to pave the way for later bigger restructuring. It will include a basic `Schnorr` signature.
  + `DLEQ` (discrete log equality) functionality into the `proof` package

nikkolasg:
After some trials and errors, I intend to propose the different PRs such as:
 1. add `sign/schnorr{_test}.go` in `refactoring_sign` merging against master
 2. add `proof/dleq.go` in `refactoring_dleq` mering against master
 3. add `share/vss{_test}.go` in `refactoring_share_vss` merging against master
 4. add `share/pvss{_test}.go` in `refactoring_share_pvss` merging against master
 5. add `share/vss{_test}.go` in `refactoring_share_vss` merging against master
 6. add `share/joint{_test}.go` in `refactoring_share_joint` merging against master
 7. add `share/schnorr{_test}.go` in `refactoring_share_schnorr` merging against master

nikkolasg:
Done.

PR #98, opened 2017-01-20
PR: Refactor poly/sharing.go

Daeinar:
See #96.


Daeinar:
@nikkolasg: Added the changes as discussed. Btw. currently you can give a `cipher.stream` to the constructor of a secret sharing polynomial for choosing the coefficients, see https://github.com/dedis/crypto/blob/master/poly/sharing.go#L36-L37. I guess the intention behind this is to have a way to generate the coefficients deterministically from a seed. Do we need / want that? If yes wouldn't it be simpler to just give a `[]byte` seed? Creating the coefficients deterministically could be nice from transparency-perspective in some form of proof, e.g., if a dealer has to defend himself against accusations of the verifiers in case something went wrong. Currently, I can't think of any other purpose ...

nikkolasg:
> Do we need / want that? 
That's a good question, let's do pros/cons:
Pros:
 + Deterministic coefficients: useful for *testing* as well as the "transparency-perspective" you mentioned above
Cons:
 + Opens up a possible security breach if creating coefficient with a wrong PRNG

Please add more to that list if you see others.
In my opinion, I would keep it because it's a standard way to do things in Golang at least it seems (look at `tls.Config.Rand`) and I'm pretty sure one day someone will ask how do we determnistically generate coefficients ... :D 

> If yes wouldn't it be simpler to just give a []byte seed?

Mhhh I'm not so sure about that. Giving a slice of byte as a seed creates some questions that we don't have to ask now: 
 + how large should be the seed ?
 + what to do with the seed ? what prng should we use ? which one is secure at the moment ? We usually leave these questions to the underlying OS by taking /dev/urandom.

It seems to me we are introducing another PRNG layer <=> complexity which does not look necessary regarding the library itself... 
By hiding that in a `Stream`, if the caller does **not** pas nil, it can pass its own PNRG already doing any kind of fancy constructions or taking numbers from any source he wants.

poly/sharing_new.go
```diff
@@ -0,0 +1,161 @@
...
+	return nil
+}
+
+// Threshold returns the secret sharing threshold.
+func (p *PriPoly) Threshold() int {
+	return len(p.coeffs)
+}
+
+// SharedSecret returns the shared secret p(0), i.e., the constant term of the polynomial.
+func (p *PriPoly) SharedSecret() abstract.Scalar {
```
nikkolasg: Is this function required ? Usually if you create the `PriPoly`, you're supposed to know the secret that you put in right ?
But if you think it is, then what do you think about adding field `secret` to the `PriPoly` struct in order to avoid doing Lagrange interpolations each time you want to recover the secret ?
It keeps the secret in memory, but avoid useless computations.

poly/sharing_new.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// XXX: Do we need that?
+func (p *PriPoly) String() String {
+	return nil
+}
+
+// xCoords creates an array of x-coordinates for Lagrange interpolation. In the
+// returned array, exactly t x-coordinates are non-nil.
+func priXCoords() []abstract.Scalar {
```
nikkolasg: According to the actual `sharing.go`, the functions's output are exactly the same (`[]Scalar`) but the inner `range` loops over different kinds (`Scalar` for `PriPoly` and `Point` for `PubPoly`).
We could merge both together and having this function taking a function parameters looking like:
```go
func xCoords(t int, func isNil(int) bool) []abstract.Scalar {
  ...
}

func (p *priPoly) coefNil(i int) bool {
  return len(p.coeffs) >= i || p.coeffs[i] == nil
}

func (p *pubPoly) coefNil(i int) bool {
  return len(p.commits) >= i || p.commits[i] == nil
}
```

In any case, I don't really mind if you think that's too much, feel free to keep it un-merged, I'm also OK with that since it's quite small function.

nikkolasg: BTW, why does this `xCoords` panics instead of returning an error ? I don't see any reason why, so I would make it return an error instead.

poly/sharing_new.go
```diff
@@ -0,0 +1,161 @@
...
+	// XXX: check if it's possible to merge with pubXCoords
+	return nil
+}
+
+// RecoverSecret reconstructs the shared secret p(0) using Lagrange interpolation.
+func RecoverSecret(shares []PriShare, t int) abstract.Scalar {
```
nikkolasg: As above, I'd like this one to return an error instead of panicking.

nikkolasg:
In general, I'm OK with the API changes now. One last thing so I can work with you alongside, is to directly make this file `sharing.go` and delete the `sharing_new.go` so we can start having things that build even if not implemented.

poly/sharing_new.go
Daeinar: The shared secret is already contained in the `coeffs` slice at index 0, so this function just returns that value. And the current constructor for a `PriPoly` allows you to also give no specific secret at which point it is just generated randomly, see https://github.com/dedis/crypto/blob/master/poly/sharing.go#L40-L41. I think we want to keep that functionality.

poly/sharing_new.go
Daeinar: Both your proposals on merging and returning an error sound good to me, lets do that.

Daeinar:
I'll move the current `sharing.go` to `sharing_old.go` so that we still have it in the branch while we are writing the new version, okay?

Daeinar:
Moreover, I am also okay with keeping the `cipher.stream` for generating the random coefficients, your points above sound reasonable.

share/vss.go
```diff
@@ -0,0 +1,493 @@
...
+//  - the second base H out of a list of public keys
+//  - the hash of the message to sign in a Complaint
+var HashFunc = sha256.New
+
+func deriveH(suite abstract.Suite, verifiers []abstract.Point) abstract.Point {
+	var b bytes.Buffer
+	for i := range verifiers {
+		verifiers[i].MarshalTo(&b)
+	}
+	h := HashFunc()
```
Daeinar: Why not just using `suite.Hash()`?

share/vss.go
```diff
@@ -0,0 +1,493 @@
...
+	Deal      *Deal
+	// Signature over the msg
+	// H(Index || deal.MarshalBinary() || verifiers)
+	Signature []byte
+}
+
+// DealerResponse is a message that is broadcasted by the Dealer in response to
+// a Complaint. It contains the original Complaint as well as the shares
+// distributed to the complainer.
+type DealerResponse struct {
```
Daeinar: Hmmm I wonder if this should be renamed to `ResponseComplaint`, `ComplaintResponse` or even just `Response` since the `Complaint` struct also does not contain a `Verifier` label?!

share/vss.go
```diff
@@ -0,0 +1,493 @@
...
+// not the same, or the session identifier is wrong, it returns a Complaint
+// which must be broadcasted to every participants including the Dealer.
+// If the complaint is deemed invalid because of a wrong index (not the same as
+// the verifier) or because the verifier already received a Deal, it does not
+// return a complaint.
+// ReceiveDeal returns an error in any case to let the user know what was the
+// error.
+// XXX API question: return value.For the moment, the default is to let the user
+// know everything that is wrong through the error.
+func (v *Verifier) ReceiveDeal(d *Deal) (*Approval, *Complaint, error) {
```
Daeinar: Would it make sense to replace the "Receive" prefixes everywhere with "Process" or so? To me "Receive" always implicates some involvement of network communication but at this point the receipt of the message/deal has already happened at another place.

nikkolasg: True. I like `Process`.

share/vss.go
nikkolasg: * `Deal` 
* `Approval`
* `Complaint`
 * `Response` ? `Justification` ? `Defense` ? `Argument` ? `Explanation` ?  (or even `Salvation` hehe)

share/vss.go
nikkolasg: That's a good question. I think there's no real inherent good reasons code-wise to do it like this. 
My main motive was simply because I'm not a fan of using `suite.Hash()` since we never know what hash function it is, so I prefer to be explicit about that at least. 

At the same time, I agree it's also going against the "code spirit" where we use these abstractions to avoid declaring things in this way...
I have absolutely *no opposition* on using the suite's hash anyway  so if you think it's overall a better solution, I'll change that.

share/vss.go
nikkolasg: I like `Argument` or `Explanation`. `Response` is too generic and `Approval` and `Complaint` are already responses from the `Dealer`'s perspective...

share/vss.go
Daeinar: I agree that with `suite.Hash()` we do not know which hash function is used. However, in most cases it will be `sha256` and somebody reading the code might wonder at this point "Huh, why is it important here to explicitly use SHA256?!". I think for consistency it would be better to just use `suite.Hash()`, yes.

share/vss.go
Daeinar: Hmm I think out of the ones listed above I like `Justification` the most, because the `Dealer` tries to *justify* his actions and proof that he didn't do anything wrong. `Argument` is also a bit too generic. `Explanation` might also work ...

share/pvss_test.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+func TestPVSS(t *testing.T) {
+
+	suite := edwards.NewAES128SHA256Ed25519(false)
+
+	G := suite.Point().Base()
+	H, _ := suite.Point().Pick(nil, suite.Cipher([]byte("H")))
+
+	n := 20
```
nikkolasg: `10` or even lower should be good enough, and it'll make the test go way faster. It's currently 1.2, 1.3 sec for this test only on my machine. Same for `TestPVSSDelete`.

nikkolasg:
This PR serves only as a reference. Smaller and cleaner PR are on the way.

ineiti:
Obsolete now

PR #101, opened 2017-02-01
[Refactoring] Adding a sign/ package with Schnorr

nikkolasg:
This PR paves the way for some functionality needed in VSS later. It adds a `sign` package with a `schnorr` signature framework, the one from `dedis/onet`.

### DIFF
--- a/sign/schnorr.go
+++ b/sign/schnorr.go
@@ -1,0 +1,90 @@
+package sign
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/random"
+)
+
+// Schnorr creates a Schnorr signature from a msg and a private key. This
+// signature can be verified with VerifySchnorr. It's also a valid EdDSA
+// signature.
+func Schnorr(suite abstract.Suite, private abstract.Scalar, msg []byte) ([]byte, error) {
+	// using notation from https://en.wikipedia.org/wiki/Schnorr_signature
+	// create random secret k and public point commitment r
+	k := suite.Scalar().Pick(random.Stream)
+	r := suite.Point().Mul(nil, k)
+
+	// create challenge e based on message and r
+	public := suite.Point().Mul(nil, private)
+	e, err := hash(suite, public, r, msg)
+	if err != nil {
+		return nil, err
+	}
+
+	// compute response s = k - x*e
+	xe := suite.Scalar().Mul(private, e)
+	s := suite.Scalar().Sub(k, xe)
+
+	var b bytes.Buffer
+	if _, err := e.MarshalTo(&b); err != nil {
+		return nil, err
+	}
+	if _, err := s.MarshalTo(&b); err != nil {
+		return nil, err
+	}
+	return b.Bytes(), nil
+}
+
+// VerifySchnorr verifies a given Schnorr signature. It returns nil iff the
+// given signature is valid.  NOTE: this signature scheme is malleable because
+// the response's unmarshalling is done directly into a big.Int modulo (see
+// nist.Int).
+func VerifySchnorr(suite abstract.Suite, public abstract.Point, msg, sig []byte) error {
+	challenge := suite.Scalar()
+	response := suite.Scalar()
+	scalarSize := challenge.MarshalSize()
+	sigSize := scalarSize * 2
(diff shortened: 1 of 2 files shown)


sign/schnorr.go
```diff
@@ -0,0 +1,95 @@
...
+	if _, err := e.MarshalTo(&b); err != nil {
+		return nil, err
+	}
+	if _, err := s.MarshalTo(&b); err != nil {
+		return nil, err
+	}
+	return b.Bytes(), nil
+}
+
+// VerifySchnorr verifies a given Schnorr signature. It returns nil iff the
```
ineiti: s/iff/if/

sign/schnorr.go
```diff
@@ -0,0 +1,95 @@
...
+// given signature is valid.  NOTE: this signature scheme is malleable because
+// the response's unmarshalling is done directly into a big.Int modulo (see
+// nist.Int).
+func VerifySchnorr(suite abstract.Suite, public abstract.Point, msg, sig []byte) error {
+	var challenge = suite.Scalar()
+	var response = suite.Scalar()
+	scalarSize := challenge.MarshalSize()
+	sigSize := scalarSize * 2
```
ineiti: style: why use twice`var` and twice `:=`? I propose to use `:=` in functions and `var` in package-global-variables.

sign/schnorr.go
```diff
@@ -0,0 +1,95 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/random"
+)
+
+// HashFunc is the hash function used to hash the message for the signature
+// generation and verification
+var HashFunc = sha256.New
```
ineiti: style: global variable that is used only once - can go into the local `hash`-function, IMO.

sign/schnorr_test.go
```diff
@@ -0,0 +1,23 @@
+package sign
+
+import (
+	"testing"
+
+	"github.com/dedis/crypto/config"
+	"github.com/dedis/crypto/ed25519"
+)
+
+func TestSchnorrSignature(t *testing.T) {
```
ineiti: does not test failing parts of the schnorr-verification.

sign/schnorr.go
Daeinar: In math "iff" sometimes means "if and only if", see https://en.wikipedia.org/wiki/If_and_only_if

Not sure if it is the case here or if it's just a typo.

sign/schnorr.go
nikkolasg: In fact, I should use the `suite.Hash()` function, don't you think ?

sign/schnorr.go
nikkolasg: Exactly, that's what I meant with this `iff`: it returns nil if and only if the signature is valid.

sign/schnorr_test.go
nikkolasg: Yep, will do.

nikkolasg:
> is it better to have a schnorr-structure (like I ask in the challenge), or to have only package-functions?

I have absolutely no arguments in favor of having a struct instead of a slice of bytes but a lot against :D 
 - one should not mess up with a signature anyway so no need to expose its fields
 - it's a new struct that user must learn about, to integrate with their code vs a plain simple slice of byte (  "complexification" of the library )
 - the usage of a struct makes the encoding responsible to the user. that means it can potentially be non valid against EdDSA signature verification and CoSi verification. 
 - we used slice of bytes for CoSi and EdDSA

sign/schnorr.go
ineiti: OK, iff you're all doing mathy stuff, then so be it...

sign/schnorr_test.go
```diff
@@ -0,0 +1,46 @@
...
+	// wrong response
+	wrResp := make([]byte, len(s))
+	copy(wrResp[:32], random.Bytes(32, random.Stream))
+	copy(wrResp[32:], s[32:])
+	assert.Error(t, VerifySchnorr(suite, kp.Public, msg, wrResp))
+
+	// wrong public key
+	wrKp := config.NewKeyPair(suite)
+	assert.Error(t, VerifySchnorr(suite, wrKp.Public, msg, s))
+
```
ineiti: style: delete empty line.

PR #102, opened 2017-02-01
 [Refactoring] Discrete Log Equality zero knowledge proof

nikkolasg:
This PR brings the DLEQ proofs library developed alongside with randhound into the crypto repository. It's needed later for pvss.

### DIFF
--- a/hash/hash.go
+++ b/hash/hash.go
@@ -1,0 +1,101 @@
+// Package hash provides utility functions to process complex data types, like
+// data streams, files, or sequences of structs of different types.
+package hash
+
+import (
+	"bytes"
+	"errors"
+	h "hash"
+	"io"
+	"os"
+
+	"encoding"
+	"reflect"
+)
+
+var defaultChunkSize = 1024
+
+// Stream returns the hash of a data stream separated into chunks of the defaultChunkSize.
+func Stream(hash h.Hash, stream io.Reader) ([]byte, error) {
+	return StreamChunk(hash, stream, defaultChunkSize)
+}
+
+// StreamChunk returns the hash of a data stream separated into chunks of the given byte size.
+func StreamChunk(hash h.Hash, stream io.Reader, size int) ([]byte, error) {
+	if size < 1 {
+		return nil, errors.New("Invalid chunk size")
+	}
+	b := make([]byte, size)
+	for {
+		n, errRead := stream.Read(b)
+		_, err := hash.Write(b[:n])
+		if err != nil {
+			return nil, err
+		}
+		if errRead == io.EOF || n < size {
+			break
+		}
+	}
+	return hash.Sum(nil), nil
+}
+
+// File returns the hash of a file processed as a data stream of chunks of the defaultChunkSize.
+func File(hash h.Hash, stream io.Reader) ([]byte, error) {
+	return StreamChunk(hash, stream, defaultChunkSize)
+}
+
+// FileChunk returns the hash of a file processed as a data stream of chunks with the given byte size.
+func FileChunk(hash h.Hash, file string, size int) ([]byte, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+	return StreamChunk(hash, f, size)
+}
+
+// Args returns the hash of all the given arguments. Each argument has to
+// implement the BinaryMarshaler interface.
+func Args(hash h.Hash, args ...interface{}) ([]byte, error) {
+	var res, buf []byte
+	bmArgs, err := convertToBinaryMarshaler(args)
+	if err != nil {
+		return nil, err
+	}
+	for _, a := range bmArgs {
+		buf, err = a.MarshalBinary()
+		if err != nil {
+			return nil, err
+		}
+		res, err = Stream(hash, bytes.NewReader(buf))
+		if err != nil {
+			return nil, err
+		}
+	}
+	return res, nil
+}
+
+// convertToBinaryMarshaler takes a slice of interfaces and returns
+// a slice of BinaryMarshalers.
+func convertToBinaryMarshaler(args ...interface{}) ([]encoding.BinaryMarshaler, error) {
+	var ret []encoding.BinaryMarshaler
+	for _, a := range args {
+		refl := reflect.ValueOf(a)
+		if refl.Kind() == reflect.Slice {
+			for b := 0; b < refl.Len(); b++ {
+				el := refl.Index(b)
+				bms, err := convertToBinaryMarshaler(el.Interface())
+				if err != nil {
+					return nil, err
+				}
+				ret = append(ret, bms...)
+			}
+		} else {
+			bm, ok := a.(encoding.BinaryMarshaler)
+			if !ok {
+				return nil, errors.New("Could not convert to BinaryMarshaler")
+			}
+			ret = append(ret, bm)
+		}
+	}
+	return ret, nil
+}
--- a/hash/hash_test.go
+++ b/hash/hash_test.go
@@ -1,0 +1,120 @@
+package hash_test
+
+import (
+	"bytes"
+	"crypto/rand"
+	"io/ioutil"
+	"testing"
+
+	"os"
+
+	"github.com/dedis/crypto/abstract"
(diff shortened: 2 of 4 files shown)


hash/hash.go
```diff
@@ -0,0 +1,101 @@
+// Package hash provides utility functions to process complex data types, like
+// data streams, files, or sequences of structs of different types.
+package hash
+
+import (
+	"bytes"
+	"errors"
+	h "hash"
```
ineiti: style: I'd prefer avoiding aliasing packages, it makes it more difficult to understand code and compare it to other places where the package keeps the `hash` name.

hash/hash.go
```diff
@@ -0,0 +1,101 @@
...
+	"os"
+
+	"encoding"
+	"reflect"
+)
+
+var defaultChunkSize = 1024
+
+// Stream returns the hash of a data stream separated into chunks of the defaultChunkSize.
+func Stream(hash h.Hash, stream io.Reader) ([]byte, error) {
```
ineiti: What is the difference with that function and the `File`-function? Can't they be merged?

hash/hash.go
```diff
@@ -0,0 +1,101 @@
...
+	return StreamChunk(hash, stream, defaultChunkSize)
+}
+
+// StreamChunk returns the hash of a data stream separated into chunks of the given byte size.
+func StreamChunk(hash h.Hash, stream io.Reader, size int) ([]byte, error) {
+	if size < 1 {
```
ineiti: can't we use `io.Copy` here? Why do we need to work with chunks? IIRC, this was my code, and I take the blame ;) Now I'd simply use `io.Copy`.

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+import (
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/hash"
+	"github.com/dedis/crypto/random"
+)
+
+// This package provides functionality to create and verify non-interactive
+// zero-knowledge (NIZK) proofs for the equality (EQ) of discrete logarithms (DL).
```
ineiti: doc: could you give some small explanation of the mathematics behind it? Just the formulas that are used to create and verify the proof.

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+
+// DLEQ resembles a NIZK dlog-equality proof. Allows to handle multiple proofs.
```
ineiti: english: resemble - have a similar appearance to or qualities in common with (someone or something); look or seem like.
Probably you wanted to write: assembles. Or even better:

```
Discrete logarithm equality (DLEQ) holds one or more non-interactive zero-knowledge (NIZK) proofs.
```

Also comment the elements of the structure.

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+type DLEQ struct {
+	suite abstract.Suite
+	Base  []DLEQBase
+	Core  []DLEQCore
+}
+
+// DLEQBase contains the base points against which the core proof is created/verified.
+type DLEQBase struct {
```
ineiti: I didn't verify, so I just ask (we need an acronym for that ;): why are the fields of `DLEQBase` private, while the fields of `DLEQCore` are public?

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+	g abstract.Point
+	h abstract.Point
+}
+
+// DLEQCore contains the core elements of the NIZK dlog-equality proof.
+type DLEQCore struct {
+	C  abstract.Scalar // challenge
```
ineiti: doc: what is the final word of commenting fields? I always preferred having the comments _before_ the field.

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+
+// NewDLEQ creates a new NIZK dlog-equality proof.
+func NewDLEQ(suite abstract.Suite, g []abstract.Point, h []abstract.Point, core []DLEQCore) (*DLEQ, error) {
+
+	if len(g) != len(h) {
+		return nil, errors.New("Received non-matching number of points")
+	}
+
+	n := len(g)
+	base := make([]DLEQBase, n)
```
ineiti: style: `base := make([]DLEQBase, len(g))`

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+	for i := range base {
+		base[i] = DLEQBase{g: g[i], h: h[i]}
```
ineiti: I don't know but ask anyway (IDKBAA?):
```
for i, b := range base{
    b = DLEQBase{g: g[i], h:h[]i}
```

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return &DLEQ{suite: suite, Base: base, Core: core}, nil
+}
+
+// Setup initializes the proof by randomly selecting a commitment v,
+// determining the challenge c = H(xG,xH,vG,vH) and the response r = v - cx.
+func (p *DLEQ) Setup(scalar ...abstract.Scalar) ([]abstract.Point, []abstract.Point, error) {
```
ineiti: doc: what goes in and what comes out? I don't understand where the `scalar`s go in the formulas.

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+
+		p.Core[i] = DLEQCore{c, r, vG, vH}
+	}
+
+	return xG, xH, nil
+}
+
+// SetupCollective is similar to Setup with the difference that the challenge
+// is computed as the hash over all base points and commitments.
+func (p *DLEQ) SetupCollective(scalar ...abstract.Scalar) ([]abstract.Point, []abstract.Point, error) {
```
ineiti: doc: again a short description of what happens with the `scalar`s, how many we can give, ...

proof/dleq.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, nil, errors.New("Received unexpected number of scalars")
+	}
+
+	n := len(scalar)
+	p.Core = make([]DLEQCore, n)
+	v := make([]abstract.Scalar, n)
+	xG := make([]abstract.Point, n)
+	xH := make([]abstract.Point, n)
+	vG := make([]abstract.Point, n)
+	vH := make([]abstract.Point, n)
```
ineiti: style: would it be preferable to create a `v,xG, xH, vG, vH`-structure?

proof/dleq_test.go
```diff
@@ -0,0 +1,106 @@
+package proof_test
```
ineiti: If this is to be used widely, I'd add some test that also checks for bad proofs.

hash/hash_test.go
```diff
@@ -0,0 +1,120 @@
...
+			t.Fatal("Could not hash", tmpfile, err)
+		}
+		if len(hash) != 32 {
+			t.Fatal("Output of SHA256 should be 32 bytes")
+		}
+	}
+}
+
+func TestArgs(t *testing.T) {
+
```
ineiti: style: delete empty line

Daeinar:
This PR seems to contain the wrong code. It should have the following content: https://github.com/dedis/crypto/pull/98/files#diff-c658e5e07e591d5648207ad2bf2d5f44

hash/hash.go
Daeinar: @ineiti I will address your issues with `hash.go` in #103.

proof/dleq.go
nikkolasg: I don't really mind, I like when it's clean and neat, and if the comment can fit in one line that's good for me. 
At the same time, when you have struct with some long comments and some short comments, using both styles is bad style. 
What about something that reconciliate everyone ?
 - if all comments can fit in one line, then write it in one line
 - if one comments can't fit in one line, write all comments above

Otherwise, if we want to stick to one general rule, I'd prefer also on top so it avoids conflicts between long and short comments.

proof/dleq.go
nikkolasg: If I'm not mistaken, I'd you can't since range gives you a copy of the value.

Daeinar:
Closing. We will make a new PR starting from #98.

PR #103, opened 2017-02-02
Utility functions for hashing and discrete log equality proofs

Daeinar:
See #96.

### DIFF
--- a/hash/hash.go
+++ b/hash/hash.go
@@ -1,0 +1,101 @@
+// Package hash provides utility functions to process complex data types, like
+// data streams, files, or sequences of structs of different types.
+package hash
+
+import (
+	"bytes"
+	"errors"
+	h "hash"
+	"io"
+	"os"
+
+	"encoding"
+	"reflect"
+)
+
+var defaultChunkSize = 1024
+
+// Stream returns the hash of a data stream separated into chunks of the defaultChunkSize.
+func Stream(hash h.Hash, stream io.Reader) ([]byte, error) {
+	return StreamChunk(hash, stream, defaultChunkSize)
+}
+
+// StreamChunk returns the hash of a data stream separated into chunks of the given byte size.
+func StreamChunk(hash h.Hash, stream io.Reader, size int) ([]byte, error) {
+	if size < 1 {
+		return nil, errors.New("Invalid chunk size")
+	}
+	b := make([]byte, size)
+	for {
+		n, errRead := stream.Read(b)
+		_, err := hash.Write(b[:n])
+		if err != nil {
+			return nil, err
+		}
+		if errRead == io.EOF || n < size {
+			break
+		}
+	}
+	return hash.Sum(nil), nil
+}
+
+// File returns the hash of a file processed as a data stream of chunks of the defaultChunkSize.
+func File(hash h.Hash, stream io.Reader) ([]byte, error) {
+	return StreamChunk(hash, stream, defaultChunkSize)
+}
+
+// FileChunk returns the hash of a file processed as a data stream of chunks with the given byte size.
+func FileChunk(hash h.Hash, file string, size int) ([]byte, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+	return StreamChunk(hash, f, size)
+}
+
+// Args returns the hash of all the given arguments. Each argument has to
+// implement the BinaryMarshaler interface.
+func Args(hash h.Hash, args ...interface{}) ([]byte, error) {
+	var res, buf []byte
+	bmArgs, err := convertToBinaryMarshaler(args)
+	if err != nil {
+		return nil, err
+	}
+	for _, a := range bmArgs {
+		buf, err = a.MarshalBinary()
+		if err != nil {
+			return nil, err
+		}
+		res, err = Stream(hash, bytes.NewReader(buf))
+		if err != nil {
+			return nil, err
+		}
+	}
+	return res, nil
+}
+
+// convertToBinaryMarshaler takes a slice of interfaces and returns
+// a slice of BinaryMarshalers.
+func convertToBinaryMarshaler(args ...interface{}) ([]encoding.BinaryMarshaler, error) {
+	var ret []encoding.BinaryMarshaler
+	for _, a := range args {
+		refl := reflect.ValueOf(a)
+		if refl.Kind() == reflect.Slice {
+			for b := 0; b < refl.Len(); b++ {
+				el := refl.Index(b)
+				bms, err := convertToBinaryMarshaler(el.Interface())
+				if err != nil {
+					return nil, err
+				}
+				ret = append(ret, bms...)
+			}
+		} else {
+			bm, ok := a.(encoding.BinaryMarshaler)
+			if !ok {
+				return nil, errors.New("Could not convert to BinaryMarshaler")
+			}
+			ret = append(ret, bm)
+		}
+	}
+	return ret, nil
+}
--- a/proof/dleq.go
+++ b/proof/dleq.go
@@ -1,0 +1,109 @@
+package proof
+
+import (
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/hash"
+	"github.com/dedis/crypto/random"
+)
+
+// This package provides functionality to create and verify non-interactive
+// zero-knowledge (NIZK) proofs for the equality (EQ) of discrete logarithms (DL).
+
+// DLEQProof represents a NIZK dlog-equality proof.
+type DLEQProof struct {
+	C  abstract.Scalar // challenge
+	R  abstract.Scalar // response
+	VG abstract.Point  // public commitment with respect to base point G
+	VH abstract.Point  // public commitment with respect to base point H
+}
+
+// NewDLEQProof computes a new NIZK dlog-equality proof by randomly selecting a
+// commitment v, determining the challenge c = H(xG,xH,vG,vH) and the response r
+// = v - cx. It also returns the encrypted base points xG and xH.
+func NewDLEQProof(suite abstract.Suite, G abstract.Point, H abstract.Point, x abstract.Scalar) (*DLEQProof, abstract.Point, abstract.Point, error) {
+
+	// Encrypt base points with secret
+	xG := suite.Point().Mul(G, x)
+	xH := suite.Point().Mul(H, x)
+
+	// Commitment
+	v := suite.Scalar().Pick(random.Stream)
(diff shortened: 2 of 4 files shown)


hash/hash.go
```diff
@@ -0,0 +1,101 @@
...
+func FileChunk(hash h.Hash, file string, size int) ([]byte, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+	return StreamChunk(hash, f, size)
+}
+
+// Args returns the hash of all the given arguments. Each argument has to
+// implement the BinaryMarshaler interface.
```
nikkolasg: `... or be a slice of a objects that implement BinaryMarshaler`

proof/dleq.go
```diff
@@ -0,0 +1,109 @@
...
+	C  abstract.Scalar // challenge
+	R  abstract.Scalar // response
+	VG abstract.Point  // public commitment with respect to base point G
+	VH abstract.Point  // public commitment with respect to base point H
+}
+
+// NewDLEQProof computes a new NIZK dlog-equality proof by randomly selecting a
+// commitment v, determining the challenge c = H(xG,xH,vG,vH) and the response r
+// = v - cx. It also returns the encrypted base points xG and xH.
+func NewDLEQProof(suite abstract.Suite, G abstract.Point, H abstract.Point, x abstract.Scalar) (*DLEQProof, abstract.Point, abstract.Point, error) {
```
nikkolasg: Should the G be an argument ? Should it not be the base point generator ? It's true that it can be any point, but usually, it's the base point. That also makes it easier since you don't have to pass one parameter more (also in the network).

proof/dleq.go
```diff
@@ -0,0 +1,109 @@
...
+	r := suite.Scalar()
+	r.Mul(x, c).Sub(v, r)
+
+	return &DLEQProof{c, r, vG, vH}, xG, xH, nil
+}
+
+// NewDLEQProofBatch computes lists of NIZK dlog-equality proofs and of
+// encrypted base points xG and xH. Note that the challenge is computed over all
+// input values.
+func NewDLEQProofBatch(suite abstract.Suite, G []abstract.Point, H []abstract.Point, secrets []abstract.Scalar) ([]*DLEQProof, []abstract.Point, []abstract.Point, error) {
```
nikkolasg: same for here, do you need the G slice ?

proof/dleq.go
```diff
@@ -0,0 +1,109 @@
...
+	rG := suite.Point().Mul(G, p.R)
+	rH := suite.Point().Mul(H, p.R)
+	cxG := suite.Point().Mul(xG, p.C)
+	cxH := suite.Point().Mul(xH, p.C)
+	a := suite.Point().Add(rG, cxG)
+	b := suite.Point().Add(rH, cxH)
+	if p.VG.Equal(a) && p.VH.Equal(b) {
+		return true
+	}
+	return false
```
nikkolasg: Directly
```
return p.VG.Equal(a) && p.VH.Equal(b)
```

proof/dleq.go
```diff
@@ -0,0 +1,109 @@
...
+	return &DLEQProof{c, r, vG, vH}, xG, xH, nil
+}
+
+// NewDLEQProofBatch computes lists of NIZK dlog-equality proofs and of
+// encrypted base points xG and xH. Note that the challenge is computed over all
+// input values.
+func NewDLEQProofBatch(suite abstract.Suite, G []abstract.Point, H []abstract.Point, secrets []abstract.Scalar) ([]*DLEQProof, []abstract.Point, []abstract.Point, error) {
+
+	if len(G) != len(H) && len(H) != len(secrets) {
+		return nil, nil, nil, errors.New("inputs of different lengths")
```
nikkolasg: can you add short lines in your tests to test that ?

hash/hash.go
```diff
@@ -0,0 +1,101 @@
...
+	"errors"
+	h "hash"
+	"io"
+	"os"
+
+	"encoding"
+	"reflect"
+)
+
+var defaultChunkSize = 1024
```
ineiti: IIRC I wrote that thing, so I take most of the blame for this:
- why `Stream` and `File` with the same function-signature?
- why chunking on your own in `StreamChunk` if you can have `io.Copy`?

I liked those functions back then, but more I know about hashing, the less I like those, because it hides a lot of complexity that is source of problems if you want to do the hashing in a cross-language way...

I would keep:
`Stream` that uses `io.Copy` to make a hash of a stream
`File` as convenience function that opens a file and calls `Stream` on it
`Args` (can we find a better name?), even though I'm not that convinced anymore...

proof/dleq.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/hash"
+	"github.com/dedis/crypto/random"
+)
+
+// This package provides functionality to create and verify non-interactive
+// zero-knowledge (NIZK) proofs for the equality (EQ) of discrete logarithms (DL).
```
ineiti: doc: Please add some more info about what you're doing here, with a short 'resume' of the algorithm used.

proof/dleq.go
```diff
@@ -0,0 +1,109 @@
...
+
+// DLEQProof represents a NIZK dlog-equality proof.
+type DLEQProof struct {
+	C  abstract.Scalar // challenge
```
ineiti: sytle: do we want comments also after the fields or only above the fields? I have a tendency for only putting comments before the fields.

proof/dleq.go
ineiti: style: I'd either include xG and xH in the `DLEQProof`, or would not return them here. A function-signature of `(struct, error)` is cleaner than `(struct, point, point, error)`.

If you want to keep the four return values, I propose you name them:
`(proof *DLEQProof, xG abstract.Point, xH abstract.Point, err error)`

proof/dleq.go
ineiti: same here with regard to the parameters - why not putting `xG` and `xH` in `DLEQProof`?

proof/dleq.go
```diff
@@ -0,0 +1,109 @@
...
+// input values.
+func NewDLEQProofBatch(suite abstract.Suite, G []abstract.Point, H []abstract.Point, secrets []abstract.Scalar) ([]*DLEQProof, []abstract.Point, []abstract.Point, error) {
+
+	if len(G) != len(H) && len(H) != len(secrets) {
+		return nil, nil, nil, errors.New("inputs of different lengths")
+	}
+
+	n := len(secrets)
+	proofs := make([]*DLEQProof, n)
+	v := make([]abstract.Scalar, n)
```
ineiti: style: is it possible to either include `v, xG, xH, vG, vH` in the `DLEQProof` structure or make a separate structure? Or is it needed only here?

proof/dleq_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/edwards"
+	"github.com/dedis/crypto/random"
+)
+
+func TestDLEQProof(t *testing.T) {
+
```
ineiti: style: delete empty line

proof/dleq_test.go
```diff
@@ -0,0 +1,81 @@
...
+		} else {
+			bad = append(bad, i)
+		}
+
+	}
+
+	if len(bad) != 0 {
+		t.Fatalf("some proofs are invalid: %v", bad)
+	}
+
```
ineiti: style: delete empty line

proof/dleq_test.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func TestDLEQProofBatch(t *testing.T) {
+
```
ineiti: style: empty line

proof/dleq_test.go
```diff
@@ -0,0 +1,81 @@
...
+			good = append(good, i)
+		} else {
+			bad = append(bad, i)
+		}
+	}
+
+	if len(bad) != 0 {
+		t.Fatalf("some proofs are invalid: %v", bad)
+	}
+
```
ineiti: style: empty line

hash/hash.go
```diff
@@ -0,0 +1,101 @@
+// Package hash provides utility functions to process complex data types, like
+// data streams, files, or sequences of structs of different types.
+package hash
+
+import (
+	"bytes"
+	"errors"
+	h "hash"
```
ineiti: style: I prefer aliasing packages to other names - makes it harder to read the code.

proof/dleq.go
nikkolasg: Ok talked with @Daeinar and it's needed and justifiable. so let it be :)

proof/dleq.go
nikkolasg: Ok talked with @Daeinar and it's needed and justifiable. so let it be :)

proof/dleq.go
Daeinar: Done.

proof/dleq.go
Daeinar: Done

hash/hash.go
Daeinar: Done, except for the renaming of `Args`. Maybe `Arguments` to be in line with `Stream` and `File`?

proof/dleq.go
Daeinar: I prefer short comments after the variables, as it looks cleaner.

proof/dleq.go
Daeinar: Done

proof/dleq.go
Daeinar: Because we don't always need `xG` and `xH`. If we would put them into `DLEQProof` then we would either send around unneeded values or would need to unpack the struct again to only send those values that are needed.

Daeinar:
Note that I addressed the issues raised here in #98. Will merge it back into this one when everything is ready.

Daeinar:
Closing. We will make a new PR starting from #98.

proof/dleq.go
ineiti: Right, I see now that `vG` and `vH` are part of the structure. I'd like to ask you to document some of this - either at the beginning of the file, at the start of the function and/or also here. Like you did for the `// Collective challenge` below.

proof/dleq.go
ineiti: Right - also the go-packages use comments after the field-definition in structs. Still breaks my reading-flow, but I think go-package beats me...

Issue #97, opened 2017-01-20
Refactor poly/sharing.go


Daeinar:
Goals:

- General code cleanup
- Simplify/Clarify API

nikkolasg:
This is the same as #96. closing.

PR #104, opened 2017-02-02
PR: Core secret sharing and PVSS

Daeinar:
Depends on #103.

See #96.

### DIFF
--- a/share/core.go
+++ b/share/core.go
@@ -1,0 +1,315 @@
+// Package share implements different secret sharing mechanisms, including
+// simple Shamir secret sharing, verifiable secret sharing (vss), and public
+// verifiable secret sharing (pvss).
+package share
+
+import (
+	"crypto/cipher"
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+)
+
+// This file provides the core functionality for secret sharing including
+// Shamir's scheme and public commitment polynomials.
+
+// Some error definitions
+var errorGroups = errors.New("non-matching groups")
+var errorCoeffs = errors.New("different number of coefficients")
+
+// PriShare represents a private share.
+type PriShare struct {
+	I int             // Index of the private share
+	V abstract.Scalar // Value of the private share
+}
+
+// PriPoly represents a secret sharing polynomial.
+type PriPoly struct {
+	g      abstract.Group    // Cryptographic group
+	coeffs []abstract.Scalar // Coefficients of the polynomial
+}
+
+// NewPriPoly creates a new secret sharing polynomial for the cryptographic
+// group g, the secret sharing threshold t, and the secret to be shared s.
+func NewPriPoly(g abstract.Group, t int, s abstract.Scalar, rand cipher.Stream) *PriPoly {
+	coeffs := make([]abstract.Scalar, t)
+	coeffs[0] = s
+	if coeffs[0] == nil {
+		coeffs[0] = g.Scalar().Pick(rand)
+	}
+	for i := 1; i < t; i++ {
+		coeffs[i] = g.Scalar().Pick(rand)
+	}
+	return &PriPoly{g, coeffs}
+}
+
+// Threshold returns the secret sharing threshold.
+func (p *PriPoly) Threshold() int {
+	return len(p.coeffs)
+}
+
+// GetSecret returns the shared secret p(0), i.e., the constant term of the polynomial.
+func (p *PriPoly) GetSecret() abstract.Scalar {
+	return p.coeffs[0]
+}
+
+// Eval computes the private share v = p(i).
+func (p *PriPoly) Eval(i int) *PriShare {
+	xi := p.g.Scalar().SetInt64(1 + int64(i)) // x-coordinate of this share
+	v := p.g.Scalar().Zero()
+	for j := p.Threshold() - 1; j >= 0; j-- {
+		v.Mul(v, xi)
+		v.Add(v, p.coeffs[j])
+	}
+	return &PriShare{i, v}
+}
+
+// Shares creates a list of n private shares p(1),...,p(n).
+func (p *PriPoly) Shares(n int) []*PriShare {
+	shares := make([]*PriShare, n)
+	for i := 0; i < n; i++ {
+		shares[i] = p.Eval(i)
+	}
+	return shares
+}
+
+// Add computes the component-wise sum of the polynomials p and q and returns it
+// as a new polynomial.
+func (p *PriPoly) Add(q *PriPoly) (*PriPoly, error) {
+
+	if p.g != q.g {
+		return nil, errorGroups
+	}
+
+	if p.Threshold() != q.Threshold() {
+		return nil, errorCoeffs
+	}
+
+	t := p.Threshold()
+	coeffs := make([]abstract.Scalar, t)
+	for i := 0; i < t; i++ {
+		coeffs[i] = p.g.Scalar().Add(p.coeffs[i], q.coeffs[i])
+	}
+
+	return &PriPoly{p.g, coeffs}, nil
+}
+
+// Equal checks equality of two secret sharing polynomials p and q.
+func (p *PriPoly) Equal(q *PriPoly) bool {
+
+	if p.g.String() != q.g.String() {
+		return false
+	}
+
+	for i := 0; i < p.Threshold(); i++ {
+		if !p.coeffs[i].Equal(q.coeffs[i]) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// Commit creates a public commitment polynomial for the given base point b or
+// the standard base if b == nil.
+func (p *PriPoly) Commit(b abstract.Point) *PubPoly {
+	t := p.Threshold()
+	commits := make([]abstract.Point, t)
+	for i := 0; i < t; i++ {
+		commits[i] = p.g.Point().Mul(b, p.coeffs[i])
+	}
+	return &PubPoly{p.g, b, commits}
+}
+
+// RecoverSecret reconstructs the shared secret p(0) using Lagrange interpolation.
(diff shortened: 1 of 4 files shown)


share/core.go
```diff
@@ -0,0 +1,315 @@
...
+
+import (
+	"crypto/cipher"
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+)
+
+// This file provides the core functionality for secret sharing including
+// Shamir's scheme and public commitment polynomials.
```
nikkolasg: this comment should be at the top level also, because it is not shown otherwise in the godoc (IIRC)

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+	return &PriPoly{g, coeffs}
+}
+
+// Threshold returns the secret sharing threshold.
+func (p *PriPoly) Threshold() int {
+	return len(p.coeffs)
+}
+
+// GetSecret returns the shared secret p(0), i.e., the constant term of the polynomial.
+func (p *PriPoly) GetSecret() abstract.Scalar {
```
nikkolasg: I'd rename that simply `Secret`. In fact I'm not even sure I'd put that as a function, because why would you need to get the secret when you are the one that built the pripoly...  but that's up to debate ;)

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+	t := p.Threshold()
+	coeffs := make([]abstract.Scalar, t)
+	for i := 0; i < t; i++ {
+		coeffs[i] = p.g.Scalar().Add(p.coeffs[i], q.coeffs[i])
+	}
+
+	return &PriPoly{p.g, coeffs}, nil
+}
+
+// Equal checks equality of two secret sharing polynomials p and q.
```
nikkolasg: Is it relevant to say here that's it's not constant time ? 
Since the secret is at position 0 knowing if both private polynomials have the same secret is easy by recording the time they take to compare.
Should we provide a constant time implementation ?

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+		return s == nil || s.V == nil || s.I < 0 || n <= s.I
+	}
+
+	c := 0
+	x := make(map[int]abstract.Scalar)
+	for _, s := range shares {
+		if isBad(s) {
+			continue
+		}
+		x[s.I] = g.Scalar().SetInt64(1 + int64(s.I))
```
nikkolasg: You should  not need to declare the `isBad` function really, since you only need here. 
Using like that:
```
for i, s := range shares {
     if s == nil || s.V == nil || s.I < 0 || n <= s.I {
         continue
     }
     x[i] = g.Scalar().SetInt64(1 + int64(s.I))
}
```
then, later, you can directly iterate over *only* the good shares (vs *all* the shares):
```
for i,si := range x { // line 150
    // no need for isBad here
    for j,sj := range x {
        if j == i { // no need for isBad here
            continue
        }
        ... // same thing here
    }
}
```

Sorry maybe I confused you with my suggestion of using a separate function. In fact, my idea was to use a common function that retrieves the good shares whether using `PriShare`s or `PubShare`s but when I tried to implement it did not look as good as I thought :D

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+		c++
+	}
+
+	if c < t {
```
nikkolasg: this c should not be necessary anymore; with the above approach a simple `len(shares)` is enough.

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+	for i := 0; i < n; i++ {
+		shares[i] = p.Eval(i)
+	}
+	return shares
+}
+
+// Add computes the component-wise sum of the polynomials p and q and returns it
+// as a new polynomial. If the base points p.b and q.b are different then the
+// base point of the resulting PubPoly cannot be computed without knowing the
+// discrete logarithm between p.b and q.b. In this particular case, we are using
```
nikkolasg: I like to emphasize these kinds of things with a big `NOTE` or `WARNING` so it's clearly visible. what do you think ?

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+func RecoverCommit(g abstract.Group, shares []*PubShare, t int, n int) (abstract.Point, error) {
+
+	isBad := func(s *PubShare) bool {
+		return s == nil || s.V == nil || s.I < 0 || n <= s.I
+	}
+
+	c := 0
+	x := make(map[int]abstract.Scalar)
+	for _, s := range shares {
+		if isBad(s) {
```
nikkolasg: the same reasoning as for `PriShare` applies here.

share/pvss.go
```diff
@@ -0,0 +1,174 @@
...
+
+import (
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/proof"
+	"github.com/dedis/crypto/random"
+)
+
+// This package implements public verifiable secret sharing as introduced by
```
nikkolasg: This must go at the top level so it is shown in godoc (IIRC)

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+		v.Mul(v, xi)
+		v.Add(v, p.coeffs[j])
+	}
+	return &PriShare{i, v}
+}
+
+// Shares creates a list of n private shares p(1),...,p(n).
+func (p *PriPoly) Shares(n int) []*PriShare {
+	shares := make([]*PriShare, n)
+	for i := 0; i < n; i++ {
```
ineiti: joke: `for i := range make([]byte, n)`

nikkolasg: Some kind of map/reduce programming, only imagine...
`0.upto(n).map([]*PriShare, i => suite.Scalar().Pick(random))`

share/core.go
Daeinar: Note that it's not sufficient to `range` over the `x` coordinates here as we need the value of the share `si.V` in the outer loop as well. So either we create some mapping between the `x` coordinates and the shares themselves or we go twice over the shares as I did.

share/core.go
Daeinar: This function is nice for testing purposes. So I would prefer that we keep it. +1 for renaming.

share/core.go
Daeinar: Done, although I am not sure that the approach I did for `PubPoly.Equal()` is good, i.e. to use `MarshalBinary()` on its coefficients to get `[]byte` for comparison.

share/core.go
nikkolasg: yep , taking back the last code snippet:
```
for i,si := range x { // line 150
    vi := shares[i]
    for j,sj := range x {
        ...    
        vj := shares[j]
    }
}
```

share/core.go
nikkolasg: Why not just simply checking all the coefficients and return the result after ?

share/core.go
nikkolasg: >This function is nice for testing purposes.

our tests are in the same package, so you don't need exported function to get the secret, you can directly do `p.secret`

nikkolasg: But again, if you feel it's something that might be useful, then let's keep it. I don't mind.

share/core.go
Daeinar: Not sure I understand what you mean, because checking all coefficients (or commits in this particular case) is exactly what I did. Or do you suggest I use `P.Equal(Q)` for points/commits `P` and `Q`?

share/core.go
Daeinar: I think that this unfortunately doesn't work either. Note that the index `i` here corresponds to the share index `s.I`, which is the key in the map `x`. (The values of the map `x` are the x coordinates of the share `s`). But this index `i` (or `s.I`) does not necessarily coincide with the index of `s` in the slice `[]shares`. Meaning: `shares[i]` might give you the wrong share and in the worst case even an out-of-bounds error.

nikkolasg: Look again the first snippet of code; the relevant line in question is now:
```
x[i] = g.Scalar().SetInt64(1 + int64(s.I))
```
The key is *not* the `s.I` !  `x` now contains the indexes of the good shares  in `shares` as keys, and the corresponding x coordinates as values. That's why doing `shares[i]` works :)

share/core.go
nikkolasg: Well sorry, let me clarify:
My comment was about "should we provide constant time impl. or not ? if not, we should say it explicitely". You seemed to imply by your reply that you did a constant time impl. (so we're ok for doing that) but using the marshal representation. 
What I suggested is simply iterate over all points and checking if they are different, but *not* returning when one is different (what's currently done), only when all points have been checked <=> constant time.

Daeinar: Well but then the question becomes if `P.Equal(Q)` is constant time, for `P` and `Q` coefficients/points. Using the binary representation of `P` and `Q` and the `subtle` package should ensure that also this comparison is constant time.

share/core.go
Daeinar: Ah right, sorry I didn't see that. Then it of course works. :)

Daeinar:
Closing. We will make a new PR starting from #97.

share/core.go
ineiti: I think ruby would allow for

```
(0..n).collect{|i| Eval(i)}
```

share/core.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	if c < t {
+		return nil, errors.New("not enough good public shares to reconstruct secret commitment")
+	}
+
+	num := g.Scalar()       // numerator
+	den := g.Scalar()       // denominator
+	tmp := g.Scalar()       // scalar buffer
+	Acc := g.Point().Null() // point accumulator
```
ineiti: Style: why is `Acc` and `Tmp` upper-case? I prefer all lower-case in go-functions.

share/core.go
ineiti: Or even:
`(0..n).collect{:Eval}`
IIRC...

PR #106, opened 2017-02-06
Secret Sharing Core (1/5)

Daeinar:
This PR integrates Shamir secret sharing and public commitment polynomials.


share/core.go
```diff
@@ -0,0 +1,282 @@
...
+	return p.b, p.commits
+}
+
+// Threshold returns the secret sharing threshold.
+func (p *PubPoly) Threshold() int {
+	return len(p.commits)
+}
+
+// GetCommit returns the secret commitment p(0), i.e., the constant term of the polynomial.
+func (p *PubPoly) GetCommit() abstract.Point {
```
nikkolasg: Renaming to `Commit` is enough, is consistent with other methods (`Threshold` and `Info` vs `GetThreshold` and `GetInfo`) and is shorter :)

share/core.go
```diff
@@ -0,0 +1,282 @@
...
+		v.Mul(v, xi)
+		v.Add(v, p.coeffs[j])
+	}
+	return &PriShare{i, v}
+}
+
+// Shares creates a list of n private shares p(1),...,p(n).
+func (p *PriPoly) Shares(n int) []*PriShare {
+	shares := make([]*PriShare, n)
+	for i := 0; i < n; i++ {
```
ineiti: style: `for i := range shares`

share/core.go
```diff
@@ -0,0 +1,282 @@
...
+		v.Mul(v, xi)
+		v.Add(v, p.commits[j])
+	}
+	return &PubShare{i, v}
+}
+
+// Shares creates a list of n public commitment shares p(1),...,p(n).
+func (p *PubPoly) Shares(n int) []*PubShare {
+	shares := make([]*PubShare, n)
+	for i := 0; i < n; i++ {
```
ineiti: `for i := range shares`

ineiti:
LGTM: twice `for ... range` and still don't know whether you need to compare `Group.String()` or if you can compare `Group`s directly (don't find the lines - network too slow ;)

Daeinar:
I decided to compare `Group.String()`. Comparing two groups via `g1 == g2` should compare the pointers which could lead to weird outcomes, I guess.

ineiti:
According to http://stackoverflow.com/questions/34245932/golang-checking-equality-of-interface this should be possible, but looking at how crypto handles the groups, I'm not sure. So `String` is surely better here!

share/core.go
```diff
@@ -0,0 +1,282 @@
...
+	commits := make([]abstract.Point, p.Threshold())
+	for i := range commits {
+		commits[i] = p.g.Point().Mul(b, p.coeffs[i])
+	}
+	return &PubPoly{p.g, b, commits}
+}
+
+// RecoverSecret reconstructs the shared secret p(0) from a list of private
+// shares using Lagrange interpolation.
+func RecoverSecret(g abstract.Group, shares []*PriShare, t int, n int) (abstract.Scalar, error) {
```
nikkolasg: One last nitpick and I'm merging so we can move forward: 
`n, t int` instead of `n int, t int` :)

share/core.go
```diff
@@ -0,0 +1,282 @@
+// Package share implements simple Shamir secret sharing and public
+// commitment polynomials.
```
ineiti: doc: can you give one or two pointers to wikipedia/paper?

Daeinar: I'd prefer to put such links into an associated README.

share/core.go
Daeinar: @nikkolasg: Done.

share/core.go
ineiti: OK - but a second paragraph, so it looks nicer in `godoc`?

Daeinar:
If there are no further concerns, then I think we can merge the PR.

ineiti:
I'd still like to have another paragraph for godoc. But if you think it's too much trouble, I'll merge it as-is.

Daeinar:
@ineiti What do you mean by "another paragraph for godoc"?

ineiti:
Yes, the documentation for the package. As this is something that other people also might want to use, I think it's a good idea of giving either a simple use-case or explaining a bit further what it does. You and Nicolas know why and how it works, but I and others don't.

PR #107, opened 2017-02-06
Utility Functions for Hashing (2/5)

Daeinar:
Some utility functions for hashing that are required later on.

PR #108, opened 2017-02-06
Discrete Logarithm Zero Knowledge Proofs (3/5)

Daeinar:
Depends on #107.

### DIFF
--- a/proof/dleq.go
+++ b/proof/dleq.go
@@ -1,3 +1,8 @@
+// This package provides functionality to create and verify non-interactive
+// zero-knowledge (NIZK) proofs for the equality (EQ) of discrete logarithms (DL).
+// This means, for two values xG and xH one can check that
+//   log_{G}(xG) == log_{H}(xH)
+// without revealing the secret value x.
 package proof
 
 import (
@@ -8,12 +13,8 @@ import (
 	"github.com/dedis/crypto/random"
 )
 
-// This package provides functionality to create and verify non-interactive
-// zero-knowledge (NIZK) proofs for the equality (EQ) of discrete logarithms (DL).
-// This means, for two values xG and xH one can check that log_{G}(xG) == log_{H}(xH)
-// without revealing the secret value x.
-
 var errorDifferentLengths = errors.New("inputs of different lengths")
+var errorInvalidProof = errors.New("invalid proof")
 
 // DLEQProof represents a NIZK dlog-equality proof.
 type DLEQProof struct {
@@ -24,8 +25,10 @@ type DLEQProof struct {
 }
 
 // NewDLEQProof computes a new NIZK dlog-equality proof by randomly selecting a
-// commitment v, determining the challenge c = H(xG,xH,vG,vH) and the response r
-// = v - cx. It also returns the encrypted base points xG and xH.
+// commitment v, determining the challenge and the response r:
+//   c = H(xG,xH,vG,vH)
+//   r = v - cx
+// It also returns the encrypted base points xG and xH.
 func NewDLEQProof(suite abstract.Suite, G abstract.Point, H abstract.Point, x abstract.Scalar) (proof *DLEQProof, xG abstract.Point, xH abstract.Point, err error) {
 	// Encrypt base points with secret
 	xG = suite.Point().Mul(G, x)
@@ -67,7 +70,6 @@ func NewDLEQProofBatch(suite abstract.Suite, G []abstract.Point, H []abstract.Po
 	vH := make([]abstract.Point, n)
 
 	for i, x := range secrets {
-
 		// Encrypt base points with secrets
 		xG[i] = suite.Point().Mul(G[i], x)
 		xH[i] = suite.Point().Mul(H[i], x)
@@ -95,14 +97,19 @@ func NewDLEQProofBatch(suite abstract.Suite, G []abstract.Point, H []abstract.Po
 	return proofs, xG, xH, nil
 }
 
-// Verify examines the validity of the corresponding NIZK dlog-equality proof
-// by checking that vG == rG + c(xG) and vH == rH + c(xH).
-func (p *DLEQProof) Verify(suite abstract.Suite, G abstract.Point, H abstract.Point, xG abstract.Point, xH abstract.Point) bool {
+// Verify examines the validity of the NIZK dlog-equality proof.
+// The proof is valid if the following two conditions hold:
+//   vG == rG + c(xG)
+//   vH == rH + c(xH)
+func (p *DLEQProof) Verify(suite abstract.Suite, G abstract.Point, H abstract.Point, xG abstract.Point, xH abstract.Point) error {
 	rG := suite.Point().Mul(G, p.R)
 	rH := suite.Point().Mul(H, p.R)
 	cxG := suite.Point().Mul(xG, p.C)
 	cxH := suite.Point().Mul(xH, p.C)
 	a := suite.Point().Add(rG, cxG)
 	b := suite.Point().Add(rH, cxH)
-	return p.VG.Equal(a) && p.VH.Equal(b)
+	if !(p.VG.Equal(a) && p.VH.Equal(b)) {
+		return errorInvalidProof
+	}
+	return nil
 }
--- a/proof/dleq.go
+++ b/proof/dleq.go
@@ -1,0 +1,108 @@
+package proof
+
+import (
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/hash"
+	"github.com/dedis/crypto/random"
+)
+
+// This package provides functionality to create and verify non-interactive
+// zero-knowledge (NIZK) proofs for the equality (EQ) of discrete logarithms (DL).
+// This means, for two values xG and xH one can check that log_{G}(xG) == log_{H}(xH)
+// without revealing the secret value x.
+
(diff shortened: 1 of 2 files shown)


proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"errors"
+
+	"github.com/dedis/crypto/abstract"
+	"github.com/dedis/crypto/hash"
+	"github.com/dedis/crypto/random"
+)
+
+// This package provides functionality to create and verify non-interactive
```
ineiti: https://blog.golang.org/godoc-documenting-go-code

Comments come _before_ the `package`, `struct`, `func`.

proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+
+// DLEQProof represents a NIZK dlog-equality proof.
+type DLEQProof struct {
+	C  abstract.Scalar // challenge
+	R  abstract.Scalar // response
+	VG abstract.Point  // public commitment with respect to base point G
+	VH abstract.Point  // public commitment with respect to base point H
+}
+
+// NewDLEQProof computes a new NIZK dlog-equality proof by randomly selecting a
```
ineiti: doc: if you want to make it really nice, remember that godoc is very simple. So it could be written like this:

```
// NewDLEQProof computes a new NIZK dlog-equality proof by randomly selecting a
// commitment v, determining the challenge c and the response r:
//   c = H(xG,xH,vG,vH)
//   r = v - cx. 
// It also returns the encrypted base points xG and xH.
```

proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return &DLEQProof{c, r, vG, vH}, xG, xH, nil
+}
+
+// NewDLEQProofBatch computes lists of NIZK dlog-equality proofs and of
+// encrypted base points xG and xH. Note that the challenge is computed over all
+// input values.
+func NewDLEQProofBatch(suite abstract.Suite, G []abstract.Point, H []abstract.Point, secrets []abstract.Scalar) (proof []*DLEQProof, xG []abstract.Point, xH []abstract.Point, err error) {
+	if len(G) != len(H) || len(H) != len(secrets) {
+		return nil, nil, nil, errorDifferentLengths
```
ineiti: @nikkolasg - would
```
err = errorDifferentLengths
return
```

be nicer here?

proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+	n := len(secrets)
+	proofs := make([]*DLEQProof, n)
+	v := make([]abstract.Scalar, n)
+	xG = make([]abstract.Point, n)
+	xH = make([]abstract.Point, n)
+	vG := make([]abstract.Point, n)
+	vH := make([]abstract.Point, n)
+
+	for i, x := range secrets {
+
```
ineiti: style: unneeded empty line

proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+		// Commitments
+		v[i] = suite.Scalar().Pick(random.Stream)
+		vG[i] = suite.Point().Mul(G[i], v[i])
+		vH[i] = suite.Point().Mul(H[i], v[i])
+	}
+
+	// Collective challenge
+	cb, err := hash.Structures(suite.Hash(), xG, xH, vG, vH)
+	if err != nil {
+		return nil, nil, nil, err
```
ineiti: `return`

proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+	c := suite.Scalar().Pick(suite.Cipher(cb))
+
+	// Responses
+	for i, x := range secrets {
+		r := suite.Scalar()
+		r.Mul(x, c).Sub(v[i], r)
+		proofs[i] = &DLEQProof{c, r, vG[i], vH[i]}
+	}
+
+	return proofs, xG, xH, nil
```
ineiti: `return`

proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Verify examines the validity of the corresponding NIZK dlog-equality proof
+// by checking that vG == rG + c(xG) and vH == rH + c(xH).
```
ineiti: doc: I'd put the formulas on a separate line, indented by two "spaces", so that they'll be code-formatted.

proof/dleq.go
```diff
@@ -0,0 +1,108 @@
...
+func (p *DLEQProof) Verify(suite abstract.Suite, G abstract.Point, H abstract.Point, xG abstract.Point, xH abstract.Point) bool {
```
ineiti: question: Other `Verify`-functions return `nil` if the verification passes or an error if it doesn't pass. Should we do the same here or change the name? `Check`?

nikkolasg: Good question. I have a preference for your first option (return an error) but I'll let @Daeinar decide...

proof/dleq.go
nikkolasg: if the same conditions are used somewhere yes, otherwise I'd say it's a bit much :)

nikkolasg: lol that's a very low level nitpick question... I'm not sure, I'm not a fan of named arguments, but here it's needed since we must differ between xG and xH. Since I'm not used to it, I don't know the best practices for that. If you think this is better, why not. In my part, due to my incomplete knowledge, I'd go for the shorter version (so what @Daeinar proposed).

Daeinar: I am not sure that I am a fan of named return arguments either. Even the Go people say that they should be used only in short functions, see https://tour.golang.org/basics/7. I would prefer going back to the unnamed versions. And having naked `return` does not improve readability at all in my opinion as it's very Golang specific. So people who do not know all the specialities of Go (including myself) might wonder what's going on here.

Daeinar:
@ineiti @nikkolasg This is also ready to be merged.

proof/dleq.go
ineiti: For me `return` is more readable than `return nil, nil, nil, errorDifferentLengths`. But as a minority I think of better arguments.

proof/dleq_test.go
```diff
@@ -0,0 +1,88 @@
...
+	var bad []int
+
+	for i := 0; i < n; i++ {
+		// Create some random secrets and base points
+		x := suite.Scalar().Pick(random.Stream)
+		g, _ := suite.Point().Pick([]byte(fmt.Sprintf("G%d", i)), random.Stream)
+		h, _ := suite.Point().Pick([]byte(fmt.Sprintf("H%d", i)), random.Stream)
+
+		proof, xG, xH, err := NewDLEQProof(suite, g, h, x)
+		if err != nil {
```
ineiti: `require.Nil`

proof/dleq_test.go
```diff
@@ -0,0 +1,88 @@
...
+	h := make([]abstract.Point, n)
+
+	for i := 0; i < n; i++ {
+		x[i] = suite.Scalar().Pick(random.Stream)
+		g[i], _ = suite.Point().Pick([]byte(fmt.Sprintf("G%d", i)), random.Stream)
+		h[i], _ = suite.Point().Pick([]byte(fmt.Sprintf("H%d", i)), random.Stream)
+	}
+
+	proofs, xG, xH, err := NewDLEQProofBatch(suite, g, h, x)
+	if err != nil {
```
ineiti: `require.Nil`

proof/dleq_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func TestDLEQLengths(t *testing.T) {
+	suite := edwards.NewAES128SHA256Ed25519(false)
+	n := 10
+	x := make([]abstract.Scalar, n)
+	g := make([]abstract.Point, n)
+	h := make([]abstract.Point, n)
+
+	for i := 0; i < n; i++ {
```
ineiti: `for i := range x`

proof/dleq_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	proofs, xG, xH, err := NewDLEQProofBatch(suite, g, h, x)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var good []int
+	var bad []int
+
+	for i := 0; i < n; i++ {
```
ineiti: `for i, proof := range proofs`

proof/dleq_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func TestDLEQProofBatch(t *testing.T) {
+	suite := edwards.NewAES128SHA256Ed25519(false)
+	n := 10
+	x := make([]abstract.Scalar, n)
+	g := make([]abstract.Point, n)
+	h := make([]abstract.Point, n)
+
+	for i := 0; i < n; i++ {
```
ineiti: `for i := range x`

proof/dleq_test.go
```diff
@@ -0,0 +1,88 @@
...
+			t.Fatal(err)
+		}
+
+		if err := proof.Verify(suite, g, h, xG, xH); err == nil {
+			good = append(good, i)
+		} else {
+			bad = append(bad, i)
+		}
+	}
+	require.Equal(t, len(bad), 0)
```
ineiti: if you don't need `good` and `bad` for anything else, I'd just `require.Nil(proof.Verify` above and don't use `good` and `bad` at all. Those are very old concepts anyway which are very relative ;)

ineiti:
Cool! Merge if you like.

PR #118, opened 2017-02-19
added gcd() method

lihiid:
I needed gcd() for Pohlig Hellman.  I Changed abstract/group.go(added gcd signature) and nist/int.go(added gcd implementation).


ed25519/point.go
```diff
@@ -20,9 +20,9 @@ import (
 	"errors"
 	"io"
 
-	"github.com/dedis/crypto/abstract"
-	"github.com/dedis/crypto/group"
-	"github.com/dedis/crypto/nist"
+	"gopkg.in/dedis/crypto.v0/abstract"
```
ineiti: This should not show up here - only depend on `dedis/crypto`

ineiti:
I propose you do the following:

```
git checkout master
git pull
git checkout -b gcd_119
git checkout ppsi_651 abstract/group.go nist/int.go
git push -u origin gcd_119
```

ineiti:
Another comment: I think now _every_ implementation of `Group` needs to implement `Gcd`. Does this even make sense?

abstract/group.go
```diff
@@ -49,6 +49,9 @@ type Scalar interface {
 
 	// Set to the modular inverse of scalar a
 	Inv(a Scalar) Scalar
+	
+	// Set to the gcd of a and its module
+	Gcd(a Scalar) Scalar
```
nikkolasg: I am not a fan of this method because it is not consistent with the other methods, as it only takes one argument whereas it should at least take two (`GCD` takes two numbers, `Add` takes two numbers, vs `Inv` takes one number). I see currently two options:
+ Change the signature to 
```go
Gcd(a Scalar, b Scalar) Scalar
```
+ Put a method the `nist/` package, that takes two `Scalar` and returns the GCD out of it.
```go
nist.GCD(a,b) Scalar
```
or
```go
abstract.GCD(a,b) Scalar
```
I think the first option is cleaner, but @bford may not like it, in that case the second option is fine to me. @ineiti what do you think ?

Now the other questions is whether to follow the golang `big.Int` [signature](https://golang.org/pkg/math/big/#Int.GCD), which can also take the `x` and `y` parameters. I think it is ok to let as is, but it depends on the application. Sounds like @lihiid you don't need these values so it make sense to let as-is for now.

lihiid: Have you seen the inv method?I haven't written it (it existed before I wrote the gcd()) and it receives only one parameter, so I thought it was ok to add another method which receives only one parameter.  I can change it to receive two arguments. Anyway, I dont think that the second option is very generic so it is better to leave it in abstract as it is now. Regarding x and y-I dont need them for my protocol, so I didn't included them so I think I will leave it as it is unless you think differently.

nikkolasg: The main reasons the whole interface of `Scalar` is defined like this is to avoid un-necessary allocations. You can for example use a `var tmp = suite.Scalar()` which holds  different steps of a computations without changing or cloning important scalars such as a private key etc. This reasoning is from `big.Int` library, and applies here as well. So for a GCD, it should be the same,i.e. it should be possible to provide both parameters in the arguments and let the caller decides which scalar will be the recipient.
I'm ok with keeping that in the `Scalar` interface also.
For the x and y then it's fine for me also :)

bford:
Lihi, why exactly do you need GCD for Pohlig-Hellman?  I thought that just required scalar multiply for encryption and multiplicative inverse for decryption?

lihiid:
@nikkolasg-I changed the gcd() so it takes two parameters. Can you have a look at the ppsi_651 branch(nist/int.go)  and tell me if we can proceed? It won't run on the ppsi_651 branch or lihiid/ppsi because of collisions in imports(crypto/crypto.v0), however when I tested the gcd() on my env. **as part of crypto.v0**(simply added the gcd() to crypto.v0/nist/int.go, and same for crypto.v0/abstract/group.go, and all my imports of crypto only included crypto.v0) it worked very well. Thanks.
@bford-I wanted the Pohlig Hellman implementation (choose e s.t gcd(e,N)=1 where e is encryption key and N is the order) to be general and to also work for non-prime order curves, as this is the impression I got  from the existence of
func (c *Curve) PrimeOrder() bool 
in abstract/group.go

ineiti:
Closing

PR #125, opened 2017-03-07
[Share] VSS pedersen scheme implementation (5/7)

nikkolasg:
This PR bring Verifiable Secret Sharing scheme from Pedersen & Rabin.

Note: re-created PR since it added share.go because it was an old PR..
Old pr: https://github.com/dedis/crypto/pull/110


share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+		}
+	}
+	return d, nil
+}
+
+// Deals returns the list of previously generated deals.
+func (d *Dealer) Deals() []*Deal {
+	out := make([]*Deal, len(d.deals))
+	copy(out, d.deals)
+	return out
```
ineiti: Hmm - this doesn't do a deep copy, so if the user modifies the returned value, he will modify the underlying structure. So why do the shallow copy here?

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+// a default safe value. If a different t value is required, it is possible to set
+// it with `verifier.SetT()`.
+func NewVerifier(suite abstract.Suite, longterm abstract.Scalar, dealerKey abstract.Point,
+	verifiers []abstract.Point) (*Verifier, error) {
+
+	pub := suite.Point().Mul(nil, longterm)
+	var ok bool
+	var index int
+	for i := range verifiers {
+		if verifiers[i].Equal(pub) {
```
ineiti: thanks for the range - I'd put it as:

```
for _, v := range verifiers{
  if v.Equal(pub) {
```

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+
+	if err := sign.VerifySchnorr(a.suite, a.dealer, msgDeal(d), d.Signature); err != nil {
+		return err
+	}
+
+	fi := d.SecShare
+	gi := d.RndShare
+	if fi.I != gi.I {
+		return errors.New("vss: not the same index for f and g share in Deal")
+	} else if fi.I < 0 || fi.I >= len(a.verifiers) {
```
ineiti: style: I would remove the `else` and put the `if` on the next line - it is easier to read than the `} else if fi`...

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+}
+
+func (a *aggregator) verifyJustification(j *Justification) error {
+	if _, ok := findPub(a.verifiers, j.Index); !ok {
+		return errors.New("vss: index out of bounds in justification")
+	}
+	r, ok := a.responses[j.Index]
+	if !ok {
+		return errors.New("vss: no complaints received for this justification")
+	} else if r.Status != StatusComplaint {
```
ineiti: style: same as above: delete the `else` and put the `if` on the next line.

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+// DealCertified returns true if there has been less than t complaints, all
+// Justifications were correct and if EnoughApprovals() returns true.
+func (a *aggregator) DealCertified() bool {
+	var comps int
+	for _, r := range a.responses {
+		if r.Status == StatusComplaint {
+			comps++
+		}
+	}
+	return a.EnoughApprovals() && !(comps >= a.t || a.badDealer)
```
ineiti: style: as this is not `C`, I would put the `!(comps...)` into its own variable and give it a meaningful name.

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+}
+
+// MinimumT the minimum safe T that should be used with this protocol as it's
+// proven to be secure with such a T. WARNING: Setting a lower T could make
+// the whole protocol insecure. Setting a higher T only makes it harder to
+// reconstruct the secret.
```
ineiti: doc: Write what `n` is in the comment.

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+	return (n + 1) / 2
+}
+
+func validT(t int, verifiers []abstract.Point) bool {
+	return t >= 2 && t <= len(verifiers) && int(uint32(t)) == t
+}
+
+func deriveH(suite abstract.Suite, verifiers []abstract.Point) abstract.Point {
+	var b bytes.Buffer
+	for i := range verifiers {
```
ineiti: `for _, v := range verifiers`

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+	}
+	h := suite.Hash()
+	h.Write(b.Bytes())
+	digest := h.Sum(nil)
+	base, _ := suite.Point().Pick(nil, suite.Cipher(digest))
+	return base
+}
+
+func findPub(verifiers []abstract.Point, idx uint32) (abstract.Point, bool) {
+	iidx := int(idx)
```
ineiti: gofail: why wan't we use `uint32` as index and compare `uint32` and `int`?

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+		return nil, false
+	}
+	return verifiers[iidx], true
+}
+
+func sessionID(suite abstract.Suite, dealer abstract.Point, verifiers, commitments []abstract.Point, t int) ([]byte, error) {
+	h := suite.Hash()
+	dealer.MarshalTo(h)
+
+	for i := range verifiers {
```
ineiti: `for _, v := range verifiers`

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+		verifiers[i].MarshalTo(h)
+	}
+
+	for i := range commitments {
```
ineiti: `for _, c := range commitments`

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+		commitments[i].MarshalTo(h)
+	}
+	binary.Write(h, binary.LittleEndian, uint32(t))
+
+	return h.Sum(nil), nil
+}
+
+func msgResponse(r *Response) []byte {
```
ineiti: question: I see you're using this for the signature. Is this the most secure way to do this? why are you hand-rolling your marshalling here? Would some `fmt.Sprintf` not work?
question: would some `responseHash` be more appropriate?

ineiti:
I looked at the coverage and saw that quite some methods are not used at all - will they be later?

- func (d *Dealer) Commits() []abstract.Point {
- func (d *Dealer) Key() (abstract.Scalar, abstract.Point) {
- func (d *Dealer) SessionID() []byte {
- func (v *Verifier) Key() (abstract.Scalar, abstract.Point) {
- func (v *Verifier) Index() int {
- func (v *Verifier) SessionID() []byte {
- func (a *aggregator) setDeal(d *Deal) {

share/vss/vss.go
```diff
@@ -0,0 +1,630 @@
...
+		suite:     suite,
+		long:      longterm,
+		secret:    secret,
+		verifiers: verifiers,
+	}
+	if !validT(t, verifiers) {
+		return nil, fmt.Errorf("dealer: t %d invalid", t)
+	}
+
+	d.t = t
```
nikkolasg: because the t is being verified just above. I just prefer setting fields once they're validated even if the end result is exactly the same...

share/vss/vss.go
nikkolasg: 1. Well I have to sign my messages in a deterministic way so verifiers can produce the same hash, so I would prefer the marshalling to be explicit and visible and language agnostic instead of using some Golang specific construction.
2. Indeed, I hesitated to use a Hash here. Maybe I should, although I have not enough expertise to know whatever it brings more to the table...

nikkolasg:
@ineiti Indeed, there has been some that proved to be useless. This is the list of methods I use in the DKG code (distributed key generation):

- `func (d *Dealer) Commits() []abstract.Point `
- `func (d *Dealer) SessionID() []byte `
- `func (v *Verifier) SessionID() []byte `

These are the ones that I think can be useful because they return some immutable information to identify one struct and they're only one line getters. I'm pretty sure one day will come when we'll need to ask "what's the key this verifier is using" in a `onet.Service` or something...

- `func (d *Dealer) Key() (abstract.Scalar, abstract.Point) `
- `func (v *Verifier) Key() (abstract.Scalar, abstract.Point) `
- `func (v *Verifier) Index() int `

This one is definitely useless:
- `func (a *aggregator) setDeal(d *Deal) `

share/vss/vss.go
nikkolasg: Ok after further thoughts and research, I guess we have nothing to loose or gain in terms of security by hashing first (if we use a hash function with the same number of bits security). However, it might be a little bit faster and I prefer the option of hash so I can have a `Hash` method on my struct that allows me to eliminate some redundant code in the dkg part. So let's do that.

nikkolasg:
Coverage `90.4%`

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+
+// ProcessDeal analyzes the Deal received from the Dealer.
+// If the Deal is valid, i.e. the verifier can verify its shares
+// against the public coefficients and the signature is valid, an approval
+// Response is returned and must be broadcasted to every participant
+// including the Dealer.
+// For the Deal itself to be invalid, it returns a complaint Response that must be
+// broadcasted to every other participant including the Dealer.
+// If the Deal has already been received, or the signature generation of the
+// Response failed, it returns an error without any Response.
```
Daeinar: > ProcessDeal first verifies the shares and signature included in the received deal and returns an approving (declining) response if the checks succeed (fail). The response must be broadcasted to all participants (including the dealer). If the deal was processed before or the signing of the response fails, only an error is returned.

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+	if err = v.aggregator.addResponse(r); err != nil {
+		return nil, err
+	}
+	return r, nil
+}
+
+// ProcessResponse analyzes the given response. If it's a valid complaint, the
+// verifier should expect to see a Justification from the Dealer. It returns an
+// error if it's not a valid response.
+// Call v.DealCertified() to check if the whole protocol is finished.
```
Daeinar: `v.DealCertified()`?

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+	if !v.EnoughApprovals() || !v.DealCertified() {
+		return nil
+	}
+	return v.deal
+}
+
+// ProcessJustification takes a DealerResponse and returns an error if
+// something went wrong during the verification. If it is the case, that
+// probably means the Dealer is acting maliciously. In order to be sure, call
+// v.EnoughApprovals() and if true, v.DealCertified()
```
Daeinar: - Ticks around functions? 
- Missing period at the end of the last sentence.

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+		if bytes.Equal(deal.SessionID, deals[0].SessionID) {
+			shares[i] = deal.SecShare
+		} else {
+			return nil, errors.New("vss: all deals need to have same session id")
+		}
+	}
+	return share.RecoverSecret(suite, shares, t, n)
+}
+
+// aggregator is used to collect all deals, responses and justification for one
```
Daeinar: justification**s**

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+	dealer    abstract.Point
+	verifiers []abstract.Point
+	commits   []abstract.Point
+
+	responses map[uint32]*Response
+	sid       []byte
+	deal      *Deal
+	t         int
+	badDealer bool
+}
```
Daeinar: Where are the justifications stored as mentioned in the main comment?

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+	fig := a.suite.Point().Base().Mul(nil, fi.V)
+	H := deriveH(a.suite, a.verifiers)
+	gih := a.suite.Point().Mul(H, gi.V)
+	ci := a.suite.Point().Add(fig, gih)
+
+	commitPoly := share.NewPubPoly(a.suite, nil, d.Commitments)
+
+	pubShare := commitPoly.Eval(fi.I)
+	if !ci.Equal(pubShare.V) {
+		return errors.New("vss: share do not verify against commitments in Deal")
```
Daeinar: > share does not verify ...

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+type Deal struct {
+	// SessionID is a unique session identifier for this protocol run
+	SessionID []byte
+	// SecShare is the private share generated by the dealer
+	SecShare *share.PriShare
+	// RndShare is the random share generated by the dealer
+	RndShare *share.PriShare
+	// T is the treshold used for this secret sharing run
+	T uint32
+	// Commitments are the coefficients used to verify the shares against.
```
nikkolasg: From what I look all sentences are finishing with a period and simple comment are without. Is that Ok ? I mean, please, let's try to not fall down into the rabbit holes of details in the comments like Linus Torvalds or stg, we all want this to move on... ^^
.

share/vss/vss.go
```diff
@@ -0,0 +1,628 @@
...
+
+	d.aggregator = newAggregator(d.suite, d.pub, d.verifiers, commitments, d.t, d.sessionID)
+	// C = F + G
+	d.deals = make([]*Deal, len(d.verifiers))
+	for i := range d.verifiers {
+		fi := f.Eval(i)
+		gi := g.Eval(i)
+		d.deals[i] = &Deal{
+			SessionID:   d.sessionID,
+			SecShare:    fi,
```
nikkolasg: Done.

share/vss/vss.go
nikkolasg: ?

share/vss/vss.go
ineiti: Take care with Linus T. references - he's my hero ;) So you just re-inforced @Daeinar's point...

nikkolasg: ... Please let us resist to the temptation of not liberate our inner Torvalds, otherwise nothing will gets ever merged. If you all do, I'll do too and it'll super annoying ^^

In any case, there are comments which form "sentence" and comments which are not sentences, ex: `list of deals this dealer has generated`. 
I don't see any reasons to put a point on non-sentences comments. Do you ?

share/vss/vss.go
```diff
@@ -334,10 +402,35 @@ func (v *Verifier) ProcessDeal(d *Deal) (*Response, error) {
...
+	ephBuff, err := e.DHKey.MarshalBinary()
+	if err != nil {
+		return nil, err
+	}
+	// verify signature
+	if err := sign.VerifySchnorr(v.suite, v.dealer, ephBuff, e.Signature); err != nil {
+		return nil, err
+	}
+
+	// compute shared key and AES128-GCM cipher
```
Daeinar: This says AES128-GCM versus AES256-GCM in line 197.

PR #126, opened 2017-03-09
corrected marshalling

### LINKED ISSUE #124: [EdDSA] Inconsistent (Un)Marshalling
Now:
 * `eddsa.Marshal` outputs the seed and the public key in a slice of bytes.
 * `eddsa.Unmarshal` reads the **secret key** and public key from a slice of bytes.

Wanted: 
 * `eddsa.Marshal` outputs the seed and the public key in a slice of bytes.
 * `eddsa.Unmarshal` reads the **seed** key and public key from a slice of bytes.

With a test.

christophebertrand:
closes #124

### DIFF
--- a/eddsa/eddsa.go
+++ b/eddsa/eddsa.go
@@ -24,7 +24,7 @@ type EdDSA struct {
 	Public abstract.Point
 }
 
-// NewEdDSAKey will return a freshly generated key pair to use for generating
+// NewEdDSA will return a freshly generated key pair to use for generating
 // EdDSA signatures.
 // If stream == nil, it will take the random.Stream.
 func NewEdDSA(stream cipher.Stream) *EdDSA {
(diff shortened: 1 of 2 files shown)


eddsa/eddsa.go
```diff
@@ -72,14 +72,20 @@ func (e *EdDSA) MarshalBinary() ([]byte, error) {
...
 	}
 
 	e.seed = buff[:32]
-	e.prefix = buff[32:64]
-	e.Secret = suite.Scalar().SetBytes(e.seed)
+
+	scalar := sha512.Sum512(e.seed)
+	scalar[0] &= 0xf8
+	scalar[31] &= 0x3f
+	scalar[31] |= 0x40
```
nikkolasg: It would be best if you could factor that part out in a method and use it here and also in `NewEdDSA` so we dont' repeat code un-necessarily.

PR #132, opened 2017-03-21
[VSS] Distributed Schnorr Signature (7/7)

nikkolasg:
This PR implements the Distributed Schnorr Signature protocol. It is a simple one, since it relies on DKG which is the foundation.


nikkolasg:
95.2% coverage

share/dss/dss_test.go
```diff
@@ -0,0 +1,224 @@
...
+	return sc, suite.Point().Mul(nil, sc)
+}
+
+func randomBytes(n int) []byte {
+	var buff = make([]byte, n)
+	rand.Read(buff[:])
+	return buff
+}
+
+func init() {
```
ineiti: style: please put `init` on top, just after `suite` and `nbParticipants`. Then the `Test*` and the private helpers at the bottom.

share/dss/dss.go
```diff
@@ -0,0 +1,214 @@
...
+	randomPoly   *share.PubPoly
+	msg          []byte
+	partials     []*share.PriShare
+	partialsIdx  map[int]bool
+	signed       bool
+}
+
+// PartialSig is partial representation of the final distributed signature. It
+// must be sent to each other participants.
+// XXX should we include a sessionID ? Can derive one from the two dist. key
```
nikkolasg: Ok. I've included a session id.

PR #133, opened 2017-03-23
Make Schnorr EdDSA Compatible

### LINKED ISSUE #131: [Sign] Schnorr should be EdDSA compatible for verification
There are three things that makes the current `sign.Schnorr` signature incompatible with the verification function of `eddsa.Verify`. 
+ The `sign.SchnorrSign` uses the hash function of the suite. While in principle it's correct, here that breaks compatibility with the `EdDSA` implementation which uses a `sha512` while the regular `ed25519` suite use `sha256`.
We should switch to `sha512` so our signatures are verifiable by the regular EdDSA verification implementations.

The signature format in `sign.Schnorr` is `c, s` with `c = H(Random || Public  || message)` and `s = r - c*x`. In `eddsa`, the signature is  `R || s`.
+ The schnorr format should follow eddsa format. No need to follow Wikipedia format, since nobody uses it.
+ The signature must be changed from `s = r - c * x` to `s = r + c * x`.

christophebertrand:
closes #131

### DIFF
--- a/sign/schnorr_test.go
+++ b/sign/schnorr_test.go
@@ -6,7 +6,6 @@ import (
 	"github.com/dedis/crypto/config"
 	"github.com/dedis/crypto/ed25519"
 	"github.com/dedis/crypto/eddsa"
-	"github.com/dedis/crypto/random"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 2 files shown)


ineiti:
Nice work. Some questions for @nikkolasg :

- is this now compatible with https://tools.ietf.org/html/rfc8032?
- will it break code that we have?
- should this go into crypto.v0? or only crypto.v1?

nikkolasg:
@ineiti :
* yes i believe as they kept backwards compatibility.
* Good question: if all of our code is running on `crypto.v0` then we should be fine if we don't merge it with v0. Otherwise, it will break for future signatures (like the dashboard in our office !)
* As mentioned above, I suggest we keep it in master and merge it in v1 when the times come to avoid any un-necessary problems.

sign/schnorr_test.go
```diff
@@ -27,19 +27,38 @@ func TestSchnorrSignature(t *testing.T) {
 	larger := append(s, []byte{0x01, 0x02}...)
 	assert.Error(t, VerifySchnorr(suite, kp.Public, msg, larger))
 
+	constantBytes := []byte{243, 45, 180, 140, 73, 23, 41, 212, 250, 87, 157, 243,
```
nikkolasg: where does this comes from .. ?

nikkolasg:
Almost ready to merge, well done !

sign/schnorr_test.go
christophebertrand: Those bytes are not a valid encoding of Scalar and Point

ineiti: Oh, I see. just move the `// wrong challenge` line before the `constantBytes`, then we'll understand it, too ;)

PR #138, opened 2017-03-29
More Flexibility for PVSS

Daeinar:
This patch adds some more flexibility to PVSS at the cost of a slight increase in usage complexity. 
Instead of taking the commitment polynomials as input and evaluating them, the functions take now the evaluated values as inputs pushing the responsibility to evaluate the polynomials to the users.


nikkolasg:
I'll take the time to review this one on Monday.

nikkolasg:
LGTM

Daeinar:
@ineiti Are you okay with this patch as well?

PR #146, opened 2017-05-05
adding schnorr signatures

nikkolasg:
Schnorr signatures were not present in v0. Since cothority uses v0, and that some people need the schnorr signature, we need to integrate them into v0.


ineiti:
I prefer the schnorr-signatures done by our hirees ;) anyway, will go away with crypto.v1, right?

PR #148, opened 2017-05-08
[v1] Suite decoupling proposition

nikkolasg:
This PR shows a possible solution discussed with @ineiti @Daeinar @bford to match the functionalities required by a package and the functionalities given to it via a Suite.
The approach is to remove the `Suite` definition and let each package define their own `Suite` with the exact set of methods as to what it needs.

If you want to get a feeling on how it behaves in practice, try to do it yourself from the `v1_structure` branch.

DISCLAIMER: I was appealed first by the idea when we talked ab out it. Then I started coding it and now I believe that this is not the right way to solve our problem. Even if it's only my first approach to the code, and there may be shorter ways to do it, it feels we're going "back" instead of moving forward with respect to the code's quality and conciseness . In short, this code misses a `Suite` interface ! :)

I list below  my reasons, some of them are repeated in the relevant part of the code directly.
Main problems:
- **Group structure**: Now that `crypto.Suite` is gone, then two choices are left to us:
  + Delete the `suiteEd25519` struct (“suite” is not defined anywhere) leaving us only with the `Curve` (which we could rename Group btw).  But how do we provide the `Hash` and `Cipher` (etc) functionalities then ? By attaching these methods to the Curve ? It does not make sense either, a curve should not provide any `Cipher` method whatsoever. 
  + Keep the `suiteEd25519` and keep the separation between the `Group` functionalities and the `Suite` functionalities. But then, why do we remove `Suite` at all, since it was a common interface that new “groups” could implement, that users could refer to. I see the following drawbacks:
     * An user will have to see if the `suiteEd25519` really implements the `Hash` method instead of just looking: “oh ed25519 has a Suite, I know all that it can do, neat!”
     * A dev has to lookup what methods his new `Curve` must implement to be interchangeable with others. This was previously done by just looking at the `abstract.Suite` interface. Now the dev have to lookup directly in the files, checking if there is a `Suite` definition somewhere in there, and implement those (**huge drawback**). And he needs to do that for all suites currently developed if he wants maximum interoperability .
- **Naming issue**: (related to previous issue) We still need to provide the `Cipher`,`Hash` methods at least. Where to put that ? It does not make such sense with `Group`.
- **Repeated code**: `pvss.go`, `anon/suite.go`, `test/test.go` all re-implement basically the same `Suite` as before. So when you actually code it, you have to look up the exact definition of the interface or, worse, function that you want to be implemented. It is mostly painful for the `Cipher` method which is long to type and remember (**huge drawback**). This is **one** example, but I can cite others of code duplication. In onet, we could provide a generic `Suite` that does `Hash` and `Cipher` so protocols / service don’t have to re-implement it, but then again… why removing `crypto.Suite` at all if it is to define it somewhere else ?


group/edwards25519/suite.go
```diff
@@ -1,24 +1,29 @@
...
+// NOTE: This structure is not removed so this brings the questions of naming and
+// consistency:
+// + what is a suite ? why would we call it a suite in this package and not something else
+// in curve25519 etc ?
+// + how do I know which method to implement ? For example. we'll want to add
+// ed448 to the our list, so we'll need to lookup here to see which methods
+// ed448 is supposed to implement to be sure one can ecchange edwards25519 and
+// edwards448 without any issues.
+// The Suite interface solves these issues by defining the common set of method
+// that be implemented (exactly as this `suiteed25519` suggests).
```
nikkolasg: NOTE

proof/proof.go
```diff
@@ -7,10 +7,25 @@ package proof
...
 )
 
+// NOTE This suite is the same as the global one. Writing it is not an easy when you
+// don't know what methods you want to have (i.e.  the same set of methods that
+// edwards25519 implements etc).
+// dleq.go uses this interface while it does not need it all of it, so we're back to same
+// problem as before. Solution is to move it out to an external package, and
+// that brings an additional amount of work for "giving out to a package only
+// what it needs".
+type Suite interface {
```
nikkolasg: NOTE

share/pvss/pvss.go
```diff
@@ -15,13 +15,28 @@ package pvss
...
 )
 
+// NOTE. Here, we need need the full suite definition for using the proof/
+// package while we only use Hash.
+// 1. the pkg has still access to lot more than what the package requires (so
+// the opposite of what we wanted to achieve)
+// 2. Again, full suite redeclaration
+// 3. Possibly troublesome inconsistencies: why would vss use a different suite
+// than share/poly.go, while vss USES share/poly.go as a building block.
+type Suite interface {
```
nikkolasg: NOTE

share/vss/vss.go
```diff
@@ -34,19 +34,29 @@ import (
...
 	"github.com/dedis/crypto/share"
 	sign "github.com/dedis/crypto/sign/schnorr"
+	"github.com/dedis/crypto/util/random"
 	"github.com/dedis/protobuf"
 )
 
+// NOTE: This suite is almost the same as the one defined in abstract/group
+// => again another place where code is repeated and adds a new abstraction to
+// take care of.
+type Suite interface {
```
nikkolasg: NOTE

sign/anon/suite.go
```diff
@@ -0,0 +1,15 @@
+package anon
+
+import (
+	"hash"
+
+	"github.com/dedis/crypto"
+)
+
+// NOTE: here again a full Suite definition...
+type Suite interface {
```
nikkolasg: NOTE

proof/proof.go
```diff
@@ -7,10 +7,25 @@ package proof
...
 import (
 	"errors"
+	"hash"
 
 	"github.com/dedis/crypto"
 )
 
+// NOTE This suite is the same as the global one. Writing it is not an easy when you
+// don't know what methods you want to have (i.e.  the same set of methods that
+// edwards25519 implements etc).
```
bford: Why isn't answering this question (mostly) as simple as copying the original global Suite as a template (as you did), but then temporarily commenting out all the members of this interface, try to compile it, see what breaks, and uncomment members as you confirm that they're real dependencies?

share/poly.go
```diff
@@ -14,10 +14,16 @@ import (
 	"crypto/subtle"
 	"encoding/binary"
 	"errors"
+	"hash"
 
 	"github.com/dedis/crypto"
 )
 
+type Suite interface {
```
bford: This looks like a good example of a more minimalist package-specific Suite interface working as expected...

share/pvss/pvss.go
```diff
@@ -15,13 +15,28 @@ package pvss
...
 
 	"github.com/dedis/crypto"
 	"github.com/dedis/crypto/proof"
-	"github.com/dedis/crypto/util/random"
 	"github.com/dedis/crypto/share"
+	"github.com/dedis/crypto/util/random"
 )
 
+// NOTE. Here, we need need the full suite definition for using the proof/
+// package while we only use Hash.
```
bford: Isn't this because you haven't yet narrowed down the set of members that proof.Suite actually needs to have?  Of course this Suite interface will be unnecessarily broad until/unless those of all the modules it depends on have been slimmed down appropriately as well.  I don't see a methodology problem here, only an (understandable) problem that the methodology hasn't been fully implemented yet.

proof/proof.go
nikkolasg: Well, the original Suite is supposed to not exist anymore ! I was lucky to still have it somewhere but if it does not exists, I have to find the right signature somewhere.
Also, this methodology of "put everything, then remove one method, see if it works, and iterate again" to me looks more like a debugging methodology than a proper coding methodology. Every dev./user of the library would potentially have to follow this methodology.

share/pvss/pvss.go
nikkolasg: Exactly, proof/ needs to be slimmed down more and packaged into sub-packages. Unfortunatly, this problem is likely to be occuring in many places in our code, rendering the approach quite expensive compared to the advantages in my opinion.

share/poly.go
nikkolasg: Indeed, when requiring only one additional member, it's relatively easy to code it.

examples/enc_test.go
```diff
@@ -6,26 +6,26 @@ import (
 	"github.com/dedis/crypto/util/random"
 )
 
-func ElGamalEncrypt(suite crypto.Suite, pubkey crypto.Point, message []byte) (
+func ElGamalEncrypt(group crypto.Group, pubkey crypto.Point, message []byte) (
```
ineiti: Shouldn't this be a `Suite` now? Even though if it only contains the `crypto.Group`, this would `ensuring that each package explicitly documents (via that package-specific Suite interface) what precise crypto functionality it depends on.`

examples/enc_test.go
```diff
@@ -54,18 +54,18 @@ see for example anon.Encrypt, which encrypts a message for
 one of several possible receivers forming an explicit anonymity set.
 */
 func Example_elGamalEncryption() {
-	suite := nist.NewAES128SHA256P256()
+	group := nist.NewAES128SHA256P256()
```
ineiti: That definitively is a `suite` and not a `group`.

examples/sig_test.go
```diff
@@ -11,68 +11,74 @@ import (
...
-	return suite.Scalar().Pick(c)
+	return group.Scalar().Pick(c)
 }
 
 // This simplified implementation of Schnorr Signatures is based on
 // crypto/anon/sig.go
 // The ring structure is removed and
 // The anonimity set is reduced to one public key = no anonimity
-func SchnorrSign(suite crypto.Suite, random cipher.Stream, message []byte,
+func SchnorrSign(group Suite, random cipher.Stream, message []byte,
```
ineiti: I'd call this `suite Suite`.

group/curve25519/suite.go
```diff
@@ -49,7 +49,7 @@ func (s *suiteEd25519) NewKey(rand cipher.Stream) crypto.Scalar {
 }
 
 // Ciphersuite based on AES-128, SHA-256, and the Ed25519 curve.
-func NewAES128SHA256Ed25519(fullGroup bool) crypto.Suite {
+func NewAES128SHA256Ed25519(fullGroup bool) *suiteEd25519 {
```
ineiti: Putting this together with Bryan's comment, I'd do:
1. make `suiteEd25519` public and group: `GroupEd25519`
2. add a `SuiteEd25519` with `sha256`

does that make sense?

group/edwards25519/suite.go
ineiti: I +1 your suggestion for `curveEd25519` or `groupEd25519` or whatever, but don't keep the `suite` in here.

proof/proof.go
ineiti: For the change of `crypto.v0` to `crypto.v1`, yes, this would be a bit awkward. But with the red-yellow-refactor cycle (http://blog.cleancoder.com/uncle-bob/2014/12/17/TheCyclesOfTDD.html), this is not too much of a problem. 

And changing the dependencies _does_ feel like debugging (I just got out of moving my 10'000 lines of ruby from 2.0 to 2.4... was not as bad as I thought, because I had some tests ;)

proof/proof.go
ineiti: As Bryan said: comment, compile, uncomment, compile...

ineiti: Or if you really want to go the hard way, take a piece of paper, go through every function, and write down what they need.

share/pvss/pvss.go
ineiti: Yes, it's a mapping of crypto-dependencies through the interfaces... And you get to see what exactly a certain package needs.

share/vss/vss.go
ineiti: Bryan's comment to this was that if you only depend on the `schnorr.Suite`, you can do

```
type Suite schnorr.Suite
```

and be done.

sign/anon/sig_test.go
```diff
@@ -258,20 +258,20 @@ var benchSig1Ed25519 = benchGenSigEd25519(1)
 var benchSig10Ed25519 = benchGenSigEd25519(10)
 var benchSig100Ed25519 = benchGenSigEd25519(100)
 
-func benchGenKeys(suite crypto.Suite,
+func benchGenKeys(g crypto.Group,
```
ineiti: Why is this `Group` and not a `Suite`?

sign/cosi/cosi.go
```diff
@@ -69,7 +69,7 @@ import (
 // CoSi struct. You can also give the full mask directly with SetMask().
 type CoSi struct {
 	// Suite used
-	suite crypto.Suite
+	group crypto.Group
```
ineiti: Why is it `Group` here?

nikkolasg: This answer holds as well as for the other places where I use a `Group`:
Using a `Group` is explicitely telling that the package is only using the `Point` & `Scalar` methods, and do not need a`Hash` or stg. You really want to add an additional level by writing `type Suite crypto.Group` ?

examples/enc_test.go
bford: Agreed.  In fact the Group interface is a fairly "suite-like" interface; we might discuss whether it should perhaps actually be renamed as such as well.  It's basically a suite interface for discrete-log groups.  (Note: I'm not suggesting at the moment we actually do this at the moment, but perhaps think about it.)

bford:
Regarding the "Group" vs "Suite" naming/consistency question, one exercise that might be useful for experimentation and education purposes (for now) would be to define a first-cut version of a "pairing-based group" Group/Suite interface.  We'll need something like this anyway for BLS signatures in DFINITY and all that.  Obviously it won't and shouldn't go into crypto v1, but the exercise of defining a first-draft of it and figuring out how the new pairing-based Group/Suite interacts with and relates to the traditional discrete-log Group/Suite interface from a software engineering perspective may help us think about what the naming and package organization conventions should be around this for v1.

nikkolasg:
Closing this one as the major work is being done directly in v1. See [epic](https://github.com/dedis/kyber/issues/151)

Issue #172, opened 2017-06-04
Bytes/SetBytes endianness control


bford:
mod.Int has a limited degree of support for "endianness control"; i.e., serving clients who might want little-endian byte-slices as opposed to the default big-endian.  However, two issues with this:

* First, it's currently completely specific to mod.Int, and thus this (I think generally useful) endian-neutrality feature is not available in the Bytes/SetBytes calls to the general Scalar interface.

* Second, I'm having second thoughts about whether the "stateful" approach to endianness control is the best idea here.  I think that approach was motivated mainly by the desire to keep the Bytes/SetBytes method signatures consistent with those in the standard big.Int package, which does not provide endianness control at all.  However, neither the Scalar nor Point interface is actually directly compatible with the big.Int package, anyway, so there doesn't seem to be a huge need to be dogged with that consistency.  Further, as I've been experiencing, that extra state sometimes leads to unpleasant surprises, especially when the same Scalar object gets used sometimes big-endian and sometimes little-endian, e.g., during transitions between different endian representations.  (This happens with ed25519 especially because ed25519 traditionally uses little-endian scalars.)

Thus, I propose that we move the ByteOrder, BigEndian, and LittleEndian names from the group/mod package to the top level, and define the Scalar.Bytes and Scalar.SetBytes to take a byteOrder parameter directly.  Just make this standard functionality for all Scalars (of which there currently aren't that many different 
implementations anyway).

Another, related bit of API inconsistency comes from the fact that mod.Int also has LittleEndian() and BigEndian() methods, which act like Bytes() for the appropriate endianness but also take 'min' and 'max' parameters defining the minimum and maximum output length.  I think this has proven to be pretty useful functionality as well - e.g., sometimes you just want the minimum-length slice that can represent an integer, sometimes you want a slice exactly some length.  So for further API consolidation, I suggest we eliminate the separate BigEndian() and LittleEndian() methods from mod.Int and instead make Scalar.Bytes() take three paramters: min, max, and byteOrder.

nikkolasg:
> I think this has proven to be pretty useful functionality as well

Do you have specific examples on that ? From my 2years experience here, I never used it explicitely. I'm always requiring a specific bitlength so min = 32 and max = 32.

I'm OK with the suggestion of moving the endianness as a parameters of the `SetBytes` and `Bytes` functions. I'm less ok with overloading them with a min and max parameter...

PR #173, opened 2017-06-04
Constant-time scalars in ed25519, and variable-time option flag

bford:
This PR implements fixes several issues that I think have been needing to get taken care of for a while:

* The biggest change, which motivates and drives the others, is in group/edwards25519, where I added real, constant-time scalar arithmetic for this curve, based on djb's code but adapted to support the generic Add/Sub/Mul operations we need rather than just the MulAdd and Reduce operations that come in the ed25519 signature scheme impl.  The Inv and Div operations are still non-constant time, unfortunately; that probably can and should be fixed too but will have to wait.

* Since constant-time operation often comes at a performance cost, and for example the Ed25519 point arithmetic code already includes both constant-time and non-constant-time versions of the main scalar multiply operation, I made those accessible by adding a SetVarTime() method to both the Point and Scalar interfaces.  The idea is that any implementation should default to constant-time implementations of all operations if constant-time operations are available; however, the caller can invoke SetVarTime(true) on particular Points and/or Scalars to indicate that operations on those particular Point/Scalar objects will never utilize secret information and hence should be safe to run in non-constant-time if that's faster.

* Added an endianness parameter to mod.Int.InitBytes(), because without it there's no way to specify the endianness the byte slice the caller is passing to this function.  (Since the stateful endianness of the object itself hasn't been initialized yet.)  But also see this more general related issue:  https://github.com/dedis/kyber/issues/172

* The Clone() operations in curv25519/proj.go and curve25519/ext.go were buggy, causing incorrect state-sharing in ways that apparently snuck past the test suites before (but for some reason got revealed while I was working on a completely different package)...  Fixed.

* The Add and Sub methods for projPoint in curve25519 similarly had state-overwrite bugs causing corruption when the output overlapped with one or both input objects; again this apparently slipped past the test suites before.

This code is NOT yet ready to commit; while the normal arithmetic tests pass, the EdDSA signing test using it doesn't yet succeed, so there's something still not quite right with it.  But I wanted to get review/discussion started on the API changes at least.  (This is for v1 of course.)


group/edwards25519/scalar.go
```diff
@@ -442,6 +618,1286 @@ func scMulAdd(s, a, b, c *[32]byte) {
...
+	c5 := 2097151 & (load3(c[13:]) >> 1)
+	c6 := 2097151 & (load4(c[15:]) >> 6)
+	c7 := 2097151 & (load3(c[18:]) >> 3)
+	c8 := 2097151 & load3(c[21:])
+	c9 := 2097151 & (load4(c[23:]) >> 5)
+	c10 := 2097151 & (load3(c[26:]) >> 2)
+	c11 := (load4(c[28:]) >> 7)
+	var carry [23]int64
+
+	s0 := 1916624 - c0 + a0
```
nikkolasg: Wow. I think I understand how you hacked the `ScAdd` but this... where does it come from ? ^^

nikkolasg:
Nice work. 
Regarding the `SetVarTime` method, I'm a bit hesitating since only one of our three suites provide somewhat constant time operations. Any caller thinking it should be safe to use this method with curve25519 will have no way to know that the concrete implementation does not support constant time operations.
Just one idea on top of my head...
```go
type Scalar interface {
  ...
  // this method can panic if constant time implementation is not available.
  // or it can return an error 
  ConstantTimeScalar() ConstantTimeScalar
type ConstantTimeScalar interface {
  Scalar
  // This should always be possible 
  VarTimeScalar() Scalar 
}
```

Another one would be to just panic in case the constant time implementation is not available. It may be simpler...

group/edwards25519/scalar.go
bford: Sorry, I know this is a bit obscure (like most of this code). :)  The constants 1916624, ..., down to s12 := 16, are the representation of the group order (L) times 16, which is the first "representation of 0 mod L" just higher than 2^256.  That way, when we subtract c from that, we ensure that the results remains on the positive side overall, which is needed since the modular reduction and carry-handling code below was only set up to handle positive results (but it has no problem handling positive results that are a bit - or a lot - too big).

bford:
Thanks Nicolas for the code review.  I agree that it's a potential source of confusion that only one of our curve implementations currently supports constant-time operation - but I think for now it's OK to leave it as a well-documented caveat stated loudly and clearly as part of the godoc for the non-constant-time curve implementation modules.  In the longer term, it would be great if we can fix that and either make the other implementations constant-time or replace them with better implementations of something else (e.g., Ed448, pairing-based curves, etc...) that do have constant-time implementations.  But that can be a project for later.

nikkolasg:
I agree that it would nice to have contstant time implementations for other groups as well. In the meantime, I still find it dangerous to not at least panic if the constant time implementation is not implemented yet. Most of the time, we/I/users? don't look at the comments of the concrete implementations so the user can think it's safe while it's not...

bford:
Can you clarify what solution you're proposing?  Should the non-constant-time implementations have a 'vartime' flag internally, which defaults to false but just causes panics if it's used that way, so a client who wants to use one of the variable-time implementations must explicitly SetVarTime(true) before using it?  That seems potentially reasonable, basically forcing the user to "check the box" that effectively says "I understand and acknowledge that I'm using non-constant-type crypto". :)

nikkolasg:
MMhhh right I was not thinking that the default is "constant time" implementations.
"Checking the box" would require a lot of boiler plate code so I'm not sure about that approach.
One more idea: what about specifying that in the `group.Scalar` method ? Either:
1)
```go
type Group interface {
   Scalar(varTime bool) Scalar
   Point(varTime bool) Point
}
```
Both methods can panic at creation time if the variable/constant time implementation is not supported, so there's absolutely no ambiguities from the caller's perspective. However, this approach has the drawback that it changes the API of the Group..
Or
2)
 ```go
type Group interface {
   VarTimeScalar() Scalar
   VarTimePoint() Point
   Scalar() Scalar
   VarTimePoint() Point
}
```
The last two methods (`Scalar` and `Point`) can panic for the groups not supporting yet constant time implementations. There is still no ambiguities. This approach's drawback is that it overloads the Group interface for "just a boolean" argument.

There's no perfect solutions in any case. I'm more in favor of 1) personnally. 
What do you think ? If you're not a fan of these ideas, then we should really make sure the comments are somewhere clearly visible, even in the readme or stg... I still find it a bit scary but I can live with it.

**Comment by bford** [comment]
The idea of defaulting to constant-time was for conservative safety: if you neglect to think about it (or don't understand), you'll get the default constant-time implementation and will at least be "more safe" even if potentially slower than necessary.  But I agree that defaulting to constant-time in implementations that don't have constant-time at all is a bit problematic. :)

I'm reluctant to add complexity to the Scalar/Point constructors in the Group interface, because the "ideal situation" I'd really like to get to is one where usually you never worry about or deal with the varTime flag at all, and instead you just almost always use the default constant-time implementations of everything (once everything has constant-time implementations :) ) - except in the rare performance-critical cases where you're performance-optimizing a particular algorithm based on our crypto library (such as signature verification or ZKP verification) and realize that no secrets are actually involved in that calculation, so you add the one or two SetVarTime(true) calls as appropriate to speed up the most performance-critical parts of those calculations.  I'd rather not have everybody have to think about constant- versus vartime all the time and pass the right parameters or call the right flavor of Point/Scalar constructor.  Better just to default to something safe and provide an uncommon-case override for when the client is really sure it's OK and performance-beneficial to use vartime operations.

File: group/edwards25519/scalar.go (hunk 1 above)
Comment (nikkolasg) [inline group/edwards25519/scalar.go]: I understand the second part of your answer but not sure I get the first part. I'll have to dig deeper in how theses computations are done and what are these representations (why s1 -> s23 ) etc. Thanks anyway!

Comment (Daeinar) [inline group/edwards25519/scalar.go]: I was facing a similar issue a while ago when I was implementing those scalar operations in Rust. To maintain readability and avoid hacky tricks, I decided to write wrappers around the `multiply_and_add(a,b,c) { return a*b + c }` method. Basically:
```
add(a,b) { return multiply_and_add(1,a,b) } // a+b
sub(a,b) { return multiply_and_add(-1,b,a) } // -b + a = a - b
neg(a) { return multiply_and_add(-1,a,0) } // -a
```
I derived the byte representation of the -1 scalar via the reference implementation (https://ed25519.cr.yp.to/python/ed25519.py) and it should be
```
            0xec, 0xd3, 0xf5, 0x5c, 0x1a, 0x63, 0x12, 0x58,
            0xd6, 0x9c, 0xf7, 0xa2, 0xde, 0xf9, 0xde, 0x14, 
            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 
            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 
```
This needs to be verified again. ;)

What do you think?

Comment (nikkolasg) [inline group/edwards25519/scalar.go]: I like this approach better: it may cost a few cycles more but at least I understand it, it's way more simpler to grasps ;)

File: group/edwards25519/point.go [hunk 2]
```diff
@@ -219,87 +217,30 @@ func (P *point) Neg(A kyber.Point) kyber.Point {
...
-	var a [32]byte
-	for i := range sb {
-		a[shi-i] = sb[i]
-	}
+	//sb := s.(*mod.Int).V.Bytes()
+	//shi := len(sb) - 1
+	//var a [32]byte
+	//for i := range sb {
+	//	a[shi-i] = sb[i]
+	//}
```
ineiti: If this is not used anymore, please delete it.

group/mod/int.go
```diff
@@ -108,7 +108,8 @@ func (i *Int) InitString(n, d string, base int, m *big.Int) *Int {
 
 // Return the Int's integer value in decimal string representation.
 func (i *Int) String() string {
-	return hex.EncodeToString(i.V.Bytes())
+	return i.V.String()
+	//return hex.EncodeToString(i.V.Bytes())
```
ineiti: remove line.

group/mod/int.go
```diff
@@ -473,3 +474,11 @@ func (i *Int) HideDecode(buf []byte) {
 	i.V.SetBytes(buf)
 	i.V.Mod(&i.V, i.M)
 }
+
+// Allow or disallow variable-time implementation.
+// This implementation unfortunately provices only variable-time operations,
```
ineiti: s/provices/provides/

ineiti:
Chiming in late to the discussion. If I understand correctly, we want to avoid that people use our library and think it is 'safe' but then leak information about their private keys. 
So far we have the following two suggestions:

1. Bryan's suggestion of adding a "make unsafe but fast"-method
` `+ have the cake and eat it, too
` `- users might think the group is safe when in fact it is not
2. Nicolas' suggestion of having const/var-methods for every group
` `+ clear what the user uses
` `- blows up the `Scalar`/`Point`-interfaces

I can think of two more:

3. Reflecting const/var-time in the name of the group (ed25519_var, ed25519_const)
` `+ clear what the user uses
` `- multiplication of groups
4. Bryan's suggestion, but per default panic if the group doesn't have a constant-time implementation, and only allow variable-timed-groups when a global `ALLOW_UNSAFE_VARIABLE_TIME` is set
` `+ safe by default
` `+ unsafe groups are available, but the user will have to set that *bad* global variable
` `- uses a global variable.

nikkolasg:
@ineiti Good ideas. I like 3. since it's clear and easy to grasp. We can make the default `NewAES128SHA256Ed25519()` return the constant time implementation, and a `NewAES128SHA256Ed25519VarTime()` which returns the variable time implementation. 
Moreover, it allows us to **gradually** allows constant time implementations for other suites. 
So by default users will use constant time implementation, without worrying about anything.

For 4., it's also a good idea but I have trouble to see where this global variable would be defined. Is it in the code ? or during the building phase (like C/C++ projects compilation option -DMY_VARIABLE ?

3. is good to me.

group/edwards25519/scalar.go
bford: Try benchmarking the two approaches.  I suspect that calling multiply_and_add with 1 or -1 will prove a lot slower, because the multiply_and_add will actually do a full multiplication and reduction, whereas in the hacky specialized functions the compiler's constant-propagation will eliminate a ton of that carry-reduction effort in the Add/Sub cases.

bford:
I don't like #3 so much because it effectively separates constant and variable-time implementations into completely different "cipher suites", which is semantically inappropriate and in its logical conclusion may make them extremely cumbersome to use.  

For example, if you have to instantiate different cipher-suites for constant-time or variable-time implementations of a curve, can you or can't you expect Scalars and Points created from one implementation to be accepted as inputs to operations on the other?  If they're different cipher suites, it would seem most natural to assume/require clients to expect they might be incompatible.  

Thus, to sign and verify signatures, the signing code will need to instantiate a constant-time ciphersuite and load the public key into a Point instance of that, whereas the verifying code will have to instantiate the other variable-time ciphersuite and load the (same!) public key into a (different) Point instance from that, just so that verification (a non-sensitive public process) can take advantage of the variable-time operations while keeping the signing process secure.

bford:
And you KNOW how I feel about global variables!!!

ineiti:
We _could_ do something like this:

https://husobee.github.io/golang/compile/time/variables/2015/12/03/compile-time-const.html

But it'll still be a variable somewhere in `kyber/group.go`. At least the variable could be private, and groups without constant-time implementations could call:

```
package kyber

var variableTimeAllowed;

def CheckVariableTimeAllowed(){
  if !variableTimeAllowed { 
    panic("This suite only knows variable timing, which might be unsafe.\n" +
              "If you're sure you want that, compile using\n" +
              "-ldflags \"-X kyber.variableTimeAllowed=true\"")
  }
}
```

But that might be another case of cuddling the users. Even though I'm all for cuddling ;)

As I see now the problem with my proposition for 3. (even though perhaps some nice use of `embedding structures` could make it possible to do casts on-the-fly, at least in one direction), I vote for 1. and 4.

group/edwards25519/scalar.go
ineiti: I wonder what optimisations golang does if this code would be rewritten in a loop...

Anyway, I'm OK to having magic constants in the code, as long as there is a link to a document that says how they've been chosen (like https://crypto.stackexchange.com/questions/10263/should-we-trust-the-nist-recommended-ecc-parameters)...

bford:
Actually, I think the Go standard approach to this would be just to make all the non-constant-time curve implementations (i.e., currently all but ed25519) conditionally compiled on a predicate such as '!vartime-curve-implementations'.  I think that would probably be a reasonable solution, and would effectively slim down the default compilation of the Kyber liberary even more.  I don't think we're currently using anything but ed25519 in "production" for onet etc anyway, are we?  (If we are, we probably shouldn't be!)

ineiti:
OK, missed those flags - I never used them. But yes, this also sounds very reasonable to me. And yes, for the moment we only use ed25519 in onet/cothority. That's the advantage of having a global variable (just kidding, it'll go away in v2)!

nikkolasg:
I've added constant time implementation for the inversion (so Div is now also in constant time). 
I'll try to play with the compilation flags and make travis happy in the next days.

bford:
Good, thanks @nikkolasg for adding constant-time inversion.

group.go
```diff
@@ -53,11 +53,18 @@ type Scalar interface {
...
 
-	// Bytes returns the raw internal representation
+	// Bytes returns a big-Endian representation of the scalar
 	Bytes() []byte
+
+	// Allow or disallow use of faster variable-time implementations
+	// of operations on this Point, returning the old flag value.
+	// This flag always defaults to false (constant-time only)
+	// in implementations that can provide constant-time operations.
+	SetVarTime(varTime bool) bool
```
nikkolasg: I just realized that we are *not* providing the variable time implementation anymore for `Scalar`s in `edwards25519/` anymore.  The simplest way to do it would be to add `if s.varTime { .. } else { ... }` in every method of the `Scalar`... Having this "dual state" / dual implementation in one struct does not look very nice at all and has its pile of drawbacks (maintability, readability mainly). May I propose an alternative approach ;) ?
What about if we change the signature to 
```go
type Scalar interface {
  ...
  // VarTime returns a Scalar using variable time arithmetic operations if varTime is true. 
  // It returns a constant time implementation of a Scalar if varTime is false. 
  VarTime(varTime bool) Scalar // plus the old value if needed ?
}
```
so in our implementation we can have a clear distinction between the two implementations
```go
type scalarVarTime mod.Int
type scalarConstTime struct  {
   v [32]byte
}
```
I see at least two main advantages here:
1. API-wise: 
  - not keeping track on "which state are we on" for a specific scalar.  
  - consistency with other methods that returns the result as Scalar 
2. Implementation-wise: it's much cleaner and easier to code separate implementation, one vartime and one const. time. If I were a new developer hacking on this library, I'd find that much more elegant than having `if / else ` in every methods.

bford: I thought of this approach, and I'm open to it but still on the fence about it.  The main downsides I see are:

* It makes constant-time-Scalars and variable-time-Scalars (for the same Suite) different and incompatible by default, in that you can't for example add a variable-time Scalar to a constant-time Scalar and such without doing awkward conversions.  The need for constant-time operations is often typically specific to particular values - i.e., whether that particular value involves secret information or not - and a lot of calculations in practice involve a combination of secret and non-secret values.  Potentially mitigating this downside somewhat, it might be the case that most of the non-secret values we work with tend to be Points and not Scalars - but if so, that would suggest just making Scalar arithmetic constant-time-only and not bothering to provide faster variable-time versions at all.

* The other functions in the same Suite, particular point.Mul, need to know how to detect and deal with two different types of scalar types: i.e., point.Mul will have to do a type-switch test whether the scalar that was passed as the parameter is actually a scalarVarTime or a scalarConstTime.  Not a big deal and perfectly doable, but a slight cost in complexity and perhaps time (type tests tend to be more expensive than just checking a bool flag, though probably not significant compared to the total cost of a scalar multiply).

bford: Perhaps it would be worth taking a bit of a "census" throughout the crypto and/or onet code to see where all we do Scalar arithmetic and how much of that arithmetic might conceivably benefit from being able to use a faster variable-time version for non-sensitive calculations?  If it turns out there's not all that much, and/or it proves hard to identify and disentangle which are the "sensitive" vs "non-sensitive" scalars, then I would propose in this case it's probably best (certainly safest) just to make the scalars always use constant-time arithmetic and take whatever performance hit that means.

Given that point.Mul is the main performance-critical operation we're concerned about in general, perhaps keeping that as the only one we bother with having constant-time and variable-time versions of would be perfectly fine.

nikkolasg: I see, it's a valid counter argument indeed. 

I am totally in favor of dropping variable time implementation for `Scalar`, following in spirit the well-coded scalar implementation in [dalek](https://github.com/isislovecruft/curve25519-dalek/blob/master/src/scalar.rs) for example.
From my experience with this library and the algorithms I coded, I don't think it should be *that* hard to distinguish between sensitive or non sensitive scalars if one really takes the time at each individual operations. Nevertheless, my main concern is that screw-ups *can* and *will* happen (Murphy's law!). Removing vartime implementation will at least prevent a large range of potential weaknesses in existant and future implementations: that's a **huge** benefit.

bford:
Wait, that's not what I had in mind (removing the SetVarTime method from the Scalar interface entirely).  The decision we discussed was for the Ed25519 implementation of scalars for now, not for all scalars forever. I think it's still useful to have the interface method, (a) for consistency with Point, (b) in case we decide later that the performance benefit of vartime scalar arithmetic is worth the trouble, (c) in case it's more worthwhile other Scalar implementations in the future.  The method implementation can just do nothing and return false.

nikkolasg:
Re-added the `SetVarTime` method (although, as you must have guessed now, I'm not a fan of that method ;) ). 
I fixed a bug introduced in the `mod.Int` package related to the `String()` method - compatibility with `CoSi` - `EdDSA` was broken.
I also fixed our examples: they now use edwards25519. SInce `curve25519` and `nist/` do not compile by default (you have to provide them the `-tags vartime` argument), it's much easier.
I think this PR is LGTM, I'll let Linus decide for himself tomorrow. @Daeinar don't hesitate to chim in if you have some time !

group/curve25519/proj.go
```diff
@@ -238,6 +248,11 @@ func (P *projPoint) Mul(s kyber.Scalar, G kyber.Point) kyber.Point {
 	return P
 }
 
+// This implementation only supports variable-time operations
+func (P *projPoint) SetVarTime(varTime bool) bool {
+	return true
+}
```
ineiti: Should `SetVarTime` throw `fatal` if it cannot honour a `false`?

group/edwards25519/const.go
```diff
@@ -6,21 +6,31 @@ package edwards25519
...
-	"gopkg.in/dedis/kyber.v1/group/mod"
 )
 
 // prime modulus of underlying field = 2^255 - 19
 var prime, _ = new(big.Int).SetString("57896044618658097711785492504343953926634992332820282019728792003956564819949", 10)
 
 // prime order of base point = 2^252 + 27742317777372353535851937790883648493
-var primeOrder, _ = new(mod.Int).SetString("7237005577332262213973186563042994240857116359379907606001950938285454250989", "", 10)
+// XXX this should probably just be a big.Int
+var primeOrder, _ = new(big.Int).SetString("7237005577332262213973186563042994240857116359379907606001950938285454250989", 10)
```
ineiti: Is there a better way to instantiate it?

group/edwards25519/scalar.go
```diff
@@ -442,6 +627,1286 @@ func scMulAdd(s, a, b, c *[32]byte) {
...
+	s[25] = byte(s9 >> 11)
+	s[26] = byte((s9 >> 19) | (s10 << 2))
+	s[27] = byte(s10 >> 6)
+	s[28] = byte((s10 >> 14) | (s11 << 7))
+	s[29] = byte(s11 >> 1)
+	s[30] = byte(s11 >> 9)
+	s[31] = byte(s11 >> 17)
+}
+
+// Hacky scMul cobbled together rather sub-optimally from scMulAdd.
```
ineiti: For this and the other functions - please add a reference to a document where this is explained and where it is cobbled from.

group/nist/doc.go
```diff
@@ -1,3 +1,8 @@
 // Package nist implements cryptographic groups and ciphersuites
 // based on the NIST standards, using Go's built-in crypto library.
+// Since that package does not implements constant time arithmetic operations
```
ineiti: s/implements/implement/

group/nist/doc.go
```diff
@@ -1,3 +1,8 @@
...
+// yet, it must be compiled with the "vartime" compilation flag.
+
```
ineiti: Should we add that comment to the other groups, too?

proof/hash_test.go
```diff
@@ -137,44 +134,31 @@ func ExampleHashProve_2() {
...
+	// 000000e0  ee f3 b2 f4 06 e4 98 a7  24 2f 51 b8 13 b4 b5 69  |........$/Q....i|
+	// 000000f0  94 ad 33 b9 c4 e3 95 8b  7f 18 6d 1e f1 07 3e 0d  |..3.......m...>.|
+	// 00000100  c3 86 a1 24 1b 3e e1 59  d5 bd 70 a1 ff f9 7c 07  |...$.>.Y..p...|.|
+	// 00000110  8c 9c 52 f7 47 34 46 c9  1a 05 4b 68 57 49 c7 0e  |..R.G4F...KhWI..|
+	// 00000120  31 68 8d ca 3f 6a 85 a1  0d f1 cf 9d 21 05 83 f2  |1h..?j......!...|
+	// 00000130  35 63 b0 65 a8 50 a5 ee  ec 95 f8 fd 78 de 73 08  |5c.e.P......x.s.|
+	// 00000140  35 e7 59 fa 2b 41 20 f8  b6 48 43 62 91 f1 c6 99  |5.Y.+A ..HCb....|
+	// 00000150  0e 64 9c 2c 06 fe 84 75  4f ca 03 7f 28 b5 6d 0c  |.d.,...uO...(.m.|
+	// 00000160  1d 63 e5 73 26 c4 9f 61  62 5b 5c 34 70 66 d0 e4  |.c.s&..ab[\4pf..|
+	// 00000170  ec ca b8 ee 9a 50 07 6b  0c 75 5a a6 77 b3 20 0f  |.....P.k.uZ.w. .|
```
ineiti: Probable stupid question: why is the result all of a sudden shorter?

proof/hash_test.go
```diff
@@ -46,13 +46,10 @@ func ExampleHashProve_1() {
...
-	// 00000010  90 fc 7c db a0 c6 e8 12  f2 0a d4 40 a4 b6 c4 de  |..|........@....|
-	// 00000020  9e e8 61 88 5e 50 fd 03  a9 ff 9c a3 c4 29 f7 18  |..a.^P.......)..|
-	// 00000030  49 ad 31 0e f9 17 15 1e  3b 8d 0e 2f b2 c4 28 32  |I.1.....;../..(2|
-	// 00000040  4a 5c 64 ca 04 eb 33 db  a9 75 9b 01 6b 12 01 ae  |J\d...3..u..k...|
-	// 00000050  4e de 7c 6b 53 85 f8 a5  76 ba eb 7e 2e 61 2c a5  |N.|kS...v..~.a,.|
-	// 00000060  e8                                                |.|
+	// 00000000  27 fd 13 c3 6e e6 df a5  00 aa 0c 93 a7 b8 21 4b  |'...n.........!K|
+	// 00000010  a5 cf 26 c2 a0 99 68 b0  a0 36 9d 7a de 92 95 7a  |..&...h..6.z...z|
+	// 00000020  1f 46 d8 67 4a 71 49 c9  7c d2 8f 2b 75 8c cc 83  |.F.gJqI.|..+u...|
+	// 00000030  b4 31 0c 6f 6c 2e 75 70  cd 8b 8e 04 b0 54 4f 07  |.1.ol.up.....TO.|
```
ineiti: Why is the result all of a sudden shorter?

ineiti:
Yes, starts to look good. Small comments about comments and some questions.

group/curve25519/proj.go
nikkolasg: I think so yes. I was suggesting the idea of panic before but no clear consensus have emerged. I'm up for it.

group/edwards25519/const.go
nikkolasg: What is wrong with this way ? What is "better" ?

group/edwards25519/scalar.go
nikkolasg: Ì think @bford cobbled it together from the document... in its head .. !

proof/hash_test.go
nikkolasg: I guess it has to do with the small size of Ed25519 scalars/points/signature. If you want to dig deeper, you can check the size of P256 points / scalars.

group/edwards25519/const.go
ineiti: With reference to the comment `this should probably just be a big.Int` I thought that there is another way to instantiate it. So I didn't understand the comment. Perhaps just take the old comment, which looks clearer to me?

group/edwards25519/scalar.go
ineiti: He's a genius, but I don't think he can do 255-bit maths in his head. So what is the source?

group/edwards25519/const.go
bford: That comment is probably obsolete and I (or someone) just forgot to delete it.  primeOrder used to be defined as a nist.Int rather than a big.Int; looks like that's already fixed.

group/edwards25519/scalar.go
bford: You mean what is the source for this code in general?  It's from the ScMulAdd implementation that Adam Langley ported from Dan Bernstein's original hand-coded implementation of Ed25519 signing.

I think Adam Langley's original Go port was at: https://github.com/agl/ed25519

And DJB's original ed25519 signing scheme web page is at: https://ed25519.cr.yp.to/index.html

But I'm not sure if this is the "source" you're asking for.

nikkolasg: And also these sources are already referenced in the top level documentations. I think Linus wants a document describing how to perform these constant time operations *from* the ed25519 implementations, showing how and why some of the magics constants you have introduced (the modular reduction for substration etc etc). I haven't heard of any but who knows.. !

bford: I didn't introduce any new modular reduction code; that was all copied verbatim from the original ScMulAdd.  Hence there are now several copies of that code, one in each of the new functions (ScAdd, ScSub, ScMul).  Of course the code duplication is non-ideal, but as I mentioned before it enables the compiler to do important automatic optimizations of each through constant propagation.

Really the only new "thing" I added anywhere was the new constant (the representation of 16*L) used in ScSub to make sure that the (existing) modular reduction and carry logic only needs to deal with positive numbers.

ineiti: OK, I understand. It's ed25519 instead of p256. So all is OK.

group/edwards25519/scalar.go
nikkolasg: To double check, I made a quick version of constant time scalar playing with the `scMulAdd` parameters to implement `Mul` `Add` and `Sub`. It turns out indeed that @bford 's implementation is two times faster ;)
```
BenchmarkNIKKOCTAdd-4           10000000               200 ns/op
BenchmarkNIKKOAdd-4             20000000               100 ns/op
BenchmarkNIKKOCTMul-4           10000000               202 ns/op
BenchmarkNIKKOMul-4             10000000               167 ns/op
BenchmarkNIKKOCTSub-4           10000000               193 ns/op
BenchmarkNIKKOSub-4             20000000                99.5 ns/op
PASS
```
where `...CTXXX` is my quick implementation. The others are @bford 's implementation.

nikkolasg:
"Lasts" (?) corrections  done in https://github.com/dedis/kyber/pull/173/commits/e72ba4d4e92f7e498fe11fe88c730ba9ec67e22e

proof/hash_test.go
ineiti: OK, I understand. It's ed25519 instead of p256. So all is OK.

group/edwards25519/scalar.go
ineiti: I did a short search on the web and the only reference how to create those constants is in the `supercop ref10` implementation:

https://github.com/anonimal/supercop-ed25519-ref10

there are several python-scripts in there that create those constants. But I couldn't find any description how they do those optimizations.

Can we add a reference to the supercop-copy at github? And if somebody knows why the python-scripts work and has a link, please add that, too.

proof/hash_test.go
ineiti: ed25519 instead of p256.

group/curve25519/ext.go
```diff
@@ -267,6 +269,11 @@ func (P *extPoint) Mul(s kyber.Scalar, G kyber.Point) kyber.Point {
 	return P
 }
 
+// This implementation only supports variable-time operations
+func (P *extPoint) SetVarTime(varTime bool) bool {
+	return true
```
ineiti: should probably also `fatal` if `varTime` is `false`.

group/edwards25519/curve.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// Curve represents the Ed25519 group.
+// There are no parameters and no initialization is required
+// because it supports only this one specific curve.
+type Curve struct {
+
+	// Set to true to use the full group of order 8Q,
+	// or false to use the prime-order subgroup of order Q.
+	//	FullGroup bool
```
ineiti: Is this still available and should be re-enabled?

group/edwards25519/curve.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// Returns 32, the size in bytes of an encoded Scalar for the Ed25519 curve.
+func (c *Curve) ScalarLen() int {
+	return 32
+}
+
+// Create a new Scalar for the prime-order subgroup of the Ed25519 curve.
+func (c *Curve) Scalar() kyber.Scalar {
+	//i := mod.NewInt64(0, primeOrder)
```
ineiti: unused code - remove or re-enable?

group/edwards25519/curve.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Returns 32, the size in bytes of an encoded Point on the Ed25519 curve.
+func (c *Curve) PointLen() int {
+	return 32
+}
+
+// Create a new Point on the Ed25519 curve.
+func (c *Curve) Point() kyber.Point {
+	P := new(point)
+	//P.c = c
```
ineiti: Unused code

group/edwards25519/scalar.go
```diff
@@ -4,11 +4,196 @@
...
+	"gopkg.in/dedis/kyber.v1"
+	"gopkg.in/dedis/kyber.v1/group/mod"
+	"gopkg.in/dedis/kyber.v1/util/bytes"
+	"gopkg.in/dedis/kyber.v1/util/marshalling"
+	"gopkg.in/dedis/kyber.v1/util/random"
+	"gopkg.in/dedis/kyber.v1/util/subtle"
+)
+
 // This code is a port of the public domain, "ref10" implementation of ed25519
 // from SUPERCOP.
```
ineiti: Add reference to supercop:

https://bench.cr.yp.to/supercop.html
and perhaps:
https://github.com/anonimal/supercop-ed25519-ref10

group/edwards25519/scalar.go
```diff
@@ -4,11 +4,196 @@
...
+	return subtle.ConstantTimeCompare(v1, v2) != 0
+}
+
+// Set equal to another Scalar a
+func (s *scalar) Set(a kyber.Scalar) kyber.Scalar {
+	s.v = a.(*scalar).v
+	return s
+}
+
+func (s *scalar) Clone() kyber.Scalar {
```
ineiti: Needs comment

group/edwards25519/scalar.go
```diff
@@ -4,11 +4,196 @@
...
+}
+
+func (s *scalar) setInt(i *mod.Int) kyber.Scalar {
+	b := i.LittleEndian(32, 32)
+	copy(s.v[:], b)
+	return s
+}
+
+// Set to a small integer value
+func (s *scalar) SetInt64(v int64) kyber.Scalar {
```
ineiti: golint error

group/edwards25519/scalar.go
```diff
@@ -4,11 +4,196 @@
...
+}
+
+// Set to a fresh random or pseudo-random scalar
+func (s *scalar) Pick(rand cipher.Stream) kyber.Scalar {
+	i := mod.NewInt(random.Int(primeOrder, rand), primeOrder)
+	return s.setInt(i)
+}
+
+func (s *scalar) SetBytes(b []byte) kyber.Scalar {
+	// XXX handle simple and scReduce cases appropriately
```
ineiti: is this a todo?

group/edwards25519/scalar.go
```diff
@@ -4,11 +4,196 @@
...
+	return s.setInt(mod.NewIntBytes(b, primeOrder, mod.LittleEndian))
+}
+
+func (s *scalar) SetVarTime(varTime bool) bool {
+	return false
```
ineiti: so this is only a constant-time implementation? Should it also `fatal` when called with `true`? Or is it OK to accept `true` because that's not a security-problem?

group/edwards25519/scalar.go
```diff
@@ -4,11 +4,196 @@
...
+}
+
+// Encoded length of this object in bytes.
+func (s *scalar) MarshalSize() int {
+	return 32
+}
+
+func (s *scalar) MarshalBinary() ([]byte, error) {
+	return s.toInt().MarshalBinary()
+	//buf := s.v
```
ineiti: Unused code

group/mod/int.go
```diff
@@ -108,6 +108,7 @@ func (i *Int) InitString(n, d string, base int, m *big.Int) *Int {
 
 // Return the Int's integer value in decimal string representation.
 func (i *Int) String() string {
+	//return i.V.String()
```
ineiti: unused code

ineiti:
Things to do:
- [ ] pass `golint`
- [ ] remove or re-enable dead code
- [ ] decide whether `SetVarTime` should fatal if called with `false` in non-constant groups

group/edwards25519/scalar.go
```diff
@@ -442,6 +627,1286 @@ func scMulAdd(s, a, b, c *[32]byte) {
...
+	s15 := int64(0)
+	s16 := int64(0)
+	s17 := int64(0)
+	s18 := int64(0)
+	s19 := int64(0)
+	s20 := int64(0)
+	s21 := int64(0)
+	s22 := int64(0)
+	s23 := int64(0)
+
```
Daeinar: To avoid code duplication, should we modularize the code below into "carry-center/uncenter" and "reduce" functions similarly to https://github.com/isislovecruft/curve25519-dalek/blob/master/src/scalar.rs#L607 ff and then use them in `scAdd()`, `scSub()`, etc.?

group/edwards25519/scalar.go
bford: I think it should just always silently return false, as per the current code.  This way:

1. a client who wants to test whether a given Point or Scalar implementation supports constant time at all can see if s.SetVarTime(false).SetVarTime(false) returns false (it's constant-time) or true (it only has a variable-time implementation).

2. a client who wants to test whether a constant-time Point or Scalar implementation supports any variable-time optimizations can test the result of s.SetVarTime(true).SetVarTime(true): if false, it's always constant-time; if true, it is either always or configurably variable-time.

group/edwards25519/scalar.go
bford: You could try doing that, but be sure to benchmark before/after, as I expect the obvious/easy ways of doing that will disable or make it more difficult for the compiler to optimize it down using constant propagation etc.  That's the main reason I didn't try to reduce code duplication in that way.

group/edwards25519/scalar.go
ineiti: That sounds method-overloading to me. I'd prefer to have a `GetVarTime` method for that.

Another proposition would be to have a `(bool, error)`-return, so the user is correctly informed if his desired action has been performed or not.

group/edwards25519/scalar.go
ineiti: +1 for this idea. I like having the more public methods on top of the file (`scAdd`, `scSub`) and the helpers (`center/uncenter`, `reduce`) at the bottom.

group/edwards25519/scalar.go
nikkolasg: `scReduce` takes only 64 bytes as input, so I'm not sure we should use it when given more than 64 bytes in `SetBytes`. Otherwise, we need to cut down the given slice of bytes and that may yield wrong results ?. i would keep it the way it is now. What do you think ?

group/edwards25519/scalar.go
nikkolasg: I would prefer avoiding adding again a new method to `Scalar` if possible. The interface starts to get really long.
Why again do we need to return the old "varTime" value ? Is it needed in practice ? I would have just imagine some part of the code that needs vartime to just call `scalar.SetVarTime(true)`, why should they care about the old value ?
If we make that method return an error as @ineiti suggested, it becomes `if err := scalar.SetVarTime(true); err != nil { ... }`. What do you think ?

ineiti: Sounds nice to me - what does @bford think?

group/edwards25519/scalar.go
nikkolasg: Golint and gofmt do not complain. it implements the kyber.Scalar interface so it should be fine.

ineiti: because it's a private structure...

group/edwards25519/scalar.go
ineiti: even 64 bytes might be too big for ed25519-scalars. I did not understand the difference between `simple` and `scReduce` cases.

group/edwards25519/scalar.go
ineiti: even if golint doesn't complain - can we still follow go's convention for commenting methods? Please? Pretty Please?

group/edwards25519/scalar.go
nikkolasg: Waiting approval by @bford before implementing or not ? (I already reverted many times some change so I'd prefer to wait until then :D )

cipher/stream.go
```diff
@@ -80,10 +80,11 @@ func (sc *streamCipher) Partial(dst, src, key []byte) {
 	// absorb cryptographic input (which may overlap with dst)
 	if key != nil {
 		nkey := ints.Min(n, len(key)) // # key bytes available
-		sc.h.Write(key[:nkey])
+		// XXX check return error ?
```
ineiti: in fact an error-check here would be very nice. But as the main method does not return any error, I'm not sure what is best to do here.

So either remove the `_, _ = ` or do some error-checking.

examples/sig_test.go
```diff
@@ -54,7 +54,7 @@ func SchnorrSign(suite Suite, random cipher.Stream, message []byte,
 	// And check that hashElgamal for T and the message == c
 	buf := bytes.Buffer{}
 	sig := basicSig{c, r}
-	suite.Write(&buf, &sig)
+	_ = suite.Write(&buf, &sig)
```
ineiti: again error-checking: either remove `_ = ` or do some error-checking.

examples/sig_test.go
```diff
@@ -87,9 +87,9 @@ func SchnorrVerify(suite Suite, message []byte, publicKey kyber.Point,
 }
 
 // Example of using Schnorr
-func ExampleSchnorr() {
+func Example_schnorr() {
```
ineiti: Please follow https://golang.org/pkg/testing/#pkg-examples

nikkolasg: Golint do not pass with that so I choose golint.

examples/sig_test.go
nikkolasg: Golint.

cipher/stream.go
nikkolasg: Golint

group/edwards25519/elligator.go
```diff
@@ -13,7 +13,7 @@ import (
 // equal the result from PublicKeyToCurve25519.
 func PrivateKeyToCurve25519(curve25519Private *[32]byte, privateKey *[64]byte) {
 	h := sha512.New()
-	h.Write(privateKey[:32])
+	_, _ = h.Write(privateKey[:32])
```
ineiti: again error-handling, like above

group/edwards25519/scalar.go
nikkolasg: @bford Indeed you were completely right !
```
BenchmarkCTScalarAdd-4                  20000000               104 ns/op
BenchmarkCTScalarSimpleAdd-4            10000000               196 ns/op
BenchmarkCTScalarFactoredAdd-4           2000000               592 ns/op
BenchmarkCTScalarMul-4                  10000000               166 ns/op
BenchmarkCTScalarSimpleMul-4            10000000               200 ns/op
BenchmarkCTScalarFactoredMul-4           2000000               722 ns/op
BenchmarkCTScalarSub-4                  20000000                98.5 ns/op
BenchmarkCTScalarSimpleSub-4            10000000               192 ns/op
BenchmarkCTScalarFactoredSub-4           2000000               610 ns/op
```
where `ScalarFactoredXXX` is is using a factored out version of `scMul` etc.
Crazy thought the big difference in time. I'd love to see this kind of benchmark on Rust as well... And also to take some compiler courses to really grasp the thing > < :)

bford: Yes, taking a compiler course would be worthwhile - you'll immediately understand the reasons behind performance effects like this. :)

ineiti: No simplification then...

cipher/stream.go
ineiti: but the `XXX` can go away.

ineiti:
Caught Bryan and asked him for `func SetVarTime(varTime bool) error`. He was not against it, so +1.

nikkolasg:
Piouf ! That might be the end of that PR ? :D

PR #174, opened 2017-06-09
Added licenses to v0

ineiti:
This PR adds the LGPL-license to the root-directory and the golang-license to the "ed25519"-directory.


nikkolasg:
This is why we did not see the changes. It was on v0 ;)

PR #176, opened 2017-06-13
Added licenses to master

ineiti:
Changed the license to MPLv2

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,375 @@
+This code is (c) by DEDIS/EPFL 2017 under the MPL v2 or later version.
+
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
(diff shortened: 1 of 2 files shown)


Daeinar:
Looks good.

LICENSE
```diff
@@ -0,0 +1,375 @@
+This code is (c) by DEDIS/EPFL 2017 under the MPL v2 or later version.
```
nikkolasg: later versions ? are we sure ?

nikkolasg:
Just a small question otherwise it's LGTM

LICENSE
ineiti: That's commonly used, e.g. the GPL is mostly used as "or any later version" - supposing mozilla will not screw up.

PR #186, opened 2017-07-10
CoSi Policy Mechanism 

Daeinar:
This PR implements a policy mechanism similar to the one in https://github.com/bford/golang-x-crypto/blob/master/ed25519/cosi/verify.go and includes some clarifications on the documentation. It also tweaks the participating mask mechanism to use "1" ("0") to denote an enabled (disabled) node as it seems to be more natural. If there's a particular reason of why it should be the other way round (as in the current reference code), I'm always open to suggestions / discussions.

Done:
- [x] Policy mechanism
- [ ] Documentation
- [x] Mark enabled (=true) nodes as "1" in the participating mask


ineiti:
IIRC, the 0 == enabled is due to the default in go of putting variables to 0. So by default, a participant is enabled. If you put 1 == enabled, you need to make sure that the bitmask is correctly initialized.

cosi/cosi.go
```diff
@@ -221,80 +227,97 @@ func (c *CoSi) VerifyResponses(aggregatedPublic abstract.Point) error {
...
 
 	return nil
 }
 
-// VerifySignature is the method to call to verify a signature issued by a Cosi
-// struct. Publics is the WHOLE list of publics keys, the mask at the end of the
-// signature will take care of removing the indivual public keys that did not
-// participate
+// Verify checks the given cosignature using the provided signing policy.
+func Verify(suite abstract.Suite, publics []abstract.Point, policy Policy, message, sig []byte) error {
```
nikkolasg: I would prefer to give the policy as the last argument. The reason for that is to be able to quickly and easily verify any signature without necessarily holding a `Policy` object. Having the "nil" argument in the middle is cumbersome because you need to remember exactly at which position is the policy one.
Pushing the idea further, if no policy is given, maybe use a default policy requiring at least one or two signers.

In short, movin from 
```go
// definition
func Verify(suite abstract.Suite, publics []abstract.Point, policy Policy, message, sig []byte) error 
// calling
cosi.Verify(suite,publics,nil,msg,sig)
```
to
```go
// definition
func Verify(suite abstract.Suite, publics []abstract.Point, message, sig []byte, policy Policy) error 
// calling
cosi.Verify(suite,publics,msg,sig,nil)
```
 What do you think about it ?

cosi/cosi.go
```diff
@@ -328,123 +351,162 @@ func (c *CoSi) genResponse() error {
...
+
+// Policy represents a fully customizable cosigning policy deciding what
+// cosigner sets are and aren't sufficient for a collective signature to be
+// considered acceptable to a verifier. The Check method may inspect the set of
+// participants that cosigned by invoking cosi.Mask and/or cosi.MaskBit, and may
+// use any other relevant contextual information (e.g., how security-critical
+// the operation relying on the collective signature is) in determining whether
+// the collective signature was produced by an acceptable set of cosigners.
+type Policy interface {
+	Check(c *CoSi) bool
```
nikkolasg: A policy on a cosi signature needs:
+ the total list of participants
+ the bitmask indicating which participants were involved during the signature process

We should try to let the Policy only access what it needs in order to perform its duty.
In order to do this, I'd just set `mask` to `Mask` and change the signature of the policy interface to:

```go
type Poiicy interface {
    Check(participants []abstract.Point, m *Mask)
}
```
I'm even wondering whether the list of participants is required and should not be given "out of band" anyway, but for the simplest threshold case, let's keep it simple. 

....

Or does a `Policy` need something else ?

cosi/cosi.go
Daeinar: Sounds good, I'll do that.

cosi/cosi.go
Daeinar: Good point. The participants (i.e., their public keys) are currently stored inside the `mask`, so passing the `mask` alone should be sufficient.

Daeinar:
This PR is deprecated due to #188.

PR #188, opened 2017-07-12
CoSi Reimplementation

Daeinar:
To clean up the CoSi implementation, reduce it to the cryptographic operations only, and simplify the API.


cosi/cosi_new.go
```diff
@@ -0,0 +1,130 @@
...
+
+func Challenge(suite abstract.Suite, commitment abstract.Point, pubkey abstract.Point, mask, message []byte) abstract.Scalar {
+	// return H(commitment || pubkey || mask || message)
+}
+
+func Response(suite abstract.Suite, random abstract.Scalar, challenge abstract.Scalar, prikey abstract.Scalar) {
+	// return random - challenge * prikey
+}
+
+func AggregateResponses(suite abstract.Suite, responses []abstract.Scalar) abstract.Scalar {
```
nikkolasg: should it not be directly integrated with the `Sign` method ?

cosi/cosi_new.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/dedis/kyber/abstract"
+)
+
+func Commit(suite abstract.Suite, s cipher.Stream) (abstract.Scalar, abstract.Point) {
+	// compute random scalar
+	// compute commitment
+	// return both
+}
+
+func AggregateCommitments(suite abstract.Suite, commitments []abstract.Point, mask *Mask) (abstract.Point, *Mask) {
```
nikkolasg: What do you do with the mask ? Do you assume that the length of the given commitments is the same as the total size of the participant set, with some nil entries for absent participants ?
Otherwise, I'm wondering if the `mask` is really belonging here. 
Another idea would be to make that function return the mask. Use a more structure-d approach like `Commiment struct { Index int, C abstract.Point}` so we know the index of each commitment. Then one just need to call 
`aggCommit, mask  = AggregateCommitments(suite, commitments, n)` with n being the length of the set of participants.
Just ideas... ;)

cosi/cosi_new.go
Daeinar: As far as I know we want to be able to aggregate responses from our children and then pass up the result to our parent in the tree communication model. So we cannot integrate response aggregation directly?

cosi/cosi_new.go
Daeinar: Same here, we want to aggregate the commitments along the way in the commit bottom-up phase of the CoSi communication tree. For that we need to be able to update the mask continuously until we reach the root node. As a consequence we also need to pass the mask from bottom to top I think.

Daeinar: After discussion with @nikkolasg we came to the conclusion that we need to provide a slice of `*Mask` as input instead of just a single one as each commitment comes with its own mask.

cosi/cosi_test.go
```diff
@@ -1,228 +1,92 @@
...
-		buff, err := p.MarshalBinary()
-		assert.Nil(t, err)
-		Ed25519Publics = append(Ed25519Publics, xEd25519.PublicKey(buff))
+	// Init masks
+	var masks []*Mask
+	var byteMasks [][]byte
+	for i := 0; i < n; i++ {
+		masks = append(masks, NewMask(testSuite, publics))
+		masks[i].SetMaskBit(i, true) // TODO: this has to go into the mask initialization somehow
+		byteMasks = append(byteMasks, masks[i].mask)
```
nikkolasg: Mhhh this whole bitmask thing is making me still uneasy ... 
As you state in your TODO, for the moment, the user still have to:
1. Create the mask
2. Set its i bit to true
3. Take its []byte version so it can be used in `AggregateCommitments`

These three steps will **always** be present (you don't create a mask if you don't participante), so it could be good to find a way to make that more transparent, more automatic. Unfortunatly, I don't have a clear idea yet how to do this, I'll think about it...

Daeinar:
@bford @nikkolasg @ineiti: I've finished a first (somewhat) polished version of the CoSi reimplementation. I will add some more tests later on. Any thoughts on this?

Daeinar:
*bump* /cc @bford @nikkolasg @ineiti

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
+func (m *Mask) SetBit(signer int, enable bool) error {
+	if signer > len(m.publics) {
+		return errors.New("index out of range")
 	}
-	if c.random == nil {
-		return errors.New("No random scalar computed in this cosi")
+	byt := signer >> 3
+	msk := byte(1) << uint(signer&7)
+	if ((m.mask[byt] & msk) == 0) && enable {
+		m.mask[byt] ^= msk // flip bit in mask from 0 to 1
```
ineiti: methinks a `|=` for setting and a `&^=` for clearing would be more understandable. Or even define a `BitClear` and a `BitSet` that does the `AggregatePublic.Sub` and `AggregatePublic.Add`, too, as it is used both here and in `SetMask`.

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
-// mask holds the mask utilities
-type mask struct {
-	mask      []byte
-	publics   []abstract.Point
-	aggPublic abstract.Point
-	suite     abstract.Suite
+// Bit returns a boolean value indicating whether the given signer is
+// enabled (true) or disabled (false).
+func (m *Mask) Bit(signer int) (bool, error) {
+	if signer > len(m.publics) {
```
ineiti: If I'm not mistaken, use `>=` here.

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
-	// Start with an all-disabled participation mask, then set it correctly
-	cm := &mask{
-		publics: publics,
-		suite:   suite,
+// CountEnabled returns the number of enabled nodes in the CoSi participation
+// mask, i.e., it returns the hamming weight of the mask.
+func (m *Mask) CountEnabled() int {
+	hw := 0
+	for i := range m.publics {
+		byt := i >> 3
```
ineiti: Use `Bit` here.

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
-		return errors.New("No private key given in this cosi")
+// SetBit enables (enable: true) or disables (enable: false) the bit
+// in the participation mask of the given cosigner.
+func (m *Mask) SetBit(signer int, enable bool) error {
+	if signer > len(m.publics) {
+		return errors.New("index out of range")
 	}
-	if c.random == nil {
-		return errors.New("No random scalar computed in this cosi")
+	byt := signer >> 3
```
ineiti: Why not use `Bit` here?

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
-	return c.response
+// SetMask sets the participation bitmask according to the given byte slice
+// interpreted in little-endian order, i.e., bits 0-7 of byte 0 correspond to
+// cosigners 0-7, bits 0-7 of byte 1 correspond to cosigners 8-15, etc.
+func (m *Mask) SetMask(mask []byte) error {
+	if m.Len() != len(mask) {
+		return fmt.Errorf("mismatching mask lengths")
+	}
+	for i := range m.publics {
+		byt := i >> 3
```
ineiti: Why not use `Bit` here?

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
-	for i := range clone {
-		clone[i] = 0xff
-	}
-	copy(clone[:], cm.mask)
-	return clone
+// ThresholdPolicy allows to specify a simple t-of-n policy requring that at
+// least the given threshold number of participants have cosigned to make a
+// collective signature valid.
+type ThresholdPolicy struct {
+	t int
```
ineiti: Can you instantiate a `ThresholdPolicy` from outside of `package cosi` with `t` being private? I don't think so.

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
-
-	// add our own commitment to the aggregate commitment
-	c.aggregateCommitment = c.suite.Point().Add(c.suite.Point().Null(), c.commitment)
-	// take the children commitments
-	for _, com := range subComms {
-		c.aggregateCommitment.Add(c.aggregateCommitment, com)
+// AggregateCommitments returns the sum of the given commitments and the
+// bitwise OR of the corresponding masks.
+func AggregateCommitments(suite abstract.Suite, commitments []abstract.Point, masks [][]byte) (abstract.Point, []byte, error) {
+	if len(commitments) != len(masks) {
```
Daeinar: A single mask (for one commitment of type `abstract.Point`) has the type `[]byte`. The number of bytes in a mask is `ceil(#nodes/8)`. So for 8 nodes you get 8 commitments and 8 masks each of length 1 byte. For 9 nodes you get 9 commitments and 9 masks each of length 2 bytes.

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
+		return nil, nil, errors.New("mismatching lengths of commitment and mask slices")
 	}
-	return c.aggregateCommitment
-
+	aggCom := suite.Point().Null()
+	aggMask := make([]byte, len(masks[0]))
```
Daeinar: The length of the masks is checked in the function `AggregateMasks()` which will throw an error if there is a mismatch.

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
 }
 
-// CreateChallenge creates the challenge out of the message it has been given.
-// This is typically called by Root.
-func (c *CoSi) CreateChallenge(msg []byte) (abstract.Scalar, error) {
-	// H( Commit || AggPublic || M)
+// Challenge creates the collective challenge from the given aggregate
+// commitment V, aggregate public key A, mask Z, and message M, i.e., it
+// returns c = H(V || A || Z || M).
+func Challenge(suite abstract.Suite, commitment abstract.Point, mask *Mask, message []byte) (abstract.Scalar, error) {
```
Daeinar: Because I need to hash the aggregate public key `A` which is stored in `Mask`. For the aggregation of masks in `AggregateCommitment` it is not (yet) relevant what the (intermediate) aggregate public key is since the leader sends the final aggregated mask down to all nodes so that they can compute the final aggregate public key for the challenge via `SetMask()` on the `Mask` struct that every node has initialized at the beginning of the protocol. 

Alternatively, we could (1) compute the aggregate public key outside of `Challenge` using the aggregate mask or (2) inside of `Challenge` which would require to give the list of all public keys and the final aggregated mask to `Challenge`.

Don't know which of all those variants is better. Thoughts?

cosi/cosi.go
```diff
@@ -1,450 +1,399 @@
...
+		byt := i >> 3
+		msk := byte(1) << uint(i&7)
+		if ((m.mask[byt] & msk) == 0) && ((mask[byt] & msk) != 0) {
+			m.mask[byt] ^= msk // flip bit in mask from 0 to 1
+			m.AggregatePublic.Add(m.AggregatePublic, m.publics[i])
+		}
+		if ((m.mask[byt] & msk) != 0) && ((mask[byt] & msk) == 0) {
+			m.mask[byt] ^= msk // flip bit in mask from 1 to 0
+			m.AggregatePublic.Sub(m.AggregatePublic, m.publics[i])
+		}
```
Daeinar: Yeah tested it with several values with the largest being 1024 nodes.

cosi/cosi.go
Daeinar: What do you mean? Rename `byt` (or `i`) to `Bit`?

Daeinar: Because I need the values of `byt` and `msk` inside the `if` statements as well and `Bit` currently only returns `(bool, error)`. Not sure if it would be good to overload `Bit` to return `byt` and `msk` as well. Actually, I just realised that `Bit` is never used in the code anyways, so maybe we can just remove that? ;)

cosi/cosi.go
Daeinar: See above.

cosi/cosi.go
Daeinar: Hmm but then I would need to return `(int, error)` in `CountEnabled()` which is not really necessary here I think.

cosi/cosi.go
```diff
@@ -1,450 +1,388 @@
...
+	}
+	return aggCom, aggMask, nil
 }
 
-// CreateChallenge creates the challenge out of the message it has been given.
-// This is typically called by Root.
-func (c *CoSi) CreateChallenge(msg []byte) (abstract.Scalar, error) {
-	// H( Commit || AggPublic || M)
+// Challenge creates the collective challenge from the given aggregate
+// commitment V, aggregate public key A, mask Z, and message M, i.e., it
```
nikkolasg: Sorry for being picky but there has been no clear consensus on the fact that we should marshal the bitmask inside the challenge. I know you seem to really want it but if we do it, then it's not possible to make participants leave at subsequent steps of the protocol, which is a point of discussion for the draft. I can't recall exactly the reason why the bitmask was necessary here also... do you ?

cosi/cosi.go
```diff
@@ -1,450 +1,388 @@
...
 
-// CreateResponse is called by a leaf to create its own response from the
-// challenge + commitment + private key. It returns the response to send up to
-// the tree.
-func (c *CoSi) CreateResponse() (abstract.Scalar, error) {
-	err := c.genResponse()
-	return c.response, err
+// Response creates the response from the given random scalar v, (collective)
+// challenge c, and private key a, i.e., it returns r = v + c*a.
+func Response(suite abstract.Suite, random abstract.Scalar, challenge abstract.Scalar, private abstract.Scalar) (abstract.Scalar, error) {
```
nikkolasg: `challenge, private abstract.Scalar`

nikkolasg: Even better if the private key comes first in my opinion but I don't mind :)

cosi/cosi.go
nikkolasg: Referring to my above comment, if we don't hash the challenge, we don't need the Mask at all, only the aggregated point.

cosi/cosi.go
ineiti: thanks for the clarification.

cosi/cosi.go
ineiti: but `AggregateMasks()` is not called in this function, is it? So I'd do the check again, anyway.

Daeinar: It is, see https://github.com/dedis/kyber/blob/3f73f33656c1c5e2ebca141ef09cfba321edb551/cosi/cosi.go#L76.

cosi/cosi.go
Daeinar: Yes you are right that the discussion around that is still not concluded. I just went for the conservative way.

Nevertheless, I think one can retain compatibility to EdDSA by just concatenating the bitmask with the message and pass that as the message to the `Challenge` function. This approach would also allow us to use the regular approach to compute the challenge and still commit to the bitmask if desired but would of course also mean that all nodes have to know whether they should hash the bitmask or not.

Daeinar:
The kyber.v1 version supersedes this PR, see https://github.com/dedis/kyber/blob/v1/sign/cosi/cosi.go.

Daeinar:
Closing, outdated.

Issue #193, opened 2017-08-08
[dfinity] Move code from v1 into master


nikkolasg:
Majors development for the dfinity project have been made on top of the v1 branch (future stable candidate) in the v1_dfinity branch.
This should be moved to the master branch:
- [ ] [pbc group](https://github.com/dedis/kyber/pull/195)
- [ ] [bls signatures](https://github.com/dedis/kyber/pull/196)
- [ ] [updated share/ package](https://github.com/dedis/kyber/pull/197)
- [ ] [pedersen vss & dkg](https://github.com/dedis/kyber/pull/198)

nikkolasg:
Finally the whole work will be easier done in a separate repo : [paper_17_dfinity](https://github.com/dedis/paper_17_dfinity)

PR #195, opened 2017-08-08
[dfinity - 1/4] Move group/point/scalar code 

nikkolasg:
This brings the dfinity pbc implementation into master. This is the first checkbox in the issue #193 .


ineiti:
I'm not sure this should go into master/v0. See comment on slack/engineer about how we could go on.

PR #200, opened 2017-08-09
[vss] changing status to boolean and remove justified status

nikkolasg:
Quick fix so that the messages used by share/vss are protobuf-able.
I'll do the porting on v0 and v1 afterwards.

### DIFF
--- a/share/vss/vss.go
+++ b/share/vss/vss.go
@@ -100,22 +100,11 @@ type Response struct {
 	SessionID []byte
 	// Index of the verifier issuing this Response
 	Index uint32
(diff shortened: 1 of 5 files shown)


share/vss/vss.go
```diff
@@ -100,21 +100,19 @@ type Response struct {
 	SessionID []byte
 	// Index of the verifier issuing this Response
 	Index uint32
-	// 0 = NO APPROVAL == Complaint , 1 = APPROVAL
-	Status byte
+	// Complaint == false, APPROVAL == true
+	Status bool
```
ineiti: If you call this `Approved`, then you don't even need the constants, no?

PR #211, opened 2017-08-14
checking if aggregator is already initialized

ineiti:
correctly verify if `aggregate` is already initialized.


ineiti:
Done in #213

Issue #212, opened 2017-08-14
crypto.v0/share/dkg - threshold question


ineiti:
Why can't I run a `crypto.v0/share/dkg`-sharing with a threshold equal to the number of nodes? It never returns true for `Finished()`.

PR #214, opened 2017-08-14
[dkg] adding index method to dkg

nikkolasg:
Closes https://github.com/dedis/kyber/issues/212

### DIFF
--- a/share/dkg/dkg.go
+++ b/share/dkg/dkg.go
@@ -200,6 +200,11 @@ func NewDistKeyGenerator(suite abstract.Suite, longterm abstract.Scalar, partici
 	}, nil
 }
 
+// Index returns the index of this generator amongst the list of participants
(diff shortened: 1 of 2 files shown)


ineiti:
Why using an accessor here? I always think it's a bad security model to hide access to fields by implementing a getter-method. And it also smells like Java. Anyway, https://golang.org/doc/effective_go.html#Getters says it's OK to do so. But we should have a more unified way of doing it throughout the project - some parts use getters, others export the fields. Perhaps we have to wait for 3 engineers to get a consensus ;)

nikkolasg:
> I always think it's a bad security model to hide access to fields by implementing a getter-method. 

Why do you think so ?
I actually prefer to use getters for structs doing complex operations so 1) no way to mess it up 2) it can be used as an interface the day you want to change the underlying complex operations.  For packets-like structs, I prefer simple exported fields.

PR #215, opened 2017-08-15
exporting index of distkeygenerator

ineiti:
it's useful to know which index we're at for a `DistKeyGenerator`, e.g. when creating shares that should be Lagrange interpolated.


ineiti:
done

PR #222, opened 2017-08-29
first try at making experimental pass

### LINKED ISSUE #221: Pass master-build
For the moment the build in master doesn't pass because of the `experimental`-branch. I started to add a `experimantal/abstract` to write a small _wrapper_ for the old `v0`-interfaces, but gave up when being confronted with `Hiding` and `Cipher`...

The other, more simple solution, is to ignore `experimental` when testing with travis.

ineiti:
Closes #221

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,6 @@ test_verbose:
 	go test -v -race -short ./...
 
 test_goverall:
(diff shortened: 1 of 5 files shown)


ineiti:
Wrong approach

PR #223, opened 2017-09-08
Porting CoSi to kyber.v1

Daeinar:
See https://github.com/dedis/kyber/issues/207.
Supersedes: https://github.com/dedis/kyber/pull/188.

PR #224, opened 2017-09-09
Porting New CoSi to Master

Daeinar:
After a discussion with @ineiti, the CoSi reimplementation should go directly to master.

Superseds https://github.com/dedis/kyber/pull/223.


ineiti:
The `Edit`-button at the upper right of the PR can change the destination-directory!

This fails for the moment as the `kyber/master` still needs to be cleaned up a bit to serve as development. https://github.com/dedis/kyber/pull/225 should take care of that.

Daeinar:
@ineiti: Added a CoSi suite using SHA512. Was this what you had intended? I am not sure that this is actually fitting here. The specification of such suites should be done ideally in a central place like a top level `suite` folder, no?

sign/cosi/cosi.go
```diff
@@ -1,452 +1,382 @@
...
-		return errors.New("Signature invalid")
+	x, err := left.MarshalBinary()
+	if err != nil {
+		return err
+	}
+	y, err := V.MarshalBinary()
+	if err != nil {
+		return err
+	}
+	if subtle.ConstantTimeCompare(x, y) == 0 || !policy.Check(mask) {
```
nikkolasg: why subtle here ? I don't see any benefits because it does not deal with any secret inputs, and costs you 2 `MarshalBinary` and a slow compare.

sign/cosi/cosi.go
```diff
@@ -1,452 +1,382 @@
...
-	c.genCommit(s)
-
-	// add our own commitment to the aggregate commitment
-	c.aggregateCommitment = c.group.Point().Add(c.group.Point().Null(), c.commitment)
-	// take the children commitments
-	for _, com := range subComms {
-		c.aggregateCommitment.Add(c.aggregateCommitment, com)
+// AggregateCommitments returns the sum of the given commitments and the
+// bitwise OR of the corresponding masks.
+func AggregateCommitments(suite Suite, commitments []kyber.Point, masks [][]byte) (kyber.Point, []byte, error) {
```
nikkolasg: So how do I get a `Mask` from this slice of byte ? It would be nice to see the flow from here to the signing which requires a  `*Mask`

Daeinar: You can look at the test to see how it's achieved. After you have (received/computed) the aggregated mask `aggMask` you use `SetMask(aggMask)` to configure the `Mask` struct correctly. We assume that every node has already initialized a `Mask` struct in the beginning of the CoSi protocol run by default.

nikkolasg:
LGTM

Daeinar:
@ineiti If you are fine as well, feel free to merge.

PR #226, opened 2017-10-20
Helper functions to check whether individual keys or indices are set in a CoSi mask or not

Daeinar:
As per request by Raphael.

### DIFF
--- a/sign/cosi/cosi.go
+++ b/sign/cosi/cosi.go
@@ -307,6 +307,27 @@ func (m *Mask) SetBit(i int, enable bool) error {
 	return nil
 }
 
+// IndexEnabled checks whether the given index is enabled in the mask or not.
+func (m *Mask) IndexEnabled(i int) (bool, error) {
+	if i >= len(m.publics) {
+		return false, errors.New("index out of range")
+	}
+	byt := i >> 3
+	msk := byte(1) << uint(i&7)
+	return ((m.mask[byt] & msk) != 0), nil
+}
+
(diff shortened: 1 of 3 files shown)


ineiti:
Failed lint-test ;)

sign/cosi/cosi_test.go
```diff
@@ -187,3 +188,61 @@ func TestCoSiThreshold(t *testing.T) {
 		}
 	}
 }
+
+func TestMask(t *testing.T) {
+	n := 5
```
ineiti: Test for `n := 10`, so that it tests the wrapping over a byte!

Daeinar:
The failing lint test is `shuffle/simple.go:174:2: redundant if ...; err != nil check, just return error instead.`. I've never even touched that file. -.-

sign/cosi/cosi.go
```diff
@@ -307,6 +307,27 @@ func (m *Mask) SetBit(i int, enable bool) error {
 	return nil
 }
 
+// IndexEnabled checks whether the given index is enabled in the mask or not.
+func (m *Mask) IndexEnabled(i int) (bool, error) {
+	if i >= len(m.publics) {
+		return false, errors.New("index out of range")
+	}
+	byt := i >> 3
+	msk := byte(1) << uint(i&7)
```
ineiti: Sorry to chime in so late: what about:

```
byt := i / 8
msk := byte(1 << ( i % 8 ))
```

It's the same, but I like the division and modulo 8, it tells me better what's going on than right-shifting by 3 and and with 7...

PR #232, opened 2017-11-07
Port fixes on DKG from v1

nikkolasg:
Since the shenanigan starts to fade away, we still have some artefacts that were not completely pushed on master. This makes my drand code work.

### DIFF
--- a/share/pedersen/dkg/dkg.go
+++ b/share/pedersen/dkg/dkg.go
@@ -5,7 +5,6 @@ package dkg
 import (
 	"crypto/cipher"
 	"errors"
-	"fmt"
 
 	"github.com/dedis/kyber"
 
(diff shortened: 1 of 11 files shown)


share/pedersen/dkg/dkg.go
```diff
@@ -248,7 +246,7 @@ func (d *DistKeyGenerator) ProcessJustification(j *Justification) error {
 // vss.Verifier.DealCertified()). If the distribution is certified, the protocol
 // can continue using d.SecretCommits().
 func (d *DistKeyGenerator) Certified() bool {
-	return len(d.QUAL()) >= d.t
+	return len(d.QUAL()) >= len(d.participants)
```
nikkolasg: This is needed currently to have every participants during the DKG time. I have a student working on the next iteration to be able to tolerate absent nodes. If you want to know the reason why, I'll be happy to explain but IRL ;)

share/pedersen/vss/vss.go
```diff
@@ -402,11 +402,20 @@ func (v *Verifier) decryptDeal(e *EncryptedDeal) (*Deal, error) {
 	return deal, err
 }
 
+// ErrNoDealBeforeResponse is an error happening if the verifiers receives a
```
jeffallen: English fixes and removing the reference to DKG, which is not really relevant here:

ErrNoDealBeforeResponse is an error returned if a verifier receives a
deal before having received any responses. For the moment, the caller must
be sure to have dispatched a deal before.

PR #233, opened 2017-11-08
Put special case key selection into the group that needs it

### LINKED ISSUE #205: NewKey
From @nikkolasg 
I recall now that `NewKey` was previously in `Suite` and so I naturally shifted it into `Group` since `Suite` does not exist anymore (sigh).
if we *really want* it out of the Group interface, we thought with @ineiti of different solutions:
1. move the bit twiddling thing in the `NewKeyPair` method, where a manual switch to check if the `Suite` is ed25519 and in that case, apply the bit twiddling which would be implemented in the `key` package (where the `KeyPair` is defined).
2. move out the functionality of creating a "key" into a `kyber.KeyFactory` functionality. That way, `NewKeyPair` can check if the group given implements `KeyFactory`, if yes, then call `group.NewKey` otherwise simply create a regular `Scalar`. The actual bit twiddling would stay in the `ed25519` package.

jeffallen:
Closes #205

### DIFF
--- a/group.go
+++ b/group.go
@@ -167,6 +167,4 @@ type Group interface {
 	Point() Point  // Create new point
 
 	PrimeOrder() bool // Returns true if group is prime-order
-
-	NewKey(cipher.Stream) Scalar
 }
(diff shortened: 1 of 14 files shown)


ineiti:
LGTM

PR #234, opened 2017-11-08
Assume a curve's order is prime unless the curve offers a way to check

### LINKED ISSUE #203: PrimeOrder
@bford wrote
The Group.PrimeOrder() method still needs to be either eliminated or renamed to something less confusing.  (It sounds like it's asking whether the group is prime-order, but what it returns is actually whether the order of the Base() point is prime-order.)  What all still depends on this?  More generally, we need to systematize the API for obtaining basic properties of the group in question - e.g., its order, whether there's a cofactor, etc.  Perhaps with several methods, perhaps with just one method that returns a GroupProperties struct of some kind, not sure.

From @nikkolasg to @bford 
1. I would advise for a `GroupProperty` struct  but given that we only have a few numbers of parameters to put (order of the group and cofactor ?), it's also OK for me to add those functions to the `Group` interface directly. What do you think ? 
2. Regarding the amgiguity you mentionned, I'm not sure about the best approach. Maybe a `PrimeCurveOrder` and `PrimeGroupOrder` ? In case of ed25519, the former returns false and the latter true.

jeffallen:
Closes #203

### DIFF
--- a/experimental/openssl/curve.go
+++ b/experimental/openssl/curve.go
@@ -256,10 +256,6 @@ func (c *curve) String() string {
 	return c.name
 }
 
-func (c *curve) PrimeOrder() bool {
-	return true // we only support the NIST prime-order curves
-}
-
 func (c *curve) ScalarLen() int {
 	return c.nlen
 }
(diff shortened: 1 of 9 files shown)


ineiti:
Now it has conflicts, due to the merge of the previous PR.

PR #236, opened 2017-11-10
Removing experimental code, which is no longer useful.

jeffallen:


ineiti:
Needs resolving of conflicts

PR #243, opened 2017-11-15
[VSS] Pedersen: Remove complaint and add Timeout

nikkolasg:
PR from @CedricCook to adapt the DKG code to allow for absent node during DKG setup.


share/vss/pedersen/vss.go
```diff
@@ -537,6 +554,21 @@ func (a *aggregator) VerifyDeal(d *Deal, inclusion bool) error {
 	return nil
 }
 
+// cleanVerifiers checks the aggregator's response array and creates a StatusComplaint 
+// response for all verifiers who have no response in the array.
+func (a *aggregator) cleanVerifiers() error {
```
nikkolasg: This function does not need to return an error I guess.

share/vss/pedersen/vss_test.go
```diff
@@ -80,428 +81,505 @@ func TestVSSWhole(t *testing.T) {
...
+//     require.Equal(t, StatusApproval, resp.Status)
+//     require.Nil(t, err)
+
+//     aggr := ver.aggregator
+
+//     for i := 1; i < aggr.t-1; i++ {
+//         aggr.responses[uint32(i)] = &Response{Status: StatusApproval}
+//     }
+//     // not enough approvals
+//     assert.Nil(t, ver.Deal())
```
nikkolasg: Try to call `setTimeout` before this assert so it timeouts everyone else but the t-1 approvals and it should in theory make the test pass.

share/vss/pedersen/vss_test.go
```diff
@@ -80,428 +81,505 @@ func TestVSSWhole(t *testing.T) {
...
+//     assert.Nil(t, dealer.SecretCommit())
+
+//     aggr.responses[uint32(aggr.t)] = &Response{Status: StatusApproval}
+//     assert.True(t, aggr.EnoughApprovals())
+
+//     for i := aggr.t + 1; i < nbVerifiers; i++ {
+//         aggr.responses[uint32(i)] = &Response{Status: StatusApproval}
+//     }
+//     assert.True(t, aggr.EnoughApprovals())
+//     assert.Equal(t, suite.Point().Mul(secret, nil), dealer.SecretCommit())
```
nikkolasg: Mhhhh this should still work. This is strange. Any ideas?

share/vss/pedersen/vss_test.go
```diff
@@ -80,428 +81,505 @@ func TestVSSWhole(t *testing.T) {
...
+// 
+// Won't work, since all aggr.responses[(i > t)] are not set. Timeout check prevents this
+// func TestVSSAggregatorDealCertified(t *testing.T) {
+//     dealer := genDealer()
+//     aggr := dealer.aggregator
+
+//     for i := 0; i < aggr.t; i++ {
+//         aggr.responses[uint32(i)] = &Response{Status: StatusApproval}
+//     }
+
```
nikkolasg: Feel free to add a settimeout here as well. We dont want to erase the rest of the test just because of this missing call.

share/vss/pedersen/vss_test.go
```diff
@@ -80,428 +81,505 @@ func TestVSSWhole(t *testing.T) {
...
+    // remaining verifiers
+    v.setTimeOut()
+
+    // Deal must be certified now
+    assert.True(t, aggr.DealCertified())
+    assert.NotNil(t, v.Deal())
+}
+
+
+// func TestVSSAggregatorVerifyDeal(t *testing.T) {
```
nikkolasg: Why is this one not working ? Again, don't hesitate to add the revelant `setTimeout` call in the test to make it pass, since the test unit here want to test for the correctness of the created share, so it's not the goal to test for the timeout thing.

share/vss/pedersen/vss_test.go
CedricCook: dealer.SecretCommit() returns nil when !DealCertified. So same problem as before, since the deal isn't certified unless the remaining verifiers are marked as timed-out.

share/vss/pedersen/vss_test.go
CedricCook: I've added the timeout, but just realised that the test is probably not working as intended. It checks for the bad dealer case, and then replaces all responses with Complaints without resetting `badDealer = False`, thus the test would always pass. I'll add the line to set the badDealer correctly.

share/vss/pedersen/vss_test.go
CedricCook: This was failing because of a call to deal.RndShare instead of deal.SecShare, so probably a copy paste error from the Rabin tests. Fixed.

share/vss/pedersen/vss_test.go
nikkolasg: Can you make it work with setTimeout then ?

nikkolasg:
@CedricCook Push your changes on your branches so it reflects here when you done solving the remaining issues. Thanks.

PR #244, opened 2017-11-17
Xof: extendable output functions

jeffallen:
A first draft of the Xof and it's application in the context of sign/anon.

Todo still:
  * Get help from a Real Cryptographer for explaining the limits of Extract()
  * replace kyber.Stream every place else


xof.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Return the sponge's secret state capacity in bytes.
+	Capacity() int
+
+	// Clone clones a Sponge.
+	Clone() Sponge
+}
+
+// Xof is an interface for extendable output functions.
+type Xof interface {
```
bford: XOF (all caps) since it's an acronym?

xof.go
```diff
@@ -0,0 +1,53 @@
+package kyber
+
+import (
+	"crypto/cipher"
+)
+
+// Sponge is an interface representing a primitive sponge function.
+type Sponge interface {
```
bford: Hmmm, not sure if it still makes sense to keep this intermediate Sponge interface or the associated code.  Is it or will it be actually used in the SHAKE and/or BLAKE2 implementations of the XOF interface?

xof.go
```diff
@@ -0,0 +1,53 @@
...
+	// TODO: Explain what the theoretical limit is on extraction.
+	Extract(dst []byte)
+
+	// An Xof can be used as a ciper.Stream. It uses Extract to
+	// take same number of bytes from the key stream as the length
+	// of src.
+	cipher.Stream
+
+	// Rate returns the rate of the underlying sponge.
+	Rate() int
```
bford: This method probably doesn't make sense anymore, since XOFs can be but aren't necessarily built on sponge primitives.  Does anything actually need to call this method?

xof/norx/sponge.go
```diff
@@ -0,0 +1,159 @@
+// Package norx implements a cryptographic sponge based on the experimental NORX cipher.
```
bford: Didn't we decide that norx will be left out of v1 (or moved to "experimental" status)?

xof.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Clone clones the Xof.
+	Clone() Xof
+}
```
bford: Didn't we decide that we want it to have an XORKeyStream function, so that it compatibly extends the Go standard cipher.Stream interface?  (Perhaps this might replace the Extract method.)

xof.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Xof is an interface for extendable output functions.
+type Xof interface {
+	// Absorb introduces state into the internal state of the Xof.
+	// Use it to put in keying material, or to put in material to be
+	// hashed. If key does not fill the internal state until Rate(),
+	// it will be padded with zeros until it does. Absorb returns
+	// itself so that chaining during creation is possible:
+	//     x := xof.New("keccak").Absorb(seed)
+	Absorb(key []byte) Xof
```
bford: Perhaps the "absorb" method(s) should be made compatible with the corresponding ones in the Go standard Hash interface?

bford: In fact, a fairly reasonable potential definition of this interface might be:

type XOF interface {
    hash.Hash
    cipher.Stream
}

...maybe along with a Clone method added.  That way an XOF can be used anywhere a hash is used, which makes sense; in that usage the Sum() and Size() methods should return the "default" output sizes of the XOF, which should be the minimum sizes to provide strong security (namely 2x the bits of the base security parameter, e.g., 256 bits for SHAKE128 or 512 bits for SHAKE256).  So anyone can use the XOF as a Hash if they want to, but can alternatively Write (absorb) stuff into it on input and then start calling XORKeyStream (using it as a cipher.Stream) to switch into output mode and suck an arbitrary amount of data out of it incrementally.  Does that make sense?

jeffallen:
I am closing this because I've got a much better implementation to look at coming soon.

PR #246, opened 2017-11-22
XOF: extendable output functions

jeffallen:
This is an implementation of XOFs, which can replace the
existing Cipher implementation. It is simpler to reason about
because it does not mix message oriented and streaming ciphers.

There are currently two implementations, using the Blake2b
and Shake256 hashes.


xof.go
```diff
@@ -0,0 +1,25 @@
...
+	io.Writer
+
+	// Read reads more output from the hash. It returns io.EOF if the limit
+	// has been reached.
+	io.Reader
+
+	cipher.Stream
+
+	Clone() XOF
+	KeySize() int
```
nikkolasg: What is KeySize ? It seems to create a XOK you dont provide a key but a seed so what is the link ?

jeffallen: I *think* it is the minimum amount of data that must be fed into to XOF to ensure that the full amount of output can be taken from it. For example, for blake, it is passed through from: https://godoc.org/golang.org/x/crypto/blake2b#BlockSize

But honestly, this is a part of the XOF that I don't feel I understand too well.

nikkolasg: From https://godoc.org/golang.org/x/crypto/blake2b#New256 it seems the key has no specific requirement on its length and since it returns a variable length hash function it's still a hash function and should not require any conditions on the input, no ? At least the doc don't mention anything in particular regarding the size of the input, or am I missing stg ?

proof/proof.go
```diff
@@ -17,8 +17,8 @@ import (
 type Suite interface {
 	kyber.Group
 	kyber.HashFactory
```
jeffallen: @bford Since XOFs can serve perfectly well as Hashes, what if we removed hashing as a primitive from kyber. Any place someone wanted to hash, they would use the XOF from their chosen suite.

xof.go
```diff
@@ -0,0 +1,37 @@
...
+	// it is ready to give secure output.
+	KeySize() int
+
+	// Reseed makes an XOF writeable again after it has been read from.
+	Reseed()
+}
+
+// An XOFFactory is an interface that can be mixed in to local suite definitions.
+type XOFFactory interface {
+	XOF([]byte) XOF
```
nikkolasg: A small comment explaining what is expected as the argument would be nice !

proof/proof.go
nikkolasg: Mhhhh. I would very much like that option to happen but we need to be careful. Some uses of `Hash` expects a fixed size output. Some uses kinda "know" that `sha256` is used, so it's compatible with external implementations (and that's bad or good depending on the point of view).

Looking at the `XOF` interface, it's not crystal clear to me how to provide a fixed size output. Should the seed be of the same size of the output's size I'm expecting ?

bford: Indeed, it would seem to be very nice for users of XOF to be able to use any XOF instance as a hash.Hash as well; i.e., if XOF might simply inherit from the standard hash.Hash.  

Since an XOF is semantically more powerful than a standard hash (includes all the capability of a Hash and then some), the fact that hash.Hash includes methods for producing fixed-length outputs is probably not a problem.  In fact having those methods as part of the XOF interface is probably good, too, because in many uses (of XOFs as well as hashes), we're probably just going to want "an output large enough to be secure but no larger".  In other words, some clients will want to use an XOF to produce an output whose size is determined by the client, whereas other clients will want to use an XOF to produce an output whose size the client does *not* care about, except that it should be "long enough to be secure" given the properties of the underlying XOF implementation in question (which is the XOF implementation's job to determine and not the client's).

Doing this would imply that any implementation of XOF would need to include hash.Hash, including the following two hash.Hash methods, which are specific to producing fixed-length output:

        // Sum appends the current hash to b and returns the resulting slice.
        // It does not change the underlying hash state.
        Sum(b []byte) []byte

        // Size returns the number of bytes Sum will return.
        Size() int

From a specification and documentation perspective, I would suggest that the XOF interface be extended to include a small block looking something like this:

type XOF interface {
   ... (existing stuff) ...

   // Any XOF may also be used as a Hash providing non-extensible output.
   hash.Hash

   // Size returns the minimum number of bytes of output that a client should read
   // from the XOF to produce a conventional hash with the XOF's full security strength.
   Size() int

   // Sum, a method of hash.Hash, reads Size bytes from the XOF's output and
   // appends them to input b (which may be null), returning the resulting output.
   Sum(b []byte) []byte
}

Yes, I know this will make the explicit inclusion of Size() and Sum() in the XOF interface redundant, in that we would get exactly the same interface in effect with just the hash.Hash line.  But from a documentation perspective, I'm suggesting that it's useful for the inteface to have these redundant definitions precisely so that the associated comment-documentation can explain precisely what the Sum() and Size() methods mean in the context of XOF.

Make sense and seem reasonable?  Other alternate suggestions?

xof.go
bford: One question related to the HashFactory and/or XOFFactory topic:  Is the "XOF version" of BLAKE2 domain-separated from the "hash version" of BLAKE2?  Meaning that calling BLAKE2-HASH("hello world") will produce a different output from calling BLAKE2-XOF("hello world", HASH_SIZE)?  I suspect it probably is domain-separated in this way (it's standard good crypto practice), but I haven't looked at the spec.  

Certainly in the case of SHA3 (hashes) versus SHAKE (XOFs), I believe these are domain separated so that no SHA3 hash can have a domain collision with any SHAKE or vice versa.

At any rate, for flexibility and binary protocol compatibility we'll still need to be able to distinguish from a "hash function used as a hash" and an "XOF used as a hash", for domain-separation purposes if nothing else.  For this reason, I suspect it's likely to be still useful to have both HashFactory and XOFFactory type interfaces, separately.  In fact a single "BLAKE2 factory" object might satisfy both HashFactory and XOFFactory interfaces, but produce instances of BLAKE2 that use different domain-separation values, one for "BLAKE2 as a hash" and one for "BLAKE2 as an XOF."  

So to implement an existing protocol that calls for a hash (e.g., BLAKE2-as-hash or SHA3), we may need the HashFactory interface to get at that, to produce a BLAKE2 instance on the proper "side" of the domain-separation fork for purposes of that particular protocol.

proof/proof.go
nikkolasg: I like this idea but I'm wondering if the `hash.Hash` is necessary at all in the end. If we clearly write the methods of `hash.Hash` directly in the `XOF` interface, then by Go rules, it implements the `hash.Hash` interface, and the documentation is clear what is means wrt to `XOF`.

xof.go
nikkolasg: Mhhh right. That's indeed a problem if we want to use XOF as a drop-in replacement for hash functions. Further tests are needed indeed.

jeffallen: Another problem I had when trying to implement this was that hash.Hash defines Sum as "Sum appends the current hash to b and returns the resulting slice. It does not change the underlying hash state." The obvious way to implement Sum on an XOF is to make a new []byte with length equal to the correct length for a hash of that type, then fill it with Read. But issuing a Read on an XOF changes the state of the XOF, which is against the hash.Hash spec.

So I'm going to stop working on removing HashFactory.

bford: Oh right - I'd forgotten about that particular behavior of Sum(), but now that you mention it I've encountered it before and the (minor) implementation challenges it creates here and there.  

One way to implement Sum() "generically", which I might have done at some point in the old Cipher code-base, is for Sum() to first Clone() the Cipher/XOF object, then call Read() on the clone, and otherwise just throw it away.  Not ideal in terms of either cleanness or memory-allocation efficiency though (memory allocations are moderately expensive in Go).

jeffallen: Unless you really push, I'm ready to stop working on this. The Cipher->XOF change is a huge improvement in the public API. It seems likely that trying to unify XOF and Hash will make XOF much dirtier, and in the likely case that some protocol needs vanilla standards-compliant hashing, they can't use XOF for that anyway.

If you agree, then let's focus reviews on the details of just what's here in this PR so that early next week I can check this in and move on to the next big cleanup.

I could especially use help understanding if Reseed is doing something cryptographically literate, and how to document KeySize, and how it is set for both impls.

(BTW: I have tested onet's kyber branch against this "Cipher-less" kyber, and there is no breakage.)

examples/sig_test.go
```diff
@@ -13,8 +13,8 @@ import (
 
 type Suite interface {
 	kyber.Group
-	Cipher(key []byte, options ...interface{}) kyber.Cipher
 	kyber.Encoding
+	kyber.XOFFactory
```
ineiti: I remember you were not fond of the `Factory`s - any better suggestion?

examples/sig_test.go
```diff
@@ -89,30 +89,30 @@ func SchnorrVerify(suite Suite, message []byte, publicKey kyber.Point,
 // Example of using Schnorr
 func Example_schnorr() {
 	// Crypto setup
-	group := edwards25519.NewAES128SHA256Ed25519()
-	rand := group.Cipher([]byte("example"))
+	suite := edwards25519.NewAES128SHA256Ed25519()
```
ineiti: Any ideas of what makes sense here? The `NewAES...` has more definitions than we need. Is that OK?

group/all.go
```diff
@@ -34,10 +31,19 @@ func init() {
...
 		return s, nil
 	}
 	return nil, ErrUnknownSuite
 }
+
+// MustSuite looks up a suite by name and panics if it is not found.
+func MustSuite(name string) kyber.Group {
+	s, err := Suite(name)
+	if err != nil {
+		panic("Suite " + name + " not found.")
```
ineiti: I try to avoid panic - isn't an error enough?

group/edwards25519/suite.go
```diff
@@ -9,24 +9,24 @@ import (
...
 	"github.com/dedis/kyber"
-	"github.com/dedis/kyber/cipher/sha3"
 	"github.com/dedis/kyber/group/internal/marshalling"
+	"github.com/dedis/kyber/xof/blake"
 )
 
-// SuiteEd25519 implements some basic functionalities such as Group, HashFactory
-// and CipherFactory.
+// SuiteEd25519 implements some basic functionalities such as Group, HashFactory,
+// CipherFactory and XOFFactory.
```
ineiti: I thought we got rid of `CipherFactory`?

group/edwards25519/suite.go
```diff
@@ -9,24 +9,24 @@ import (
...
 type SuiteEd25519 struct {
 	Curve
 }
 
-// Hash return a newly instanciated sha256 hash function
+// Hash return a newly instanciated sha256 hash function.
```
ineiti: s/return/returns/

group/edwards25519/suite.go
```diff
@@ -9,24 +9,24 @@ import (
...
 func (s *SuiteEd25519) Hash() hash.Hash {
 	return sha256.New()
 }
 
-// Cipher returns the SHA3/SHAKE128 Sponge Cipher
-func (s *SuiteEd25519) Cipher(key []byte, options ...interface{}) kyber.Cipher {
-	return sha3.NewShakeCipher128(key, options...)
+// XOF returns an XOF which is implemented via the Blake2b hash.
```
ineiti: Do you pronounce `Ksof` or `X.O.F.`? Just for the `an`/`a`...

group/edwards25519/suite.go
```diff
@@ -9,24 +9,24 @@ import (
...
+func (s *SuiteEd25519) XOF(key []byte) kyber.XOF {
+	return blake.New(key)
```
ineiti: Perhaps we should reflect the chosen XOF also in the `New...`-name? As we 'lost' the cipher, the correct name could be:
`NewBlakeSHA256Ed25519`
?

group/nist/qrsuite.go
```diff
@@ -26,9 +26,8 @@ func (s QrSuite) Hash() hash.Hash {
 	return sha256.New()
 }
 
-// SHA3/SHAKE128 Sponge Cipher
-func (s QrSuite) Cipher(key []byte, options ...interface{}) kyber.Cipher {
-	return sha3.NewShakeCipher128(key, options...)
+func (s QrSuite) XOF(key []byte) kyber.XOF {
```
ineiti: Same thing as in ed25519 - perhaps change the name of the suite to:

`NewBlakeSHA256QR512`
?

group/nist/suite.go
```diff
@@ -24,9 +24,8 @@ func (s *Suite128) Hash() hash.Hash {
 	return sha256.New()
 }
 
-// SHA3/SHAKE128 Sponge Cipher
-func (s *Suite128) Cipher(key []byte, options ...interface{}) kyber.Cipher {
-	return sha3.NewShakeCipher128(key, options...)
+func (s *Suite128) XOF(key []byte) kyber.XOF {
```
ineiti: And one more:

`NewBlakeSHA256P256`

?

sign/anon/enc_test.go
```diff
@@ -80,16 +87,24 @@ func ExampleEncrypt_anonSet() {
...
-	// 00000010  be 1d 8c fc 52 b1 85 98  31 a9 81 08 24 bb f0 d0  |....R...1...$...|
-	// 00000020  db 47 f9 b3 ee 3b 14 f2  2d 8f 0a c9 83 9d 47 1a  |.G...;..-.....G.|
-	// 00000030  69 0f a4 b2 5b 44 c8 a0  ca 33 1e c6 04 9d 98 35  |i...[D...3.....5|
-	// 00000040  31 cd 3a a9 0b 44 64 d5  a5 54 d4 5d 33 67 9e 2e  |1.:..Dd..T.]3g..|
-	// 00000050  35 e8 05 f3 17 c9 a5 14  9f 5b b6 9a c3 ee 57 54  |5........[....WT|
-	// 00000060  64 2f c2 06 36 ae aa af  f8 61 9d c3 cd 09 c2 d7  |d/..6....a......|
-	// 00000070  74 8d 32 bf 08 cb ef 1d  06 af 35 52 99 1f b1 16  |t.2.......5R....|
-	// 00000080  a7 3c 1b 02 8a 5f bd eb  f0 28 94 df 36 44 07 be  |.<..._...(..6D..|
-	// 00000090  22 01 7c dc ad 06 09 7a  62 8e 45 98              |".|....zb.E.|
+	// 00000000  1d 94 9b 66 98 f7 66 ec  2c 66 7f 00 1a aa 90 59  |...f..f.,f.....Y|
```
ineiti: Do you know why this is nearly double the length?

xof.go
```diff
@@ -0,0 +1,44 @@
+package kyber
+
+import (
+	"crypto/cipher"
+	"io"
+)
+
+// An XOF is an extendable output function, which is a cryptographic
+// primitive that can take arbitrary input like a hash, and then create
```
ineiti: I'd remove `like a hash` - it made me think that the input should be a hash.

xof.go
```diff
@@ -0,0 +1,44 @@
...
+// a stream of output, up to a limit determined by the size of the internal
+// state of the XOF.
+type XOF interface {
+	// Write absorbs more data into the hash's state. It panics if called
```
ineiti: is it the hash's state or the XOF's state?

xof.go
```diff
@@ -0,0 +1,44 @@
...
+	// after Read. Use Reseed() return an XOF into a state where more data
```
ineiti: s/return an/to reset the/

xof.go
```diff
@@ -0,0 +1,44 @@
...
+	// can be absorbed via Write.
+	io.Writer
+
+	// Read reads more output from the hash. It returns io.EOF if the limit
```
ineiti: if which limit has been reached?

xof.go
```diff
@@ -0,0 +1,44 @@
...
+	// has been reached.
+	io.Reader
+
+	// An XOF implements cipher.Stream, so that callers can use XORKeyStream
+	// to encrypt/decrypt data. The key stream is read from the XOF using
+	// the io.Reader interface.
```
ineiti: What happens if the limit is reached during XORKeyStream?

xof.go
```diff
@@ -0,0 +1,44 @@
...
+	cipher.Stream
+
+	// KeySize is the number of bytes that must be written to the XOF before
+	// it is ready to give secure output.
+	KeySize() int
```
ineiti: What happens if a read takes place before that limit is reached?

xof/blake/blake.go
```diff
@@ -0,0 +1,82 @@
...
+	if seed2 != nil {
+		_, err := b.Write(seed2)
+		if err != nil {
+			panic("blake2b.XOF.Write should not return error: " + err.Error())
+		}
+	}
+	return &xof{impl: b}
+}
+
+func (x *xof) KeySize() int {
```
ineiti: No need for comments because it implements the interface?

xof/xof_test.go
```diff
@@ -0,0 +1,333 @@
...
+	xof2.Read(dst2)
+
+	d := bitDiff(dst1, dst2)
+	if math.Abs(d-0.50) > 0.1 {
+		t.Fatalf("reseed bitDiff %v", d)
+	}
+}
+
+//
+// TODO: port this test to XOF
```
ineiti: Is this code still needed? Delete? Rewrite?

examples/sig_test.go
jeffallen: Nope, it does what it says, so no complaint from me. Based on Java nightmares, I don't like factories, but kyber uses them in a reasonable way so let's call it what it is.

examples/sig_test.go
jeffallen: I don't understand the question.

group/all.go
jeffallen: By analogy to https://godoc.org/regexp#MustCompile
"It simplifies safe initialization of global variables holding compiled regular expressions."

group/edwards25519/suite.go
jeffallen: Fixed.

group/edwards25519/suite.go
jeffallen: Fixed.

group/edwards25519/suite.go
jeffallen: I say ex.oh.ef in my head. Thus the "an".

examples/sig_test.go
ineiti: With the new definition of `Suite`, we tell the outside world what we need. In this case it's: `Group`, `Encoding`, and `XOFFactory`. But the `New...` gives also the hash and the symmetric encryption. But that's probably not a problem, is it?

group/all.go
ineiti: And where is it used?

group/edwards25519/suite.go
ineiti: OK, ex.oh.ef in the head and ksof when speaking - I wonder what that means for usage of `an`/`a` ;)

group/edwards25519/suite.go
jeffallen: Fixed.

group/nist/qrsuite.go
jeffallen: Fixed.

group/nist/suite.go
jeffallen: Fixed.

sign/anon/enc_test.go
jeffallen: This was because I used xof.KeySize() as the size of the output hash. I have added a new xof.Size() method, which is set correctly for both blake and keccak and is used by sign/anon now.

xof.go
jeffallen: Fixed to be clearer.

xof.go
jeffallen: The underlying hash's state. Fixed to be hopefully clearer.

xof.go
jeffallen: Done.

xof.go
jeffallen: Fixed to be clearer.

xof.go
jeffallen: Panic. But the limit for blake is 2^31 and there is no limit for sha3.Shake256.

xof.go
jeffallen: It gives not-guaranteed-secure output.

xof/blake/blake.go
jeffallen: yes.

xof/xof_test.go
jeffallen: I don't really understand what it is doing, or even if it is a valid/useful test for XOF. I made #248 to remember it.

xof.go
```diff
@@ -0,0 +1,50 @@
...
```
nikkolasg: I'm very surprised about this. As I understood it, and from Bryan's comment, a XOF is supposed to be kinda of "superset" of a hash function. A Hash function, even fed with one byte outputs a "safe" output in the sense that it's looking random. Why does a XOF do not have this property ? Furthermore, I can't find any restrictions of the kind in  https://godoc.org/golang.org/x/crypto/blake2b for example.

jeffallen: For blake, I set KeySize to blake.BlockSize, but I can't remember why. I need advice from a crypto expert on this, because I am having trouble sorting out what Cipher did from what I guessed myself while learning this...

It seems to me that the caller needs some way to know what the maximum number of bits the underlying sponge can take up is, so that they can decide how much original key material to provide. Because while it is true that the output of the XOF is indistinguishable from random with zero or 1 bits of key, giving a tiny key leaves an attacker a tiny search space.

Perhaps it's a terminology problem: reading from an XOF is like getting a hash of the input, and there size does not matter. But reading from an XOF in order to call XORKeyStream is encrypting/decrypting, and then input size matters, A LOT. This is why I called it "KeySize".

sign/anon/enc_test.go
ineiti: Thumbs-up is: OK, won't bother you with this anymore ;)

xof.go
ineiti: Should this be put in the comments?

xof/blake/blake.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (x *xof) XORKeyStream(dst, src []byte) {
+	if len(dst) < len(src) {
+		panic("dst too short")
+	}
+	if len(x.key) < len(src) {
+		x.key = make([]byte, len(src))
+	}
+
+	n, err := x.Read(x.key)
```
nikkolasg: Does `key` has to be a member of the `xof` struct ? I don't see any use for it to be, you never refer to `x.key` elsewhere.

xof/blake/blake.go
```diff
@@ -0,0 +1,86 @@
...
+	// Steal the XOF implementation, and put it inside of x.
+	x.impl = y.(*xof).impl
+	return
+}
+
+func (x *xof) XORKeyStream(dst, src []byte) {
+	if len(dst) < len(src) {
+		panic("dst too short")
+	}
+	if len(x.key) < len(src) {
```
nikkolasg: I think we need the current `x.key` to be *as long* as the src no ? No need to read more than we need to. Plus it makes things easier to think about with respect to side channels attacks (maybe something can be deduced from the fact the source is shorter or not than the key during the XOR etc).

xof.go
jeffallen: Done

xof/blake/blake.go
jeffallen: This is so that repeated calls to XORKeyStream will not make garbage. But I found a bug while re-reading this. Calling XORKeyStream with src and later with src2 where len(src2) < len(src) would have resulted in len(src) bytes being read from the XOF, possibly causing a decrypt with different chunk alignment to fail. Fixed.

xof/blake/blake.go
jeffallen: (yeah, I noticed this while fixing last comment)

xof.go
nikkolasg: Yes I agree for your last paragraph. So `KeySIze` means `the maximum key size you can give to the XOF,i.e. the required key size if you wanna be secure`, but the `Size` ... I'm not sure we need it in fact ? It's confusing for me at least, maybe @Daeinar can chim in. It seems we don't really need the `Size` then because we should be able to retrive any number of bytes from a XOF, so it's not clear what this size represents.

nikkolasg: and BTW, we need to be more consistent: what's the difference between `seed` and `key`. I create a `XOF` with a `seed` but I can retrieve its `KeySize` ? Same for in-method code.

jeffallen: I added Size to help sign/anon know how big a MAC should be. Before this, it was using KeySize() to decide how much to take out of the hash, and was making signatures bigger than it needed to.

nikkolasg: From what you are saying it seems like it's `sign/anon` that does a bad job in the sense that it should know how many random looking bytes it has to take no ? I mean random looking bytes from sha3 or blake2 are still supposed to be random looking so regarding the security, the size of the hash to take is not dependant on the underlying hash function (as long as the hash function is correct), right ?
Is there no way to make `sign/anon` work while removing that `Size` function ?

jeffallen: Sure, just hard code it to be what it was before. Doing that now.

PR #249, opened 2017-11-28
Use Go's ModSqrt and Jacobi instead of our own

### LINKED ISSUE #231: kyber/group/mod/sqrt.go
This implementation can probably be replaced with the one that is now in Go math/big:

https://godoc.org/math/big#Int.ModSqrt

jeffallen:
Closes #231.


Daeinar:
Looks good

bford:
Have you double-checked that the Go standard versions are "just as constant-time" as ours?

jeffallen:
Yes. They use the same algorithm. As far as I can see, both are NOT constant time. Is that what you expect?

Also: While developing this, I went through several steps. My first (incorrectly implemented) attempt caused test failures, so then I rewrote it to calculate Go's answer and our answer side by side and panic on difference. Our test suite passed in that mode .Then I switched it to using Go's results and it continued passing, and that showed me how I'd screwed up the first time. I am confident that our test suite covers this code.

bford:
It doesn't surprise me that they're both not constant-time, but I just couldn't remember for sure.  Fixing that will presumably be part of the much bigger challenge of making big.Int support constant time in general; see this Golang proposal thread I started a while ago on this topic: https://github.com/golang/go/issues/20654

PR #251, opened 2017-11-30
Refine caller opt-in to vartime implementations

### LINKED ISSUE #250: Remove SetVarTime from interface Point (but still support the concept for callers who opt-in)
First principles:
1. smaller interfaces are better
2. YAGNI (you aren't gonna need it)

So the goal here is to make it possible to do this, but not have it be in every implementation of kyber.Point.

(this comment was edited based on Bryan's feedback in comment 2 and 3)

jeffallen:
Remove SetVarTime from the group interface, to make it lighter.

Create a new interface to allow callers to detect if vartime
is available and use it for public data if they so choose.

In group/edwards25519, make the existing vartime algorithm
correctly built only with build tag vartime.

Closes #250.


group.go
```diff
@@ -118,13 +111,18 @@ type Point interface {
...
 	Mul(s Scalar, p Point) Point
+}
 
-	// SetVarTime allows or disallows use of faster variable-time implementations
-	// of operations on this Point. It returns an error if the desired
-	// implementation is not available.
-	// This flag always defaults to false (constant-time only)
-	// in implementations that can provide constant-time operations.
-	SetVarTime(varTime bool) error
+// AllowsVarTime allows callers to determine if a given kyber.Scalar
```
bford: VERY nice design and explanatory comment, thanks!

group/edwards25519/ge_mult.go
```diff
@@ -0,0 +1,8 @@
+// +build !vartime
+
+package edwards25519
+
+func geScalarMultVartime(h *extendedGroupElement, a *[32]byte,
```
ineiti: Add a small comment saying this is needed for linking.

PR #253, opened 2017-12-01
A setting to help with debugging

jeffallen:
And a test to make sure it is never, ever turned on in
checked in code.


ineiti:
Fix coverage - find 0.7% somewhere ;)

PR #254, opened 2017-12-01
[VSS] Change VSS to work only with Approvals

CedricCook:
VSS: add SetTimeout, makes sure that all verifiers have either responded, or have a response StatusComplaint set.
DKG: add bypass method on VSS to allow DKG to continue functioning
DKG: use & expose VSS.SetTimeout method


share/vss/pedersen/vss.go
```diff
@@ -537,6 +563,20 @@ func (a *aggregator) VerifyDeal(d *Deal, inclusion bool) error {
 	return nil
 }
 
+// cleanVerifiers checks the aggregator's response array and creates a StatusComplaint
+// response for all verifiers who have no response in the array.
```
nikkolasg: '... for all verifiers that did not respond to the Deal`

share/dkg/pedersen/dkg.go
```diff
@@ -242,6 +247,14 @@ func (d *DistKeyGenerator) ProcessJustification(j *Justification) error {
 	return v.ProcessJustification(j.Justification)
 }
 
+// SetTimeout triggers the timeout on all verifiers, and thus makes sure
+// all verifiers have either responded, or have a StatusComplaint response.
```
nikkolasg: The user should only need to care about "what this functions does, when should I call it, what's the output"
so more something like:
```
SetTimeout marks the end of a round, invalidating any missing (or future) response for this DKG protocol round. The caller is expected to call this after a long timeout so each DKG node can still compute its share if enough Deals are valid.
```
stg along those lines.

share/dkg/pedersen/dkg_test.go
```diff
@@ -492,6 +491,56 @@ func TestDKGProcessDeal(t *testing.T) {
...
+
+	// 3. make sure everyone has the same QUAL set
+	for _, dkg := range dkgs {
+		for _, dkg2 := range dkgs {
+			require.False(t, dkg.isInQUAL(dkg2.index))
+		}
+	}
+
+	for _, dkg := range dkgs {
+		dkg.SetTimeout()
```
nikkolasg: This should not be needed because all responses are supposed to be dispatch. So you must first make this test fail by not `ProcessResponse` one response on one node, check that it fails, then call `SetTimeout` and check that it succeeds after.

CedricCook: I think that is what I tried to simulate? in line 513 I only let nodes processResponses until they have the strict minimum approvals (>= t). I then check that not all DKGs are in all other DKGs's qualifying set, and then I call SetTimeout and check that the qualifying sets are complete this time. Is that incorrect?

nikkolasg: Oups, you're definitely right, I glossed over the test and thought it was the same as before. My bad!

PR #255, opened 2017-12-04
Work around incorrect Exp

### LINKED ISSUE #245: Audit for "r.Exp(x, 1, m) wrong if r is initially non-zero"
In golang/go#22830, a posible security problem for bignum users is reported.

We need to find out if this affects us.

jeffallen:
The consequences of golang/go#22830 could be serious, so out of an abundance
of caution fix these up.

Closes #245.


ineiti:
@kc1212 - can you add a test to make sure this is correct now?

jeffallen:
Our existing tests were already covering this code. I even added a panic("got here") in there, to be sure (because I don't completely trust coveralls yet).

Issue #257, opened 2017-12-04
remove random.Stream


jeffallen:
...and make the heavy API changes necessary so that all places it was used can specify a seeded RNG if they need stability during testing.

See #253 for the hacky way we make this possible today, that we want to remove for v1, hopefully.

ineiti:
Thinking out loud: this means that all signing/encrypting methods will need to take a `random-Stream` as input parameter, so that tests can give a 'static' random for reproducible tests.

jeffallen:
Yes, unless you add the "current random stream" to the suite. Which would be sweet, but my first experiments with it were fragile and would have further increased the burden of explaining the Suite system to new programmers. So I am taking the more conservative, idiomatic Go approach.

nikkolasg:
Sorry I did not follow through but the `random.Stream` object is quite useful to generate truely random keys to use in production, and not in testing. How would we achieve that now ?

PR #258, opened 2017-12-07
Replace random.Stream global

jeffallen:
Globals are bad, bad for testing and predictability. Better to require the caller to decide where they get random data from.

Also, reduce the surface area of the Kyber API further by removing util/bytes.

Closes dedis/kyber#257.


jeffallen:
This is obsolete. Expect a replacement branch called norandsuite coming soon.

PR #261, opened 2017-12-07
Move random into Suite

### LINKED ISSUE #257: remove random.Stream

jeffallen:
Before randomness was passed in via arguments, or sometimes
taken from the global random.Stream. With this change, randomness
comes from the Suite, simplifying the APIs and also making it
possible for tests to elect to have predictable random numbers.

Closes #257.


group/curve25519/curve.go
```diff
@@ -391,3 +388,16 @@ func (c *curve) data(x, y *mod.Int) ([]byte, error) {
 	}
 	return b[1 : 1+dl], nil
 }
+
+// reverse copies src into dst in byte-reversed order and returns dst,
+// such that src[0] goes into dst[len-1] and vice versa.
+// dst and src may be the same slice but otherwise must not overlap.
+func reverse(dst, src []byte) []byte {
+	l := len(dst)
+	for i, j := 0, l-1; i < (l+1)/2; {
```
ineiti: code:
```
for i, j := 0, l-1; i < (l+1)/2; i++, j--
```

group/edwards25519/scalar.go
```diff
@@ -2231,3 +2230,16 @@ func scReduce(out *[32]byte, s *[64]byte) {
 	out[30] = byte(s11 >> 9)
 	out[31] = byte(s11 >> 17)
 }
+
+// reverse copies src into dst in byte-reversed order and returns dst,
+// such that src[0] goes into dst[len-1] and vice versa.
+// dst and src may be the same slice but otherwise must not overlap.
+func reverse(dst, src []byte) []byte {
+	l := len(dst)
+	for i, j := 0, l-1; i < (l+1)/2; {
```
ineiti: code:
```
for i, j := 0, l-1; i < (l+1)/2; i++, j--
```

group/mod/int.go
```diff
@@ -469,3 +468,19 @@ func (i *Int) HideDecode(buf []byte) {
 	i.V.SetBytes(buf)
 	i.V.Mod(&i.V, i.M)
 }
+
+// reverse copies src into dst in byte-reversed order and returns dst,
+// such that src[0] goes into dst[len-1] and vice versa.
+// dst and src may be the same slice but otherwise must not overlap.
+func reverse(dst, src []byte) []byte {
```
ineiti: Now we use it three times - might be worth exporting it here, no?

group/curve25519/curve.go
jeffallen: That's not vaild Go. Only one expression is allowed after the second semi-colon. (surprised me too!)

group/mod/int.go
jeffallen: exporting into which package? into util/bytes? The point of kyber is to provide crypto primitives, not byte reversing. A new reader exploring the API should find things in util/ that are related to crypto, like encoding scalars and points, making keys. Not reversing bytes.

jeffallen:
The way to override is implemented only in Edwards25519 curves for the moment: edwards25519.NewBlakeSHA256Ed25519WithRand(blake.New(nil)).

It is used in sign/anon/enc_test.go.

I will call more attention to it when I go through the docs this week.

group/curve25519/curve.go
ineiti: Now I see how they did it at https://github.com/golang/go/wiki/SliceTricks :
```
for i, j := 0, l-1; i < (l+1)/2; i, j = i+1, j-1
```

PR #262, opened 2017-12-08
Fix scalar encoding

jeffallen:
Because scalar encoding depended on converting to a mod.Int
first, the most significant bytes could be lost.

Also fix problems in the place that decodes it, which was hiding
the real error.

### DIFF
--- a/util/encoding/encoding.go
+++ b/util/encoding/encoding.go
@@ -144,14 +144,14 @@ func write64(wc io.WriteCloser, data ...kyber.Marshaling) error {
 	return wc.Close()
 }
 
-func getHex(r io.Reader, len int) ([]byte, error) {
-	bufHex := make([]byte, len*2)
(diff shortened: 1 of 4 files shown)


kc1212:
looks good

util/encoding/encoding.go
```diff
@@ -144,14 +144,14 @@ func write64(wc io.WriteCloser, data ...kyber.Marshaling) error {
 	return wc.Close()
 }
 
-func getHex(r io.Reader, len int) ([]byte, error) {
```
ineiti: I'll go and do penitence - that was mine...

jeffallen:
```
r := rules.New("Don't shadow keywords.")
r.Penance = func () { for range dedis.Engineers() { 🍰.buy() } }
```

PR #266, opened 2017-12-12
Convert proof and shuffle to the new random API.

jeffallen:
Also: Mark them experimental.


ineiti:
Looks good to me. Please also remove the experimental directories from `stable/directories`

PR #274, opened 2018-01-30
Fix endianness of SetBytes

ecnerwala:
The Scalar documentation says that SetBytes uses big-endian integers, and scalar.Bytes() in fact does use big-endian. Fix SetBytes to use it too.

### DIFF
--- a/group/edwards25519/scalar.go
+++ b/group/edwards25519/scalar.go
@@ -136,9 +136,9 @@ func (s *scalar) Pick(rand cipher.Stream) kyber.Scalar {
 	return s.setInt(i)
 }
 
-// SetBytes s to b, interpreted as a little endian integer.
+// SetBytes s to b, interpreted as a big endian integer.
 func (s *scalar) SetBytes(b []byte) kyber.Scalar {
-	return s.setInt(mod.NewIntBytes(b, primeOrder, mod.LittleEndian))
+	return s.setInt(mod.NewIntBytes(b, primeOrder, mod.BigEndian))
 }
 
 // Bytes returns a big-endian representation of the scalar


ineiti:
Thanks for the PR - but now all our tests are failing, because they suppose that `SetBytes` uses little-endian. Can you point me to where the code says `scalar.Bytes()` uses big-endian? We're cleaning up this code and my understanding was that all `Scalar` representations will be little-endian.

ecnerwala:
https://github.com/dedis/kyber/blob/de19cc714aa0e6931ed9e15be42028e26d912b38/group.go#L56 in the main Scalar interface specifies that it should be Big-Endian. (I think this is to be consistent with big.Int.)

Also, `Bytes()` (right below) returns a big-endian representation. At the very least, Bytes() and SetBytes() should have consistent endianness.

nikkolasg:
MMhhh I see. The problem is that we only used (and maybe tested) `Marshal` and `Unmarshal` but never `Bytes` and `SetBytes`. Thr formers are using little endian representation but there is indeed a inconsistency with the `Bytes` and `SetBytes` methods. 

Regarding the fix, I'd be more in favor of changing `Bytes` to return little endian number, exactly how a `edwards25519.Scalar`'s buffer is represented.

ecnerwala:
I think it's better to obey the interface, especially since it allows interop with big.Int (you can cast Scalars to/from big.Int via big-endian arrays). It is a bit of a performance hit, though.

jeffallen:
I'm currently trying to find out what would have to change to implement this completely, so that tests pass. It seems like calls to SetBytes (would become always BE with this change) would need to change to calls to UnmarshalBinary (always LE).

jeffallen:
I am working on a fix for this in #276. Closing this.

Thanks for your help noticing this.

PR #276, opened 2018-02-01
Edwards25519 SetBytes must follow the interface definition

### LINKED ISSUE #275: SetBytes endianess should be better documented
This was first reported by @ecnerwala in PR #274.

After a lot of consideration and prototyping other fixes, we've decided it is ok for this to stay implementation-defined.

jeffallen:
SetBytes is defined in kyber.Suite as bytes are big endian
byte order, so things to work like that.

Marshal/UnmarshalBinary is still LE, so that we interoperate with
the rest of the world's Ed25519 implementations.

Also: Remove some dead code I noticed.

Also: Package encoding had circular dependency problems. Fix
with the _test trick.

Fixes #275.


group/edwards25519/curve.go
```diff
@@ -51,6 +51,5 @@ func (c *Curve) NewKey(stream cipher.Stream) kyber.Scalar {
 	scalar[31] &= 0x3f
 	scalar[31] |= 0x40
 
-	secret := c.Scalar().SetBytes(scalar[:32])
-	return secret
+	return c.Scalar().SetBytes(scalar[:32])
```
nikkolasg: In fact here we want a non-reduced one, secret key is never modified "as-is" in the RFC.

group/edwards25519/scalar.go
```diff
@@ -175,8 +181,8 @@ func (s *scalar) MarshalBinary() ([]byte, error) {
 
 // UnmarshalBinary reads the binary representation of a scalar.
 func (s *scalar) UnmarshalBinary(buf []byte) error {
-	if len(buf) != 32 {
-		return errors.New("wrong size buffer")
+	if len(buf) < 32 {
```
nikkolasg: Why is that ? UnmarshalBinary should only consider valid scalars, i.e. 32 bytes. SetBytes takes whatever and reduces it to 32 bytes.

group/edwards25519/scalar.go
```diff
@@ -136,15 +137,20 @@ func (s *scalar) Pick(rand cipher.Stream) kyber.Scalar {
 	return s.setInt(i)
 }
 
-// SetBytes s to b, interpreted as a little endian integer.
+// SetBytes s to b, interpreted as a big endian integer.
+// The scalar will then be reduced modulo the prime order.
 func (s *scalar) SetBytes(b []byte) kyber.Scalar {
-	return s.setInt(mod.NewIntBytes(b, primeOrder, mod.LittleEndian))
+	return s.setInt(mod.NewIntBytes(b, primeOrder, mod.BigEndian))
```
nikkolasg: And to be able to save a useless allocation, we can use `scReduce` (removed in this PR) instead of allocation a big.Int and making a reverse copy in LE of the buffer (just this line).

group/edwards25519/scalar.go
jeffallen: I think this is left over from when I was trying to send the results of the hash (64 bytes) straight into UnmarshalBinary. Will revert.

group/edwards25519/scalar.go
jeffallen: I wondered what scReduce was.

share/dss/dss.go
```diff
@@ -208,7 +209,9 @@ func (d *DSS) hashSig() kyber.Scalar {
 	_, _ = d.random.Commitments()[0].MarshalTo(h)
 	_, _ = d.long.Commitments()[0].MarshalTo(h)
 	_, _ = h.Write(d.msg)
-	return d.suite.Scalar().SetBytes(h.Sum(nil))
+	hBuff := h.Sum(nil)
+	encoding.ConvertEndian(hBuff, hBuff)
```
nikkolasg: a small comment explaining it's because of eddsa verification expecting LE would be worth it I think after all the trouble we had with this one :D

jeffallen:
(this is in WIP still because it also implies breaking changes in cothority)

I agree that this is a mess, and not necessarily the right way forward. Let's sleep on it. 😴

jeffallen:
We are going to fix the docs instead of the code. Closing.

PR #277, opened 2018-02-05
Improve documentation

jeffallen:
* Vartime
* SetBytes

Also: Simplify travis config.

### DIFF
--- a/group.go
+++ b/group.go
@@ -51,12 +51,11 @@ type Scalar interface {
 	// Set to a fresh random or pseudo-random scalar
 	Pick(rand cipher.Stream) Scalar
 
-	// SetBytes sets the scalar from a big-endian byte-slice,
+	// SetBytes sets the scalar from a byte-slice,
 	// reducing if necessary to the appropriate modulus.
+	// The endianess of the byte-slice is determined by the
+	// implementation.
 	SetBytes([]byte) Scalar
-
-	// Bytes returns a big-endian representation of the scalar
-	Bytes() []byte
 }
 
 // A Point kyber.y represents an element of a public-key cryptographic Group.
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: go
 
-go:
-  - 1.8.x
-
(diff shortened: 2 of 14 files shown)


nikkolasg:
Since this original bug described in the original issue (which I lost) was not covered by any tests, and we were lucky that someone fell on it, it would be nice to have a small test inside edwards25519 and p256 that checks if the endianess is takien correctly. Also, if we could add for each group what is the expected endianness  in the documentation of the respective scalar implementations and check that we can see it in `godoc`, that would be HUGE..

group.go
```diff
@@ -51,12 +51,11 @@ type Scalar interface {
 	// Set to a fresh random or pseudo-random scalar
 	Pick(rand cipher.Stream) Scalar
 
-	// SetBytes sets the scalar from a big-endian byte-slice,
+	// SetBytes sets the scalar from a byte-slice,
 	// reducing if necessary to the appropriate modulus.
+	// The endianess of the byte-slice is determined by the
+	// implementation.
 	SetBytes([]byte) Scalar
```
kc1212: @jeffallen I remember you suggested that maybe we don't need `SetBytes` in the interface because it's only really needed in one place, eddsa I believe, the other places can be replaced with `Pick`. Is there a decision on that?

nikkolasg: Let me recap the story:
When `edwards25519.Scalar` were using a `big.Int` (well a `mod.Int` to be exact) for its computation, it was not possible to have an EdDSA implementation because of the "automatic" modulo. That is the reason we added a `SetBytes()` method (and its counterpart `Bytes` which turned out to be useless). 
Since a few months already, Bryan ported the usual constant time implementation to `edwards25519.Scalar` , so that now it only uses a `[]byte`. I *think* (notice that uncertainty here) that with this change, the need for `SetBytes` (/`Bytes`) can be alleviated, i.e. we can remove those. Need for some tests to confirm that.

That's one way, Another way, which is easier but faster, but still correct and "good", is also to correct the doc as @jeffallen does here :)

PR #283, opened 2018-02-20
Integrate BN256 Pairing

Daeinar:
This PR integrates the BN256 pairing using Cloudflare's implementation, see https://github.com/cloudflare/bn256. It replaces the public BN256 interface (i.e., https://github.com/cloudflare/bn256/blob/master/bn256.go) by Kyber's usual `point` and `scalar` representation.


nikkolasg:
In the README I would add at the beginning that this is a copy of cloudflare's library.

nikkolasg:
Also if the implementation is complete, you can write a small test for testing G1 and G2 (maybe Gt will be more complex), there's already code that does that, so you only have to write one line. See https://github.com/dedis/kyber/tree/master/util/test

jeffallen:
Please make this two commits, one with Cloudflare's code completely unchanged, and one with our changes.

pairing/bn256/scalar.go
```diff
@@ -0,0 +1,148 @@
...
+func (s *scalar) SetBytes(buf []byte) kyber.Scalar {
+	s.UnmarshalBinary(buf)
+	return s
+}
+
+func (s *scalar) Bytes() []byte {
+	return s.x.Bytes()
+}
+
+func (s *scalar) SetVarTime(varTime bool) error {
```
jeffallen: This is no longer needed/wanted. The new way to do what that was doing is: https://godoc.org/github.com/dedis/kyber#AllowsVarTime

But in your case, since this is a 100% vartime library, it should be marked "vartime" with build flags. (AllowsVarTime is for a predominantly constant time library to let callers opt in to vartime on a point by point basis.)

Daeinar:
@nikkolasg Thanks for your comments. Yes this is not yet the final version, I just copy&pasted the existing code and started with the integration. The README will definitely change later on.

jeffallen:
If this is done, I'd be happy to take one final review and then merge it. Just move to R4M if so.

Daeinar:
@jeffallen I still need to figure out some minor things, like how to implement the embed functionality for G1, G2, and GT in a secure / reasonable way, see https://github.com/dedis/kyber/blob/pairing/pairing/bn256/point.go#L67-L70, and I would like to discuss this with @nikkolasg and you. Once that's done I think we can merge it.

Daeinar:
@jeffallen @nikkolasg Deactivated the data embedding for now. I think the PR is ready for review.

pairing/pairing.go
```diff
@@ -0,0 +1,15 @@
+package pairing
+
+import "github.com/dedis/kyber"
+
+// Suite ...
```
jeffallen: Are these ... a TODO?

Daeinar: Yes. I'll fix that later.

pairing/bn256/bn256.go
```diff
@@ -0,0 +1,97 @@
...
+func (s *SuiteBN256) G2() kyber.Group {
+	return s.g2
+}
+
+// GT returns the group GT of the BN256 pairing.
+func (s *SuiteBN256) GT() kyber.Group {
+	return s.gt
+}
+
+// Pair computes takes the points p1 and p2 in groups G1 and G2, respectively,
```
ineiti: s/computes takes/takes/

pairing/bn256/bn256.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Write is the default implementation of Encoding interface Write.
+func (s *SuiteBN256) Write(w io.Writer, objs ...interface{}) error {
+	return fixbuf.Write(w, objs)
+}
+
+// New implements the kyber.Encoding interface.
+func (s *SuiteBN256) New(t reflect.Type) interface{} {
+	return nil // TODO
```
ineiti: Should this `panic` because it's not implemented?

pairing/bn256/bn256_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/dedis/kyber/util/random"
+	"golang.org/x/crypto/bn256"
+)
+
+func TestG1(t *testing.T) {
+	suite := NewSuiteBN256()
+	k := suite.G1().Scalar().Pick(random.New())
+	pa := suite.G1().Point().Mul(k, nil)
+	ma, err := pa.MarshalBinary()
+	if err != nil {
```
ineiti: `require.Nil(t, err)` and other places, too...

pairing/bn256/bn256_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Fatal(err)
+	}
+
+	pb := new(bn256.G1).ScalarBaseMult(k.(*scalar).x)
+	mb := pb.Marshal()
+
+	if !bytes.Equal(ma, mb) {
```
ineiti: `require.Equal(t, ma, mb)`

pairing/bn256/constants.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"math/big"
+)
+
+func bigFromBase10(s string) *big.Int {
+	n, _ := new(big.Int).SetString(s, 10)
+	return n
+}
+
+// u is the BN parameter that determines the prime: 1868033³.
```
ineiti: Hmm - this is not a prime - so why does it `determine the prime`?

pairing/bn256/constants.go
```diff
@@ -0,0 +1,56 @@
...
+var u = bigFromBase10("6518589491078791937")
+
+// p is a prime over which we form a basic field: 36u⁴+36u³+24u²+6u+1.
+var p = bigFromBase10("65000549695646603732796438742359905742825358107623003571877145026864184071783")
```
ineiti: This is a prime ;)

pairing/bn256/constants.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Order is the number of elements in both G₁ and G₂: 36u⁴+36u³+18u²+6u+1.
+// order-1 = (2**5) * 3 * 5743 * 280941149 * 130979359433191 * 491513138693455212421542731357 * 6518589491078791937
+var Order = bigFromBase10("65000549695646603732796438742359905742570406053903786389881062969044166799969")
+
+// xiToPMinus1Over6 is ξ^((p-1)/6) where ξ = i+3.
```
ineiti: 2018 - the year of UTF-8!

pairing/bn256/gfp.h
```diff
@@ -0,0 +1,32 @@
+#define storeBlock(a0,a1,a2,a3, r) \
```
ineiti: Will this be compiled into the code? That might be worth a comment...

pairing/bn256/gfp.s
```diff
@@ -0,0 +1,95 @@
+#include "gfp.h"
```
ineiti: Same - where is this compiled?

pairing/bn256/gfp6.go
```diff
@@ -0,0 +1,213 @@
...
+
+	e.x.Set(c2)
+	e.y.Set(c1)
+	e.z.Set(c0)
+	return e
+}
+
+func (e *gfP6) Invert(a *gfP6) *gfP6 {
+	// See "Implementing cryptographic pairings", M. Scott, section 3.2.
+	// ftp://136.206.11.249/pub/crypto/pairings.pdf
```
ineiti: I always love ftp-references with IP addresses...

wouterl:
Please keep keep in mind that the BN256 curve has been shown to offer significantly less than 128 bits security. This [ZCash blog](https://blog.z.cash/new-snark-curve/) mention 110 bits, but I've also read claims of 96 bits real security before. There are more academic references if you need them. The [paper by Menezes, Sarkar and Sing](https://eprint.iacr.org/2016/1102.pdf) gives some recommendations.

pairing/bn256/gfp.h
kc1212: cgo will be somewhere if these are to be used, right? I don't see cgo anywhere in this PR

nikkolasg: I think it's already by default that you can compile this with `go` directly, cgo is included in the default Golang shipment: https://golang.org/cmd/cgo/

pairing/bn256/gfp6.go
jeffallen: It's more secure that way, because DNSSEC is not widely deployed. :)

Daeinar:
@wouterl Thanks for your comment and yes we are aware of this issue. We decided to use BN256 as a first step because Cloudflare released a very efficient implementation of BN256 recently and it was relatively easy to port to Kyber which finally allows us to use pairings in our projects. We will definitely look into more secure alternatives further down the road such as https://blog.z.cash/new-snark-curve/.

pairing/bn256/gfp.h
Bren2010: This is a macro, it is expanded in the .s file.

And none of this is cgo, there is no cgo overhead. It is assembly that the go compiler uses instead of whatever it would've generated for a pure-Go implementation of the same function. Functions are defined in the .s file.

pairing/bn256/README.md
```diff
@@ -0,0 +1,18 @@
+bn256
+-----
+
+Package bn256 implements a particular bilinear group at the 128-bit security
+level. The basis for this package is the code from
+https://github.com/cloudflare/bn256 which itself is a modification of the
```
Bren2010: Why is this copied rather than vendored? That is, what are your modifications?

pairing/bn256/constants.go
Bren2010: There are polynomials that give the trace of the frobenius, the order of the curve, and the characteristic of the field in terms of this parameter. This is how BN curves are constructed.

pairing/bn256/README.md
Daeinar: Thanks for your comments! I've copied everything over since I wasn't sure what modifications I would need to do. Now that I have a better overview, I think we can use your library directly using the wrappers I wrote. I will do the necessary tweaks next week.

ineiti:
There is already a bug-report filed by David Froelicher: #287 - I'm pretty sure it's legit.

pairing/bn256/README.md
Daeinar: @Bren2010: To answer your question on the modifications, I replaced the high-level API in [bn256.go](https://github.com/cloudflare/bn256/blob/master/bn256.go) with implementations of the Kyber interfaces [point](https://github.com/dedis/kyber/blob/pairing/pairing/bn256/point.go), [scalar](https://github.com/dedis/kyber/blob/pairing/pairing/bn256/scalar.go), [group](https://github.com/dedis/kyber/blob/pairing/pairing/bn256/group.go), and [suite](https://github.com/dedis/kyber/blob/pairing/pairing/bn256/suite.go). I did not touch the lower level files except for `gfp12.go` where I added the [neutral element of GT](https://github.com/dedis/kyber/blob/pairing/pairing/bn256/gfp12.go#L48-L77).

For now we probably need to keep a copy of your library within Kyber because we need access to several of the unexported structs and constants to implement the above interfaces. This includes `curvePoint`, `twistPoint`, `gfP12`, `curveGen`, `twistGen`, `gfP12Inf`, `gfP12Gen` and probably a few others that I have overlooked. I could write wrappers around your exported points G1, G2, GT etc. but that might become quite confusing and even then I am not sure we can access all the functionality we need. What do you think?

Bren2010: Sounds good to me. I'd recommend putting this info in the readme.

Daeinar: Good idea, will do that.

Daeinar:
@jeffallen @ineiti From my side this PR is ready now.

PR #285, opened 2018-02-22
Change blake to blake2xb

### LINKED ISSUE #282: change blake to blake2x

jeffallen:
In order to be clearer about which underlying implementation
we are using.

Fixes #282.


kc1212:
This will break cothority and other places that uses blake. Do you have fixes in other repos in the pipeline?

jeffallen:
There are no impacts outside of kyber.

PR #288, opened 2018-03-08
Fix Scalar (Un)marshaling

Daeinar:
This PR fixes the bug found by @froelich triggered by scalars being marshaled into variable-sized slices. Now scalars are marshaled to zero-padded 32-byte slices.

### DIFF
--- a/pairing/bn256/scalar.go
+++ b/pairing/bn256/scalar.go
@@ -3,6 +3,7 @@ package bn256
 import (
 	"crypto/cipher"
 	"crypto/subtle"
+	"errors"
 	"io"
 	"math/big"
 
@@ -20,8 +21,15 @@ func newScalar() kyber.Scalar {
 }
 
 func (s *scalar) Equal(a kyber.Scalar) bool {
-	x := a.(*scalar).x
-	return subtle.ConstantTimeCompare(s.x.Bytes(), x.Bytes()) == 1
+	sm, err := s.MarshalBinary()
+	if err != nil {
+		return false
+	}
+	am, err := a.MarshalBinary()
+	if err != nil {
+		return false
+	}
+	return subtle.ConstantTimeCompare(sm, am) == 1
 }
 
 func (s *scalar) Set(a kyber.Scalar) kyber.Scalar {
@@ -107,7 +115,8 @@ func (s *scalar) SetBytes(buf []byte) kyber.Scalar {
 }
 
 func (s *scalar) Bytes() []byte {
-	return s.x.Bytes()
+	sm, _ := s.MarshalBinary()
+	return sm
 }
 
 func (s *scalar) SetVarTime(varTime bool) error {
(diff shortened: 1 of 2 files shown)


pairing/bn256/suite_test.go
```diff
@@ -8,6 +8,22 @@ import (
 	"golang.org/x/crypto/bn256"
 )
 
+func TestScalarMarshal(t *testing.T) {
+	suite := NewSuiteBN256()
+	a := suite.G1().Scalar().Pick(random.New())
```
jeffallen: This test would be more "convincing" if it made sure that a ended up a value that caused len(a.Bytes()) < a.MarshalSize(). As it is, this will only exercise that code path once per 256 executions on average (i.e. ones where the MSB is zero)

pairing/bn256/scalar.go
```diff
@@ -20,8 +21,15 @@ func newScalar() kyber.Scalar {
...
-	return subtle.ConstantTimeCompare(s.x.Bytes(), x.Bytes()) == 1
+	sm, err := s.MarshalBinary()
+	if err != nil {
+		return false
+	}
+	am, err := a.MarshalBinary()
+	if err != nil {
+		return false
+	}
+	return subtle.ConstantTimeCompare(sm, am) == 1
```
kc1212: not sure if this is crucial, but having `if err != nil {` makes this not constant time, no?

pairing/bn256/scalar.go
```diff
@@ -116,8 +125,14 @@ func (s *scalar) SetVarTime(varTime bool) error {
...
 	n := s.MarshalSize()
-	buf := s.x.Bytes()
-	return buf[:n], nil
+	buf := make([]byte, n)
+	bytes := s.x.Bytes()
+	if n < len(bytes) {
+		return nil, errors.New("bn256.Scalar: unexpected size")
+	}
+	m := n - len(bytes)
+	copy(buf[m:n], bytes)
```
ineiti: Question: is `s.x.Bytes()` little- or big-endian? Here you suppose it's big-endian, no?

pairing/bn256/scalar.go
ineiti: I'm not even sure that `MarshalBinary` is constant-time. Most probably not, because it uses `Bytes()` which can return a variable number of bytes. So we might as well be honest and say it's variable time and use `bytes.Compare` ;)

pairing/bn256/suite_test.go
Daeinar: Added another test that should trigger that case more regularly.

pairing/bn256/scalar.go
Daeinar: I was sticking to the Big.Int behavior which does big-endian, see https://golang.org/pkg/math/big/#Int.Bytes.

pairing/bn256/scalar.go
Daeinar: Yeah it's redundant. I removed those ifs and let the constant time comparison deal with errors from the marshaling.

PR #289, opened 2018-03-12
BLS Signing

Daeinar:
BLS signing scheme implementation.

Merge sequence:
1. #292 
2. #291 
3. #290 
4. #289

### DIFF
--- a/sign/bls/bls.go
+++ b/sign/bls/bls.go
@@ -1,3 +1,6 @@
+// Package bls implements the Boneh-Lynn-Shacham (BLS) signature scheme which
+// was introduced in the paper "Short Signatures from the Weil Pairing". BLS
+// requires pairing-based cryptography.
 package bls
 
 import (
@@ -16,8 +19,8 @@ func NewKeyPair(suite pairing.Suite, random cipher.Stream) (kyber.Scalar, kyber.
 	return x, X
 }
 
-// Sign creates a BLS signature s = x * H(m) on a message m using the private
-// key x. The signature s is a point on curve G1.
+// Sign creates a BLS signature S = x * H(m) on a message m using the private
+// key x. The signature S is a point on curve G1.
 func Sign(suite pairing.Suite, x kyber.Scalar, msg []byte) ([]byte, error) {
 	HM := hashToPoint(suite, msg)
 	xHM := HM.Mul(x, HM)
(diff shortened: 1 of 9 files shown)


sign/bls/bls.go
```diff
@@ -0,0 +1,54 @@
...
+		return err
+	}
+	right := suite.Pair(s, suite.G2().Point().Base())
+	if !left.Equal(right) {
+		return errors.New("bls: invalid signature")
+	}
+	return nil
+}
+
+// Hash a message to a point on curve G1. XXX: This should be replaced
```
nikkolasg: Make that an issue even so we don't forget that it's here and it's bad :D

sign/bls/bls_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/dedis/kyber/pairing/bn256"
+	"github.com/dedis/kyber/util/random"
+)
+
+func TestBLS(t *testing.T) {
+	msg := []byte("Hello Boneh-Lynn-Shacham")
+	suite := bn256.NewSuite()
+	private, public := NewKeyPair(suite, random.New())
+	sig := Sign(suite, private, msg)
+	if err := Verify(suite, public, msg, sig); err != nil {
```
nikkolasg: Could you also write a test that changes the signature to random bytes and "verifies that it does not verify" ? ;)

sign/bls/bls.go
```diff
@@ -0,0 +1,54 @@
...
+// eventually by a proper hash-to-point mapping like Elligator.
+func hash(suite pairing.Suite, msg []byte) kyber.Point {
+	h := suite.Hash()
+	h.Write(msg)
+	digest := h.Sum(nil)
+	x := suite.G1().Scalar().SetBytes(digest)
+	return suite.G1().Point().Mul(x, nil)
```
ineiti: Question: Doesn't this invalidate the security proof? I always thought that the point we're chosing should not have a known `Scalar`, else the security proof fails.

sign/bls/bls_test.go
```diff
@@ -0,0 +1,29 @@
...
+	require.Nil(t, err)
+}
+
+func TestBLSFail(t *testing.T) {
+	msg := []byte("Hello Boneh-Lynn-Shacham")
+	suite := bn256.NewSuite()
+	private, public := NewKeyPair(suite, random.New())
+	sig := Sign(suite, private, msg)
+	sig[0] ^= 0x01
+	if Verify(suite, public, msg, sig) == nil {
```
ineiti: Can you also make it fail when verifying with another public key? Else we could have a signature that just holds the hash of the message, and it would pass this test.

sign/bls/bls.go
nikkolasg: It's absolutely not ideal we all agree I think but that's the easiest way to get it done. The thing is it does not return the scalar in question, it creates it and then it gets GC'd so in theory there should not be any problem (if the "Mul" operation is constant time ?).
The best solution would be to implement Elligator mappings for any elliptic curve so we can map random bytes to a point easily and safely... that's longterm...

Daeinar: Yes it is not ideal but as long as the implementation does not leak the secret scalar everything should be fine. To fix this we should either implement the `Embed` functionality that is currently missing in our `bn256` code or even better the Elligator map as @nikkolasg mentioned.

PR #290, opened 2018-03-12
Threshold BLS

Daeinar:
Implementation of the threshold-based Boneh-Lynn-Shacham signing scheme.

Merge sequence:
1. #292 
2. #291 
3. #290 
4. #289

### DIFF
--- a/sign/tbls/tbls.go
+++ b/sign/tbls/tbls.go
@@ -1,3 +1,13 @@
+// Package tbls implements the (t,n)-threshold Boneh-Lynn-Shacham signature
+// scheme. During setup a group of n participants runs a distributed key
+// generation algorithm (see kyber/share/dkg) to compute a joint public signing
+// key X and one secret key share xi for each of the n signers. To compute a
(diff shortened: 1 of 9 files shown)


sign/tbls/tbls.go
```diff
@@ -0,0 +1,97 @@
...
+
+// SigShare encodes a threshold BLS signature share s = i || v where the 2-byte
+// big-endian value i corresponds to the share's index and v represents the
+// share's value.
+type SigShare []byte
+
+// Index returns the index i of the TBLS share.
+func (s *SigShare) Index() (int, error) {
+	var index uint16
+	buf := bytes.NewReader([]byte(*s))
```
nikkolasg: You can remove the []byte() portion here. at least it builds without it for me.

Daeinar:
I have no clue what travis is trying to tell me with this error: https://travis-ci.org/dedis/kyber/builds/353254616#L508.

There is no line 37 in https://github.com/dedis/kyber/blob/tbls/sign/bls/bls_test.go. When I run the test locally everything works. Anybody an idea?

PR #291, opened 2018-03-14
bn256.Scalar to mod.Int

Daeinar:
This PR removes code duplication in bn256.Scalar by using the existing mod.Int library.

Merge sequence:
1. #292 
2. #291 
3. #290 
4. #289

PR #292, opened 2018-03-14
Remove Interface from Secret Sharing Code

Daeinar:
This PR removes the suite interface from the secret sharing code which complicated implementations at other places (e.g., the code for pairings).

Merge sequence:
1. #292 
2. #291 
3. #290 
4. #289


Daeinar:
`NewPriPoly` is a fairly low-level function and mostly used within other packages such as `pvss` or the various `dkg` implementations. It should not be used directly anyways.

PR #295, opened 2018-03-16
Relocate DSS Package

Daeinar:
Moves the DSS package from `kyber/share/dss` to `kyber/sign/dss`.


ineiti:
@Daeinar - is this still relevant?

Daeinar:
@ineiti Done

stable/script
```diff
@@ -0,0 +1,8 @@
+SRC="$1"
+DST="$2"
+
+(
+cp -a $SRC/sign/dss $DST/share/
```
jeffallen: Barf.

ineiti: All that in the name of backward-compatibility... Will go away in `v3`.

PR #296, opened 2018-03-22
Curve25519 key generation

nikirill:
This pull request adds a NewKey method to curve25519 package so generated keys are not vulnerable to subgroup attack. It also adds an omitted assignment of public key hiding which was previously resulting into the program entering an infinite loop when generating NewHidingKeyPair.

### DIFF
--- a/group/curve25519/curve.go
+++ b/group/curve25519/curve.go
@@ -4,12 +4,14 @@ package curve25519
 
 import (
 	"crypto/cipher"
+	"crypto/sha512"
 	"errors"
 	"fmt"
 	"math/big"
 
 	"github.com/dedis/kyber"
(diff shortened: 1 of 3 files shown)


kc1212:
It looks fine to me, but we're at non-breaking phase so the changes like
```diff
-type SuiteEd25519 struct {
+type SuiteCurve25519 struct { 
```
probably need to be changed back. Maybe with a comment saying that the name is not right and should be changed in future releases. What do you think @ineiti ?

@nikirill Is this PR ready for merge?

nikkolasg:
Yep I confirm the changes are good ! thanks @nikirill I guess you're the only one using that curve ^^

PR #297, opened 2018-03-26
[VSS] changing dhkey to bytes

nikkolasg:
This PR introduces a small and hopefully non-invasive change to make the Diffie Hellman key field a slice of bytes instead of point. The goal is to be able to translate the vss structs one-to-one into a protobuf packet. 
This introduces no additional marshallisation compared to what it was. It's also more consistent wrt the other fields such as signature which are all slice of bytes.

### DIFF
--- a/share/vss/pedersen/vss_test.go
+++ b/share/vss/pedersen/vss_test.go
@@ -205,7 +205,8 @@ func TestVSSVerifierDecryptDeal(t *testing.T) {
 
 	// wrong dh key
 	goodDh := encD.DHKey
-	encD.DHKey, _ = suite.Point().Null().MarshalBinary()
+	encD.DHKey, err = suite.Point().Null().MarshalBinary()
+	require.Nil(t, err)
 	decD, err = v.decryptDeal(encD)
 	assert.Error(t, err)
 	assert.Nil(t, decD)
--- a/share/vss/pedersen/vss_test.go
+++ b/share/vss/pedersen/vss_test.go
@@ -205,7 +205,7 @@ func TestVSSVerifierDecryptDeal(t *testing.T) {
 
 	// wrong dh key
 	goodDh := encD.DHKey
-	encD.DHKey = suite.Point()
+	encD.DHKey, _ = suite.Point().Null().MarshalBinary()
 	decD, err = v.decryptDeal(encD)
 	assert.Error(t, err)
 	assert.Nil(t, decD)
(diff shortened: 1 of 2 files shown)


share/vss/pedersen/vss_test.go
```diff
@@ -205,7 +205,7 @@ func TestVSSVerifierDecryptDeal(t *testing.T) {
 
 	// wrong dh key
 	goodDh := encD.DHKey
-	encD.DHKey = suite.Point()
+	encD.DHKey, _ = suite.Point().Null().MarshalBinary()
```
ineiti: always check error - just in case.

ineiti:
is this for easier interaction with javascript?

share/vss/pedersen/vss_test.go
nikkolasg: If this line fails, then I should even panic, since that means our kyber library can't even marshal a "0" point it just created, and that means super big trouble, no ?

nikkolasg:
This is not strictly necessary per se but it allows easier integration with protobuf definitions yes, now the whole set of DKG packets _transmitted over the wire_ can be expressed directly, as-is, as protobuf message definitions. I'm trying to explore ways of trying to use protobuf the right way: that the protobuf files are at the core of the code and then we move to Go structures, instead of putting go structures at the core and then derive the protobuf files. We're doing the latter from now on, and we've already seen issues with that approach because both sets (Go structs & protobufs messages) are sometimes not in sync and we did not detect / see anything wrong. It's also much more easier when starting a new projet to just go look the protobuf files to know what to expect from the network, rather than lookup individually all structs in Go, determine if they will be transmitted over the network and derive a protobuf translation manually (or by code, which may even take longer in most cases). I know for cothority we have this external folder that contains all definitions and that is already very nice, but I wanna try with drand to even go a little bit further ;)
Again this PR is not strictly necessary, but makes things nicer and easier.

ineiti:
OK - sounds nice. Then why don't you want to do the following? This would remove the `bytes` meaning different things like `Point`, `Scalar`, ..., depending on the context.

```
message Point{
  bytes Data = 1;
}

message Signature{
  bytes Data = 1;
}

message type EncryptedDeal struct {
 	// Ephemeral Diffie Hellman key
	Point DHKey = 1;
 	// Signature of the DH key by the longterm key of the dealer
 	Signature Signature = 2;
 	// Nonce used for the encryption
 	bytes Nonce = 3;
 	// AEAD encryption of the deal marshalled by protobuf
 	bytes Cipher = 4;
 }
```

nikkolasg:
After discussion, I'm all for it and will give it a try, also using DER encoding for points and scalar.
I'll share my thoughts afterwards... ;)

PR #298, opened 2018-03-26
Group information from Scalar and Point

nikkolasg:
This PR is a proposal to start bringing some self-describing functionalities to our beloved kyber library.
Namely, this introduce the `Groupable` interface which returns the `Group` an element is using. 
This goal here is to be able to safely /un/marshal `Point`s and `Scalar`s from any `Group`. For example, my drand protobuf definition is starting to look like:
```protobuf
message Point {
    uint32 oid = 1;
    bytes point = 2;
}

message Scalar {
    uint32 oid = 1;
    bytes scalar = 2;
}
```
That schema combined with the `Groupable` functionality implemented allows us to decode any (or allowed) type of `Point`s and `Scalar`s in the same "message" (which is HUGE).

For the moment it is only implemented for edwards25519 and bn256 to show the concept and discuss it. If it's  approved, it may be good to enforce that interface for all `Point`s and `Scalar`s by moving it inside the `Scalar` and `Point` definition.

--------------------------------------------------------------------------------------------------------------------------------------

See below my PoC protobuf converter that allows to pass from the above protobuf definition to our dear kyber.Point/Scalar and vice-versa:
```go
// mapping from oid to group
// note that we'll mostly likely instantiate the groups at init time
var mapping = map[uint32]kyber.Group{
	1:  edwards25519.NewBlakeSHA256Ed25519(),
	21: bn256.NewSuite().G1(),
	22: bn256.NewSuite().G2(),
	23: bn256.NewSuite().GT(),
}
// mapping from group name to oid
// best would be to directly assign oid to group but that's far too intrusive and error prone
var reverseMapping = map[string]uint32{
	edwards25519.NewBlakeSHA256Ed25519().String(): 1,
	bn256.NewSuite().G1().String():                21,
	bn256.NewSuite().G2().String():                22,
	bn256.NewSuite().GT().String():                23,
}

type ProtobufPoint = kyber_proto.Point
type ProtobufScalar = kyber_proto.Scalar

// converter from protobuf definitions to kyber.Point/Scalar and vice-versa
type protobufConverter struct{}

func (c *protobufConverter) PointFromProtobuf(p *ProtobufPoint) (kyber.Point, error) {
	group, exists := mapping[p.GetOid()]
	if !exists {
		return nil, fmt.Errorf("oid %d unknown", p.GetOid())
	}
	point := group.Point()
	return point, point.UnmarshalBinary(p.GetPoint())
}

func (c *protobufConverter) ProtobufFromPoint(p kyber.Point) (*ProtobufPoint, error) {
	desc, ok := p.(kyber.Groupable)
	if !ok {
		return nil, errors.New("given point is not self describing")
	}
	group := desc.Group()
	oid, exists := reverseMapping[group.String()]
	if !exists {
		return nil, fmt.Errorf("group %s is not registered to the protobuf mapping", group.String())
	}
	buffer, err := p.MarshalBinary()
	return &ProtobufPoint{
		Oid:   oid,
		Point: buffer,
	}, err
}

func (c *protobufConverter) ScalarFromProtobuf(p *ProtobufScalar) (kyber.Scalar, error) {
	group, exists := mapping[p.GetOid()]
	if !exists {
		return nil, fmt.Errorf("oid %d unknown")
	}
	scalar := group.Scalar()
	return scalar, scalar.UnmarshalBinary(p.GetScalar())
}

func (c *protobufConverter) ProtobufFromScalar(s kyber.Scalar) (*ProtobufScalar, error) {
	desc, ok := s.(kyber.Groupable)
	if !ok {
		return nil, errors.New("given point is not self describing")
	}
	group := desc.Group()
	oid, exists := reverseMapping[group.String()]
	if !exists {
		return nil, fmt.Errorf("group %s is not registered to the protobuf mapping", group.String())
	}
	buffer, err := s.MarshalBinary()
	return &ProtobufScalar{
		Oid:    oid,
		Scalar: buffer,
	}, err
}
```


group/edwards25519/scalar.go
```diff
@@ -24,7 +24,12 @@ import (
 // The scalars are GF(2^252 + 27742317777372353535851937790883648493).
 
 type scalar struct {
-	v [32]byte
+	v     [32]byte
```
jeffallen: This increases the size of a scalar from 32 bytes to 40 bytes (32 + one pointer). Why? It seems like for ed25519, Group() could return a pointer to a fixed kyber.Group which is shared by all instances.

(Same for point.)

pairing/bn256/group.go
```diff
@@ -73,3 +97,84 @@ func (c *common) PrimeOrder() bool {
 func (c *common) NewKey(rand cipher.Stream) kyber.Scalar {
 	return mod.NewInt64(0, Order).Pick(rand)
 }
+
+type scalarDescribing struct {
+	*mod.Int
```
jeffallen: It seems to me that if this *mod.Int was mod.Int, then you would not need all these forwarding methods.

And forwarding methods in addition to being gross to read, potentially prevent inlining...

nikkolasg: I agree but the `mod.Int` package is in ... `internal/` so not accessible from the pairing package :|
It's just a proof of concept again, so I am definitely for using `mod.Int` directly but that requires to put the `mod.Int` package on its own public package.

group/edwards25519/scalar.go
nikkolasg: No because now the suite has a `random.Stream` that can produces deterministic output when using `XOF` or `Hash()`. It's again related to that `Suite` functionality... A lot of things would be simpler if we killed that thing ;)

pairing/bn256/group.go
jeffallen: mod.Int is public as far as I can see. The only internal package is kyber/group/internal/marshalling.

ineiti:
Would this also work for `x509` encoded points if we add them to the `mapping`?

nikkolasg:
I believe so. 
We would need two things I think:
+ a `GroupID X509` that represents X509 encoded points
+ a registered a meta suite `X509Suite` in `suites/all.go` that is able to (1) decode x509 and (2) instantiate the right suite/group to decode the point.

nikkolasg:
I finally use it in the new version of drand:
+ [Here](https://github.com/dedis/drand/blob/master/protobuf/drand/client.proto#L53) is the protobuf definition where I send a point.
+ [Here](https://github.com/dedis/drand/blob/master/core/drand.go#L163) is where I decode the point

csbenz:
For reference, here are the two lines quoted by @nikkolasg above, at the correct commit at time of writing:

- [Here](https://github.com/dedis/drand/blob/73693b172b8d7c29c294a5fbc5b19b82cd91d952/core/client.go#L53) is the protobuf definition where I send a point.
- [Here](https://github.com/dedis/drand/blob/73693b172b8d7c29c294a5fbc5b19b82cd91d952/core/drand.go#L163) is where I decode the point

pairing/bn256/group.go
nikkolasg: Ok I have found back my initial reasoning. In fact, both `*mod.Int` and `mod.Int` don't work because when we use the scalar methods, such as `a.Equal(b)`, the caller `a` will try to cast `b` to a `*mod.Int` which is not working, in **both** cases. I have tried your approach and it gives me:
```go
panic: interface conversion: kyber.Scalar is *bn256.scalarDescribing, not *mod.Int [recovered]
panic: interface conversion: kyber.Scalar is *bn256.scalarDescribing, not *mod.Int
```
It turns out that putting `kyber.Scalar` at that place, it works. Seems like Go dispatch methods correctly when using embedded interfaces but is stricter when using embedded structs... Not sure I understand the full rule yet, would be happy to hear your thoughts on it.
In the meantime I'm refactoring to use `kyber.Scalar`, it costs us a few casting here and there more but it's much cleaner.

nikkolasg: So short answer: I could not find solutions using a separate struct (such as `scalarDescribing`) that alleviates us from redefining the methods. I have reproduced the issue on a [playground](https://play.golang.org/p/RrjecTlJzHa). If you can find something that works, please let me know !
So I see two potential solutions:
1. Use this solution, which I agree, does not look very nice but is working
2. Put a `Group` field in `mod.Int` which is settable via `SetGroup`. It is much more clean code-wise but much less clean logically-wise, it feels like it taints the "pure" `mod.Int` that we have so far,  and without making API breaking change, we can't garantee that there is a group associated with a `mod.Int`. So even if we cast to `kyber.Groupable`, we might end up with a null `Group` which is **not** a correct behavior in my opinion, since casting should tell you whether or not it has a `Group` inside already.

jeffallen:
Closing this because we do not expect to merge it. @nikkolasg if I've mis-understood, please re-open and tell us your proposed next steps with it. Thanks.

AnomalRoil:
Notice this is preventing `drand` in its latest version to work, since drand `protobuf/crypto/converter.go:56` and `:84` is assuming the existence of `kyber.Groupable`. @nikkolasg

ineiti:
Is it possible to use the `dedis/protobuf` 

https://github.com/dedis/protobuf/blob/12addf9e6c31a33586f95495915aef655e332992/interface.go#L52

calls? We use it now internally in the `dedis/cothority` project.

nikkolasg:
@AnomalRoil Normally go modules should be able to handle different branches. The one registered in drand points to the branch `drand`  in kyber.
@ineiti everything's possible ;) Though like I said to Jeff, (1) it takes time and this is clearly not a priority for me to change the wire-protocol and (2) I'm not a fan of enforcing a specific wire-protocol at the kyber level, even more with 8 byte (we dont expect 2^64 different identifiers/groups, aren't we ?). The drand solution works fine at the moment and drand is still failing at other places so I need to pay attention elsewhere unfortunately !

ineiti:
OK - but if you don't change the wire-protocol now, you will never do so later ;)

And the 8 bytes are for human-readability and easier debugging when you look at byte transfers. So you can more easily see what's going over the wire.

Another option would've been to use ASN.1 encodings, but they don't have any standards yet for bc256.

PR #301, opened 2018-04-05
BN256: g1 g2 gt being registered in all suites

nikkolasg:
This PR augments bn256 g1 g2 and gt with some generic suite functionality so these groups can be registered as `Suite` in the global list of suites kyber has that can be referenced by name.
This is firstly motivated by the fact that it is currently not possible to run a simulation with bn256 for the moment since nothing is registered from this package.

### DIFF
--- a/suites/all.go
+++ b/suites/all.go
@@ -2,8 +2,13 @@ package suites
 
 import (
 	"github.com/dedis/kyber/group/edwards25519"
+	"github.com/dedis/kyber/pairing/bn256"
 )
 
 func init() {
 	register(edwards25519.NewBlakeSHA256Ed25519())
+	register(bn256.NewSuite().G1().(Suite))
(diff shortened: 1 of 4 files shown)


suites/all.go
```diff
@@ -2,8 +2,12 @@ package suites
 
 import (
 	"github.com/dedis/kyber/group/edwards25519"
+	"github.com/dedis/kyber/pairing/bn256"
 )
 
 func init() {
 	register(edwards25519.NewBlakeSHA256Ed25519())
+	register(bn256.NewSuite().G1().(Suite))
```
jeffallen: Is bn256 constant time?

jeffallen:
@nikkolasg asked me to wait on merging this because he might decide to do it in a less invasive way somehow in his local repo.

ineiti:
Probably superseded by #314 - at least I built it on the code from here ;)

PR #302, opened 2018-04-10
Kopiga branch

rkopiga:
This PR introduces the ability to refresh a share without changing the distributed public key associated. In a nutshell, it enables one to run a DKG with a "0" as the distributed secret, then each participants adds up his new share to his old one. The shares are refreshed; as long as one honest servers uses the new shares, old shares are innfective and the final distributed public key stays the same.

This is the first improvement in a potential series of improvements: share renewal, addition of nodes in a DKG, deletion of nodes in a DKG, increasing/decreasing threshold etc... (keeping the same distributed public key in most cases !)

### DIFF
--- a/share/dkg/pedersen/dkg.go
+++ b/share/dkg/pedersen/dkg.go
@@ -121,6 +121,43 @@ func NewDistKeyGenerator(suite Suite, longterm kyber.Scalar, participants []kybe
 	}, nil
 }
 
+
+func ResharingDKG(suite Suite, longterm kyber.Scalar, participants []kyber.Point, t int) (*DistKeyGenerator, error){
+	pub := suite.Point().Mul(longterm, nil)
+	// find our index
+	var found bool
+	var index uint32
+	for i, p := range participants {
+		if p.Equal(pub) {
+			found = true
+			index = uint32(i)
+			break
+		}
+	}
+	if !found {
+		return nil, errors.New("dkg: own public key not found in list of participants")
+	}
+	var err error
+	// generate our dealer / deal
+	ownSec := suite.Scalar().Zero() //THIS HAS TO BE 0
+	dealer, err := vss.NewDealer(suite, longterm, ownSec, participants, t)
+	if err != nil {
+		return nil, err
+	}
+
+	return &DistKeyGenerator{
+		dealer:       dealer,
+		verifiers:    make(map[uint32]*vss.Verifier),
+		t:            t,
+		suite:        suite,
+		long:         longterm,
+		pub:          pub,
+		participants: participants,
+		index:        index,
+	}, nil
+}
+
+
 // Deals returns all the deals that must be broadcasted to all
 // participants. The deal corresponding to this DKG is already added
 // to this DKG and is ommitted from the returned map. To know which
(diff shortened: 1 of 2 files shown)


share/dkg/pedersen/dkg.go
```diff
@@ -121,6 +121,43 @@ func NewDistKeyGenerator(suite Suite, longterm kyber.Scalar, participants []kybe
 	}, nil
 }
 
+
+func ResharingDKG(suite Suite, longterm kyber.Scalar, participants []kyber.Point, t int) (*DistKeyGenerator, error){
```
nikkolasg: I would prefer to have one `NewDistKeyGenerator` and one `NewDistKeyGeneratorWithSecret` (for example) so `NewDistKeyGenerator` only calls `NewDistKeyGeneratorWithSecret` with a random secret, so the code is not repeated.
Plus that will serve as a basis for future extensions like increasing/decreasing threshold etc..

share/dkg/pedersen/dkg.go
```diff
@@ -343,6 +380,32 @@ func (d *DistKeyGenerator) DistKeyShare() (*DistKeyShare, error) {
 	}, nil
 }
 
+
+func (f *DistKeyShare) Renew(g *DistKeyShare, suite Suite) (*DistKeyShare, error){
```
nikkolasg: Please put the suite as the first argument. It seems an annoying request but it's much much easier when the code is consistent with the rest of the repo. :)

share/dkg/pedersen/dkg.go
```diff
@@ -343,6 +380,32 @@ func (d *DistKeyGenerator) DistKeyShare() (*DistKeyShare, error) {
...
+	//Check G(0) = 0*G.
+	if !g.Public().Equal(suite.Point().Base().Mul(suite.Scalar().Zero(),nil)){
+		return nil, errors.New("G(0) != 0*G, wrong renewal function!")
```
nikkolasg: I think go lint will tell you but avoid "!" at the end of your errors, stick to simple error sentences that can be embedded in larger error report etc...

share/dkg/pedersen/dkg.go
nikkolasg: Also you need to write some comments on this function explaining from a higher level point of view what it is doing.

share/dkg/pedersen/dkg_test.go
```diff
@@ -310,6 +311,80 @@ func dkgGen() []*DistKeyGenerator {
...
+
+	shares := make([]*share.PriShare, nbParticipants)
+	for i, dks := range dkss {
+		shares[i] = dks.Share
+	}
+	secret, err := share.RecoverSecret(suite, shares, nbParticipants, nbParticipants) // SUMf_j(0),j:0->#participants-
+	assert.Nil(t, err)
+
+	commitSecret := suite.Point().Mul(secret, nil)
+	assert.Equal(t, dkss[0].Public().String(), commitSecret.String())
```
nikkolasg: We also want to check if the first public key generated during the first DKG stays the same ! Here if I'm not mistaken you are only checking if the newly shares, reconstruct to a secret that when commited (i.e. `s * G`) gives the same public point that is generated from a renewed share... I wanna know if it holds with th public point generated from the original shares !

PR #304, opened 2018-05-02
better cosi error

ineiti:
Have cosi say why a signature is invalid. This helps for debugging...

### DIFF
--- a/sign/cosi/cosi.go
+++ b/sign/cosi/cosi.go
@@ -209,8 +209,11 @@ func Verify(suite Suite, publics []kyber.Point, message, sig []byte, policy Poli
 	sB := suite.Point().Mul(r, nil)
 	left := suite.Point().Add(kA, sB)
 
-	if !left.Equal(V) || !policy.Check(mask) {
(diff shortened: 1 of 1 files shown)


nikkolasg:
I'm saying that because I can't refrain but it's usually a good idea to hide the exact reason why a cryptographic protocol fails because it can sometimes lead to an exploit (already happened).
On top of that it should preferably be constant time (can we detect what is the error depending on the total time, if there is one "if" per verification clause?) but that's another issue.
Anyway, it's probably not very important here but to be considered on the long run...

kc1212:
Yes, a classical one is the padding oracle attack. But I can't answer whether the proposed change has a security impact.

jeffallen:
I gave a thumb up to @nikkolasg 's point, I really believe it. But I can also see that there are two checks here: one is cryptographic, and we give no extra info. And one is policy. So we help an attacker who is trying to twiddle bits that are trying to squeak through the permitted policy. So, is that scary or not? Hard to tell. @ineiti can you explain why that's not scary?

PR #306, opened 2018-05-09
new stable from master

ineiti:
Latest changes included from `make create_stable`


nikkolasg:
Mhhh not really but they can be. Let me try to do a quick PR against master today that will keep the signature of `NewDistKeyGenerator` the same as before. IF you need backward-compatibility.

ineiti:
Waiting for merge of #307

PR #309, opened 2018-05-14
Changed marshalBinary to be thread safe

JoaoAndreSa:
Dear dedis friends,

I discussed this morning with Linus and Kelong and to prevent some race conditions (for example when marshalling the roster public key with parallel sends) we changed the MarshalBinary() to be thread safe by using a copy of the point instead of the point itself. 

@Daeinar please let me know if you agree with this change :)

### DIFF
--- a/pairing/bn256/point.go
+++ b/pairing/bn256/point.go
@@ -108,6 +108,8 @@ func (p *pointG1) Mul(s kyber.Scalar, q kyber.Point) kyber.Point {
 
 func (p *pointG1) MarshalBinary() ([]byte, error) {
 	n := p.ElementSize()
+	// Take a copy so that p is not written to, so calls to MarshalBinary
+	// are threadsafe.
 	pgtemp := *p.g
 	pgtemp.MakeAffine()
 	ret := make([]byte, p.MarshalSize())
--- a/pairing/bn256/point.go
+++ b/pairing/bn256/point.go
@@ -108,15 +108,16 @@ func (p *pointG1) Mul(s kyber.Scalar, q kyber.Point) kyber.Point {
 
 func (p *pointG1) MarshalBinary() ([]byte, error) {
 	n := p.ElementSize()
-	p.g.MakeAffine()
+	pgtemp := *p.g
+	pgtemp.MakeAffine()
 	ret := make([]byte, p.MarshalSize())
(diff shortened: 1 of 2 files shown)


nikkolasg:
An alternative I can see would be to make the point affine at creation time, so there would no mutation during the lifetime of the point.
After each operation, we would transform it in affine coordinates, if needed. 
The pro is that it obviously remove the need for a "copy" of the point for the marshalization. The con is that it may require more computations at each operations. 
I'm merely saying that an altenrative exists, but I'm not sure which one's best though ;)

pairing/bn256/point.go
```diff
@@ -108,15 +108,16 @@ func (p *pointG1) Mul(s kyber.Scalar, q kyber.Point) kyber.Point {
 
 func (p *pointG1) MarshalBinary() ([]byte, error) {
 	n := p.ElementSize()
-	p.g.MakeAffine()
+	pgtemp := *p.g
```
jeffallen: There needs to be a comment here explaining why we take a copy. I propose:
```
// Take a copy so that p is not written to, so calls to MarshalBinary
// are threadsafe.

jeffallen: The real problem here is that there's an undocumented assumption that MarshalBinary is read only. This change brings bn256 back in line with the undocumented assumption, so that onet is no longer getting a data race.

It is not clear where/how to document the assumption, since MarshalBinary is an interface we inherit from Go itself. (And Go's definition is not even clear on this point: https://golang.org/pkg/encoding/#BinaryMarshaler )

ineiti:
@nikkolasg - as far as I understand, we use non-x,y representation to speed up operations. I'm pretty sure that we lose the speed up if we call `MakeAffine` every time. So I don't think this is a good option.

pairing/bn256/point.go
ineiti: Yes, correct! We could put it here:

https://github.com/dedis/kyber/blob/master/encoding.go#L13

And add a comment in there stating that the implementers of `Marshaling` should make it thread-safe.

Bad joke: let's add `Promises` to go - I miss them from javascript. They make life so much easier.

JoaoAndreSa:
Just added the proposed comments @jeffallen. As for the documentation I leave it for you guys to decide what is best :P.

PR #310, opened 2018-05-16
Elliptic Curve Integrated Encryption Scheme

Daeinar:
This PR implements ECIES, see https://en.wikipedia.org/wiki/Integrated_Encryption_Scheme.

We will need this for the new version of drand (https://github.com/dedis/drand/).

### DIFF
--- a/encrypt/ecies/ecies.go
+++ b/encrypt/ecies/ecies.go
@@ -13,16 +13,16 @@ import (
 	"golang.org/x/crypto/hkdf"
 )
 
-// Encrypt first performs a DH key exchange, then HKDF-derives a shared secret
-// key (and nonce) and finally encrypts the given message using AES-GCM.
-// Encrypt returns the ephemeral elliptic curve point of the DH key exchange
-// and the ciphertext or an error.
+// Encrypt first performs a DH key exchange using the given public key, then
+// HKDF-derives a shared secret key (and nonce) from that, and finally encrypts
+// the given message using AES-GCM. Encrypt returns the ephemeral elliptic
+// curve point of the DH key exchange and the ciphertext or an error.
 func Encrypt(group kyber.Group, public kyber.Point, message []byte, hash func() hash.Hash) ([]byte, []byte, error) {
 	if hash == nil {
 		hash = sha256.New
 	}
 
-	// Generate an ephemeral elliptic curve point
+	// Generate an ephemeral elliptic curve scalar and point
 	r := group.Scalar().Pick(random.New())
 	R := group.Point().Mul(r, nil)
 	Rb, err := R.MarshalBinary()
@@ -59,9 +59,9 @@ func Encrypt(group kyber.Group, public kyber.Point, message []byte, hash func() 
 }
 
 // Decrypt first performs a DH key exchange using the received ephemeral
-// elliptic curve point, then HKDF-derives a shared secret key (and nonce) and
-// finally decrypts the given ciphertext using AES-GCM. Decrypt returns the
-// plaintext message or an error.
+// elliptic curve point, then HKDF-derives a shared secret key (and nonce) from
+// that, and finally decrypts the given ciphertext using AES-GCM. Decrypt
+// returns the plaintext message or an error.
 func Decrypt(group kyber.Group, private kyber.Scalar, dhPoint []byte, ciphertext []byte, hash func() hash.Hash) ([]byte, error) {
 	if hash == nil {
 		hash = sha256.New
--- a/encrypt/ecies/ecies.go
+++ b/encrypt/ecies/ecies.go
@@ -2,6 +2,7 @@
 package ecies
 
 import (
+	"bytes"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/sha256"
@@ -16,9 +17,10 @@ import (
 // Encrypt first computes a shared DH key using the given public key, then
 // HKDF-derives a symmetric key (and nonce) from that, and finally uses these
 // values to encrypt the given message via AES-GCM. If the hash input parameter
-// is nil then SHA256 is used as a default. Encrypt returns the ephemeral
-// elliptic curve point of the DH key exchange and the ciphertext or an error.
-func Encrypt(group kyber.Group, public kyber.Point, message []byte, hash func() hash.Hash) ([]byte, []byte, error) {
+// is nil then SHA256 is used as a default. Encrypt returns a byte slice
+// containing the ephemeral elliptic curve point of the DH key exchange and the
+// ciphertext or an error.
+func Encrypt(group kyber.Group, public kyber.Point, message []byte, hash func() hash.Hash) ([]byte, error) {
 	if hash == nil {
 		hash = sha256.New
 	}
@@ -26,10 +28,6 @@ func Encrypt(group kyber.Group, public kyber.Point, message []byte, hash func() 
 	// Generate an ephemeral elliptic curve scalar and point
 	r := group.Scalar().Pick(random.New())
 	R := group.Point().Mul(r, nil)
-	Rb, err := R.MarshalBinary()
-	if err != nil {
-		return nil, nil, err
-	}
 
 	// Compute shared DH key
 	dh := group.Point().Mul(r, public)
@@ -41,7 +39,7 @@ func Encrypt(group kyber.Group, public kyber.Point, message []byte, hash func() 
 	len := 32 + 12
 	buf, err := deriveKey(hash, dh, len)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 	key := buf[:32]
 	nonce := buf[32:len]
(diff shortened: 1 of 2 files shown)


encrypt/ecies/ecies.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, err
+	}
+	hkdf := hkdf.New(hash, dhb, nil, nil)
+	key := make([]byte, len, len)
+	n, err := hkdf.Read(key)
+	if err != nil {
+		return nil, err
+	}
+	if n < len {
+		return nil, errors.New("ECIES: HDKF-derived key too short")
```
nikkolasg: nitpick: could you put that in lowercase please ;) ? It's always better when you don't see errors yelling at your face on your screen :D

encrypt/ecies/ecies_test.go
```diff
@@ -0,0 +1,25 @@
...
+func TestECIES(t *testing.T) {
+	message := []byte("shake that cipher")
+	suite := bn256.NewSuite()
+	private := suite.G2().Scalar().Pick(random.New())
+	public := suite.G2().Point().Mul(private, nil)
+
+	ephKey, ciphertext, err := Encrypt(suite.G2(), suite.Hash, public, message)
+	require.Nil(t, err)
+
+	plaintext, err := Decrypt(suite.G2(), suite.Hash, private, ephKey, ciphertext)
```
nikkolasg: Just because we can: could you also try to make it fail with (1) an invalid DH point and (2) with an invalid ciphertext?

encrypt/ecies/ecies.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/dedis/kyber"
+	"github.com/dedis/kyber/util/random"
+	"golang.org/x/crypto/hkdf"
+)
+
+// Encrypt first performs a DH key exchange, then HKDF-derives a shared secret
+// key (and nonce) and finally AEAD encrypts the given message using AES-GCM.
+// Encrypt returns the ephemeral elliptic curve point of the DH key exchange
+// and the ciphertext or an error.
+func Encrypt(group kyber.Group, hash func() hash.Hash, public kyber.Point, message []byte) ([]byte, []byte, error) {
```
nikkolasg: I'm wondering if the new hash function could maybe be at the end as an optional argument where it defaults to `sha256.New`. What do you think ?

encrypt/ecies/ecies.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	ciphertext := aesgcm.Seal(nil, nonce, message, nil)
+	return Rb, ciphertext, nil
+}
+
+// Decrypt first performs a DH key exchange using the received ephemeral
+// elliptic curve point, then HKDF-derives a shared secret key (and nonce) and
+// finally AEAD decrypts the given ciphertext using AES-GCM. Decrypt returns
+// the plaintext message or an error.
+func Decrypt(group kyber.Group, hash func() hash.Hash, private kyber.Scalar, Rb []byte, ciphertext []byte) ([]byte, error) {
```
nikkolasg: It's not super clear what is the `Rb` argument here, what the caller should pass. Maybe call it `dhPoint` and/or explicitely say what it is in the comment?

encrypt/ecies/ecies.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/dedis/kyber"
+	"github.com/dedis/kyber/util/random"
+	"golang.org/x/crypto/hkdf"
+)
+
+// Encrypt first performs a DH key exchange using the given public key, then
+// HKDF-derives a shared secret key (and nonce) from that, and finally encrypts
+// the given message using AES-GCM. Encrypt returns the ephemeral elliptic
+// curve point of the DH key exchange and the ciphertext or an error.
```
nikkolasg: last nitpick: now we need to precise that the last argument can be nil and it defaults to sha256 ;)

encrypt/ecies/ecies.go
```diff
@@ -0,0 +1,113 @@
...
+func Encrypt(group kyber.Group, public kyber.Point, message []byte, hash func() hash.Hash) ([]byte, []byte, error) {
```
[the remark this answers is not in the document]
Daeinar: If we would use a `Suite` here, then we could not use ECIES with pairings. In `drand` we want to compute encryptions using keys from group `G2`. This comes again from the problem that the `Suite` concept is too inflexible, unfortunately. We had similar issues with the secret sharing code a while ago, see: https://github.com/dedis/kyber/commit/ae007179f02288ef08d3a9f2794d16baa06ff637

What we definitely can do is change `hash func() hash.Hash` to `hash kyber.HashFactory`.

@ineiti What do you think?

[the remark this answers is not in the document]
Daeinar: Thanks, I see what you mean but I am not sure about the advantage in comparison to simply passing group and hash as separate arguments.

encrypt/ecies/ecies.go
Daeinar: Done.

encrypt/ecies/ecies.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/dedis/kyber"
+	"github.com/dedis/kyber/util/random"
+	"golang.org/x/crypto/hkdf"
+)
+
+// Encrypt first computes a shared DH key using the given public key, then
+// HKDF-derives a symmetric key (and nonce) from that, and finally uses these
+// values to encrypt the given message via AES-GCM. If the hash input parameter
+// is nil then SHA256 is used as a default. Encrypt returns the ephemeral
+// elliptic curve point of the DH key exchange and the ciphertext or an error.
```
jeffallen: The output is a point and a ciphertext, or an error. So why is the signature not (eph kyber,Point, ctx []byte, err error)?

I think I can see why: because the ephemeral point will be sent across the wire anyway, and we've already marshalled it into Rb, so why throw that work away and make the caller repeat it?

But then why not return one sincel []byte, which is ready to be sent across the wire? Why would the caller be interested in having the Rb separate from the ciphertext?

This is more curiosity on my part, I don't think it needs to be changed from how it is, really.

If this was in the critical path for performance, it would need to be reworked for the API to allow reuse of existing buffers in order to create less garbage. If you think that's important ask and I can point out how.

nikkolasg: I also asked my self this question if we should just return a pure slice of bytes directly, in the spirit of other implementation that we have in kyber (such as tlbs and bls).

I can't find an appropriate use of this Rb alone. It's maybe important to note that here, this ECIES scheme only "authenticates" the receiver, the sender does not use any private information. Doing "mutual authentication" in the future is possible by adding a or multiple signatures from the sender. One signature could be over this Rb, but it's only one choice amongst other possibilities that seem equivalently secure. 
So for the sake of clarity, I'm also supporting the change to just return one slice of byte containing everything.

encrypt/ecies/ecies_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+
+	"github.com/dedis/kyber/group/edwards25519"
+	"github.com/dedis/kyber/util/random"
+	"github.com/stretchr/testify/require"
+)
+
+func TestECIES(t *testing.T) {
```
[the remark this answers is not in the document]
jeffallen: It seems like the way IEEE 1609.2 uses ECIES is not quite like this implementation. I was in contact with an IEEE 1609.2 guy (William Whyte) and the test vectors he recommended need an additional input and create an additional "authentication tag" output, which your implementation includes inside of the AEAD's output, I think.

So, umm, no test vectors, I guess?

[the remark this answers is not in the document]
nikkolasg: Well if you have test vectors for ed25519, we'd happily take them ! 
But you bring up an interesting point: authentication tag.
We could allow a client to include some specific tag. Upon decryption, we'd need to add a function to retrieve the tag from the "slice of bytes output" so the client knows how/where to decrypt this packet, and then the client would call the decryption routine accordingly.
Tags can be useful to indicate which group / algo have been used to compute this ECIES encryption, for example.

This is only a suggestion, and by no mean we HAVE to implement this, but I'll be happy to hear your thoughts on it.

encrypt/ecies/ecies.go
Daeinar: Thanks for all your comments. I agree that there seems to be no obvious advantage to keep the ephemeral point and ciphertext separate, I will change that so that the function just returns a single byte slice.

kc1212:
I think a lot of our kyber code can benefit from property based testing, i.e. quickcheck https://golang.org/pkg/testing/quick 
I'm not saying to add it to this PR now, just something to keep in mind.

encrypt/ecies/ecies.go
jeffallen: So I understand you are going to change this still, so I'm moving back to WIP.

Daeinar:
@jeffallen From my side this is ready to merge.

PR #314, opened 2018-06-06
Bn256 register 2

### LINKED ISSUE #305: Integrate quick fixes in pairing
Hello,
We are currently using the pairing suite from kyber. In order to use UnLynx with either ed25519 or bn256, we had to do some changes and we wanted to know if these changes could be integrated in kyber or if there is a better way to do some of them. Here is the list of our quick fixes:
1. The function String() in dedis/kyber/pairing/bn256/curve.go has to be changed to:
`func (c *curvePoint) String() string {`
	`c.MakeAffine()`
	`x, y := &gfP{}, &gfP{}`
	`montDecode(x, &c.x)`
	`montDecode(y, &c.y)`
	`return x.String() + y.String()`
`}`

2. The function String() in dedis/kyber/pairing/point.go has to be changed to:
`func (p *pointG1) String() string {`
	`return p.g.String()`
`}`

--> the parentheses and the "," that were used in the first function and the "bn..." used in the second made all the simulations fail with this error:
`E : (network.(*Router).Start.func1: 104) - receive server identity failed: Error while receiving ServerIdentity during negotiation Unknown Error`
as the identity contains the public key which is written with the String functions.

3. Above the function New() in dedis/kyber/pairing/bn256/suite.go we changed this:
`// Not used other than for reflect.TypeOf()`
`var aScalar mod.Int`
`var aPointG1 pointG1`
`var aPointG2 pointG2`
`var aPointGT pointGT`

to this:
`// Not used other than for reflect.TypeOf()`
`var aScalar kyber.Scalar`
`var aPointG1 kyber.Point`
`var aPointG2 kyber.Point`
`var aPointGT kyber.Point`

otherwise we got this error:
`panic: unsupported null pointer type: kyber.Scalar

goroutine 189 [running]:
github.com/dedis/fixbuf.(*decoder).value(0xc420061428, 0x14dad20, 0xc4200bd4a0, 0x194, 0x3, 0xc4200bd4a0, 0x194)
	/Users/froelich/Documents/Go/src/github.com/dedis/fixbuf/encoding.go:106 +0x9ea
github.com/dedis/fixbuf.(*decoder).value(0xc420061428, 0x144d360, 0xc42043f980, 0x97, 0x2, 0x97, 0xc42043f9e0)
	/Users/froelich/Documents/Go/src/github.com/dedis/fixbuf/encoding.go:134 +0x19a
github.com/dedis/fixbuf.(*decoder).value(0xc420061428, 0x1474f20, 0xc42013b6e0, 0x194, 0x1, 0xc42013b6e0, 0x194)
	/Users/froelich/Documents/Go/src/github.com/dedis/fixbuf/encoding.go:116 +0x5c0
github.com/dedis/fixbuf.(*decoder).value(0xc420061428, 0x144e560, 0xc42043f9e0, 0x97, 0x0, 0x0, 0xc4202b1448)
	/Users/froelich/Documents/Go/src/github.com/dedis/fixbuf/encoding.go:134 +0x19a
github.com/dedis/fixbuf.BinaryEncoding.Read(0x156ef20, 0xc420092fc0, 0x144e560, 0x156ef40, 0xc420162a80, 0xc4200614a8, 0x1, 0x1, 0x0, 0xc4202b1550)
	/Users/froelich/Documents/Go/src/github.com/dedis/fixbuf/encoding.go:79 +0xdc
github.com/dedis/fixbuf.Read(0x156ef40, 0xc420162a80, 0x156ef20, 0xc420092fc0, 0xc42013b6e0, 0x6, 0x6, 0xc4202b15d0, 0x10125f9)
	/Users/froelich/Documents/Go/src/github.com/dedis/fixbuf/encoding.go:237 +0x112
github.com/dedis/kyber/pairing/bn256.(*Suite).Read(0xc420092fc0, 0x156ef40, 0xc420162a80, 0xc42013b6e0, 0x6, 0x6, 0xc42043f9c0, 0x0)
	/Users/froelich/Documents/Go/src/github.com/dedis/kyber/pairing/bn256/suite.go:87 +0x70
github.com/dedis/kyber/proof.(*hashProver).PriRand(0xc4202ee100, 0xc42013b6e0, 0x6, 0x6, 0xa, 0xa)
	/Users/froelich/Documents/Go/src/github.com/dedis/kyber/proof/hash.go:68 +0x73
github.com/dedis/kyber/shuffle.(*PairShuffle).Prove(0xc4201fd6c0, 0xc4203f44b0, 0xa, 0xa, 0x0, 0x0, 0x1576de0, 0xc4206a2390, 0xc4200bc000, 0xa, ...)
	/Users/froelich/Documents/Go/src/github.com/dedis/kyber/shuffle/pair.go:151 +0x47d
github.com/lca1/unlynx/lib.shuffleProofCreation.func2(0x1572780, 0xc4202ee100, 0x1517dc5, 0xb)
	/Users/froelich/Documents/Go/src/github.com/lca1/unlynx/lib/proofs.go:508 +0x133
github.com/dedis/kyber/proof.HashProve(0x1e76b30, 0xc4201005c0, 0x1517dc5, 0xb, 0xc4201bc000, 0x1, 0xc4200bc000, 0xa, 0xa, 0xc420246460)
	/Users/froelich/Documents/Go/src/github.com/dedis/kyber/proof/hash.go:138 +0x71
github.com/lca1/unlynx/lib.shuffleProofCreation(0xc4200c0480, 0xa, 0x10, 0xc42041c000, 0xa, 0xa, 0xc4201241e0, 0xa, 0xa, 0xc4203f44b0, ...)
	/Users/froelich/Documents/Go/src/github.com/lca1/unlynx/lib/proofs.go:511 +0x671
github.com/lca1/unlynx/lib.ShufflingProofCreation(0xc4200c0480, 0xa, 0x10, 0xc42041c000, 0xa, 0xa, 0x0, 0x0, 0x1576de0, 0xc4206a2390, ...)
	/Users/froelich/Documents/Go/src/github.com/lca1/unlynx/lib/proofs.go:520 +0x117
github.com/lca1/unlynx/protocols.(*ShufflingProtocol).Start(0xc420248460, 0x0, 0x0)
	/Users/froelich/Documents/Go/src/github.com/lca1/unlynx/protocols/shuffling_protocol.go:172 +0xaa5
created by github.com/lca1/unlynx/services/lemal.(*ServiceLeMal).StartProtocol
	/Users/froelich/Documents/Go/src/github.com/lca1/unlynx/services/lemal/service.go:348 +0x2d8`

4. This last change should probably not be integrated and we did it to be able to do: `suites.MustFind("bn256.g1")` in order to change from one curve to the other easily.
To do that we changed dedis/kyber/suites/all.go to:

`package suites

import (
"github.com/dedis/kyber/group/edwards25519"
"github.com/dedis/kyber/pairing"
"github.com/dedis/kyber/pairing/bn256"
"github.com/dedis/kyber"
)
// ChimeraSuite is the fancy name of the suite create from the pairing curve of kyber
type ChimeraSuite struct {
	pairing.Suite
	kyber.Group
}

// SuiTe is the instantiation of the suite
var    SuiTe = ChimeraSuite{
	Suite: bn256.NewSuiteBN256(),
	Group: bn256.NewSuiteBN256().G1(),
}

func init() {
	register(edwards25519.NewBlakeSHA256Ed25519())
	register(SuiTe)
}`

ineiti:
This work builds on #301, supposing bc256 is non-constant time. It also includes the changes needed by David et al described in #305. Finally it makes it possible that dedis/onet#431 can set the suite also for pairing-based crypto in localhost, mininet, and deterlab.

Comments and all very appreciated, as probably one of our students will need this tomorrow ;)

Closes #301
Closes #305


pairing/bn256/suite.go
```diff
@@ -16,29 +16,64 @@ import (
...
+	s := &Suite{commonSuite: &commonSuite{}}
+	s.g1 = &groupG1{commonSuite: s.commonSuite}
+	s.g2 = &groupG2{commonSuite: s.commonSuite}
+	s.gt = &groupGT{commonSuite: s.commonSuite}
 	return s
 }
 
+// SuiteCombined is for convenience, to define a pairing based crypto suite
+// that is at the same time usable as a group.
+type SuiteCombined struct {
```
nikkolasg: I'd probably make that non-exported if possible.

suites/all.go
```diff
@@ -2,8 +2,13 @@ package suites
 
 import (
 	"github.com/dedis/kyber/group/edwards25519"
+	"github.com/dedis/kyber/pairing/bn256"
 )
 
 func init() {
 	register(edwards25519.NewBlakeSHA256Ed25519())
+	register(bn256.NewSuite().G1().(Suite))
```
nikkolasg: These should not be registerd in all.go because they are not constant time.

(the question of whether we should unify both is something maybe we should ask ourselves somewhere else, and I'm thinking that it'd be best if we can unify both and have a way to extract information out of a group, for example if it is constant time or not).

suites/all_vartime.go
```diff
@@ -5,11 +5,18 @@ package suites
...
 
 func init() {
 	register(curve25519.NewBlakeSHA256Curve25519(false))
 	register(curve25519.NewBlakeSHA256Curve25519(true))
 	register(nist.NewBlakeSHA256P256())
 	register(nist.NewBlakeSHA256QR512())
+	register(bn256.NewSuiteG1())
+	register(bn256.NewSuiteG2())
+	register(bn256.NewSuiteGT())
+	register(bn256.NewSuite().G1().(Suite))
```
nikkolasg: What's the point of registering twice every group ?

pairing/bn256/curve.go
```diff
@@ -25,7 +25,7 @@ func (c *curvePoint) String() string {
 	x, y := &gfP{}, &gfP{}
 	montDecode(x, &c.x)
 	montDecode(y, &c.y)
-	return "(" + x.String() + ", " + y.String() + ")"
```
nikkolasg: I don't have a strong objection against that (and the change in `String`) but it is important to note that this is a **hack** because of the way **onet** works with suite group and encodings. It has nothing to do with kyber. 
So we should not forget that we'll probably have another error of this type in the future, and onet is the place to look for a fix !
I would have preferred a way to fix onet rather than hack kyber but this is clearly something reasonable so I'm fine  with it :)

suites/all_vartime.go
ineiti: one registration is the 'clean' `G1`, another registration is the `bn256`-suite with a `G1` `group` in it. Two examples:
```
s := suites.MustFind("bn256.g1")
// This does not work
private, public := bls.NewKeyPair(s, random.New())
```
```
s := suites.MustFind("combined:bn256.g1")
// This works
private, public := bls.NewKeyPair(s, random.New())
```

Depending on the use case you can have one or the other. Even though, only the `NewSuiteG2` might make sense. @JoaoAndreSa is using `G1` as a suite for the cothority - but that needs perhaps to be changed.

ineiti:
WRT the changes for `reflect.TypeOf()`, it works without them. So I suppose it's not needed.

suites/all_vartime.go
nikkolasg: What's the reason the first example does not work ? I thought the whole point of making G1 into a suit was to make it work as such ... ><

pairing/bn256/suite.go
ineiti: I'm not sure - it might be used in other places outside of bn256.

pairing/bn256/curve.go
ineiti: you're right - I had the same feeling. Checking when putting it back, it doesn't crash onet anymore!?! So perhaps this was only needed for an old version. I'll check with @JoaoAndreSa

suites/all_vartime.go
ineiti: The first example creates a `Suite` with only a `Group` of `G1` - this is of course not compatible with `pairing.Suite`.

ineiti: Anyway - most of this PR is due to dedis/onet#432...

pairing/bn256/curve.go
ineiti: OK, not needed anymore. I put back the 'original' strings.

ineiti:
I did change the following line in the end:
```
var aScalar kyber.Scalar
```
but not the different point-types, which makes sense to me. But I hope for a better explanation from @nikkolasg !

suites/all_vartime.go
nikkolasg: So ok i see the problem. So that's one solution. 
Another solution could be to juste make PairingSuite returns G1 and G2 and GT as Suite ?
But that''ll change too much code and is uglier because it even puts more thing into the Suite.. 
Ok for solution 1 ;)
aaarrgggggg I am starting to be allergic to these suites thing !!

nikkolasg:
I haven't had the time to dig into that unfortunately... It was only up to me, I'd rip apart all suites and put all encoding code in a separate repo or package so we never have to deal with these strange things ever again... :D

ineiti:
OK, this is yet another iteration - putting it back in WIP ;)

What I changed from the last time: no `SuiteCombined` anymore, just using `commonSuite` to carry an eventual `Group` in the `bn256.Suite`.

Question: what should be the _main_ group of a bn256 suite where we can happily create keypairs? Should it be always `G2`? Or are there cases where `G1`, or even `GT` would make sense?

If it should always be `G2`, then we can remove `NewSuiteG1`... and just put `G2` as the `Group` in `commonSuite`. Then everybody should be happily everafter.

Waiting for @nikkolasg and @jeffallen to confirm / infirm.

suites/all_vartime.go
ineiti: As I wrote above, it's mostly due to dedis/onet#432 and how people want to use it in onet/cothority. Thinking it through, most probably the best place to add it then is onet or cothority. Now the changes David needed to do is the part in Marshaling, that is already merged. I'll have a look if I can take it to `cothority/suite.go`.

ineiti:
Closing this in favour of #315

PR #316, opened 2018-06-13
Publish proof and shuffle as stable.

jeffallen:
Package shuffle is currently used in PriFi and e-voting, so
they must be stable enough.

Shuffle depends on proof, so it must also be stable.

Also: Slight improvements to proof docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,4 @@ entry_tunnel
 exit_tunnel
 .DS_Store
 *.cov
-
+.idea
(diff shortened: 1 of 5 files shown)


ineiti:
Putting it into WorkInProgress till we have a sign-off on this.

ineiti:
Bryan said he wants to have a code-review on this first. @Daeinar , @nikkolasg , and @lbarman volunteered to do so. But probably only on the 18th of June.

ineiti:
kyber/shuffle/pair.go:240, "mismatched vector lengths".

jeffallen:
A new Kyber is coming in Feb, we might publish it to that.

PR #317, opened 2018-07-04
Dkg private polynomial access

nikkolasg:
This PR allows the users of DKG and VSS to retrieve the private polynomial used for generating the shares.
Previously, we did not store it for DKG as we had no use for it, it was only useful for generating the shares. It turns out we need this information to be able to dynamically modify the group structure, like adding a node for example (without changing the distributed public key counterpart).
This is quite needed for drand, where such a feature is highly desirable !


ineiti:
LGTM.

share/dkg/pedersen/dkg.go
```diff
@@ -18,8 +18,12 @@ type Suite vss.Suite
 type DistKeyShare struct {
 	// Coefficients of the public polynomial holding the public key
 	Commits []kyber.Point
-	// Share of the distributed secret
+	// Share of the distributed secret which is a private information
 	Share *share.PriShare
+	// coefficients of the private polynomial generated by the node holding the
```
jeffallen: s/coef/Coef.

share/dkg/pedersen/dkg.go
```diff
@@ -18,8 +18,12 @@ type Suite vss.Suite
 type DistKeyShare struct {
 	// Coefficients of the public polynomial holding the public key
 	Commits []kyber.Point
-	// Share of the distributed secret
+	// Share of the distributed secret which is a private information
```
jeffallen: Remove the "a": which is private information

End complete sentences with a period. (Also for Commits.)

PR #318, opened 2018-07-19
Don't panic in group tests

cag:
This injects `t *testing.T` into the `GroupTest`s and uses it to allow tests to continue after individual failures, and for tests to report extra info on specific failure cases.

### DIFF
--- a/util/test/test.go
+++ b/util/test/test.go
@@ -3,6 +3,7 @@ package test
 import (
 	"bytes"
 	"crypto/cipher"
+	"testing"
 
 	"github.com/dedis/kyber"
 	"github.com/dedis/kyber/util/key"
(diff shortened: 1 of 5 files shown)


util/test/test.go
```diff
@@ -50,81 +51,82 @@ func testEmbed(g kyber.Group, rand cipher.Stream, points *[]kyber.Point,
...
 	one := g.Scalar().One()
 	for i := 0; i < N; i++ {
 		s1 := g.Scalar().Pick(rand)
 		s2 := s1.Clone()
 		if !s1.Equal(s2) {
-			panic("Clone didn't create a scalar s2 with same value as s1's.")
+			t.Errorf("Clone didn't work for scalar: %v != %v", s1, s2)
 		}
-		if !s1.Equal(one) {
+		if !s1.Equal(zero) && !s1.Equal(one) {
```
cag: I forgot to mention that it also considers this edge case correctly.

jeffallen:
Thank you for this.

PR #319, opened 2018-07-20
Add a toy group <2, *> in Z/127Z

cag:
Toy group does not have enough of a state space to allow non-deterministic tests, so I've also changed the signature of the GroupTest. I'm open to reverting this though.


jeffallen:
Thank you for this. Can you explain a bit more where you are going with this? What problem does this help you solve?

cag:
I'm so sorry, I totally thought I had responded here earlier! :sweat: 

So we're making an implementation of ECDKG as described in the paper by Tang, and I felt like having a toy group available during the tests could help with debugging any issues which arise.

jeffallen:
OK, I'm generally ok with this, but I'm heading out on vacation right now. Please be patient, and we'll merge this in a week or so.

cag:
Yes, I'm aware. Actually, I now suspect the implementation in the Tang paper is very similar to, if not the same as, the GJKR one...

ineiti:
Code comment:
- why change the signature of `GroupTest`? I don't see what it's doing differently now. Once all is unrolled, it still uses `random.New()`

Usage:
- what is the supposed usage of this? Can you give a usage example, please?

nikkolasg:
Hi there,
 
I also have to express my concerns about this PR. I don't see any benefits in merging that to Kyber tbh, for multiple reasons:
1. We never really had a need to test with "fake" groups so far. On the contrary I believe you should test with "real" groups your implementation
2. This toy group is simply a integer modulo'd, which is what we already have in the [`mod`](https://github.com/dedis/kyber/tree/master/group/mod) package in the end.

On a related note, I've looked over the paper [ECDKG paper](https://www.semanticscholar.org/paper/ECDKG-%3A-A-Distributed-Key-Generation-Protocol-Based-Tang/3c5235523be1d305de6dbf3433965c99d9fe4aea) by Tang and indeed, it really looks the same thing as the one from Rabin. The only "benefit" I have found is against adaptive adversary where he includes a Zero Knowledge Proof of node *i* having the share *s_i*. If needed, that could be included in both DKG for example (but we'd need to review exactly how that would work cause I'm not 100% convinced  of all the claims in this paper...).

ineiti:
You can also implement your own group in your code without having to touch the kyber library at all...

cag:
The reason for changing the signature was to accommodate this group which doesn't have enough space to reliably produce distinct random keys. Being able to specify the random seed for the tests make testing group properties for toy groups more reliable (but shouldn't affect testing for the other groups).

Also, `mod` implements `kyber.Scalar`. There isn't an implementation of a traditional DH exponentiation group in a ring modulo some large (say, 4096-bit) prime.

I can understand, however, objections 1 and 3, and will gladly close this.

Side note: I think you may be right about the Tang paper. I'm personally questioning the value of a ZK proof at that point in the protocol, though I suppose one can be implemented easily enough if there is a legitimate risk.

PR #322, opened 2018-09-05
Implementation of recovering public polynomial

nikkolasg:
This PR implements the recovering of a public polynomial (ie. all its coefficients) from a list of public shares. This is needed for the newest features coming into DKG.

### DIFF
--- a/share/poly.go
+++ b/share/poly.go
@@ -237,7 +237,7 @@ func RecoverPriPoly(g kyber.Group, shares []*PriShare, t, n int) (*PriPoly, erro
 	//den := g.Scalar()
 	// Notations follow the Wikipedia article on Lagrange interpolation
 	// https://en.wikipedia.org/wiki/Lagrange_polynomial
-	for j, _ := range x {
+	for j := range x {
 		basis := LagrangeBasis(g, j, x)
 		for i := range basis.coeffs {
 			basis.coeffs[i] = basis.coeffs[i].Mul(basis.coeffs[i], shares[j].V)
@@ -429,7 +429,7 @@ func RecoverPubPoly(g kyber.Group, shares []*PubShare, t, n int) (*PubPoly, erro
 	var accPoly *PubPoly
 	var err error
 
-	for j, _ := range x {
+	for j := range x {
 		basis := LagrangeBasis(g, j, x)
 		// compute the L_j * y_j polynomial in point space
 		tmp := basis.Commit(shares[j].V)
--- a/share/poly.go
+++ b/share/poly.go
@@ -234,29 +234,13 @@ func RecoverPriPoly(g kyber.Group, shares []*PriShare, t, n int) (*PriPoly, erro
 
 	var accPoly *PriPoly
 	var err error
-	den := g.Scalar()
+	//den := g.Scalar()
 	// Notations follow the Wikipedia article on Lagrange interpolation
 	// https://en.wikipedia.org/wiki/Lagrange_polynomial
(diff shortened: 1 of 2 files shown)


Daeinar:
@nikkolasg Looks good in general, I only have one small question: https://github.com/dedis/kyber/pull/322/files#r215196152

share/poly.go
```diff
@@ -426,3 +426,66 @@ func RecoverCommit(g kyber.Group, shares []*PubShare, t, n int) (kyber.Point, er
...
+	}
+
+	if len(x) < t {
+		return nil, errors.New("share: not enough good public shares to reconstruct secret commitment")
+	}
+
+	var accPoly *PubPoly
+	var err error
+
+	num := g.Scalar()
```
Daeinar: `num` seems to be never used other than being set to 1 in line 457?

Daeinar:
Modulo the two smaller comments above this looks good to me.

share/poly.go
```diff
@@ -426,3 +410,69 @@ func RecoverCommit(g kyber.Group, shares []*PubShare, t, n int) (kyber.Point, er
...
+	var err error
+
+	for j := range x {
+		basis := LagrangeBasis(g, j, x)
+		// compute the L_j * y_j polynomial in point space
+		tmp := basis.Commit(shares[j].V)
+		if accPoly == nil {
+			accPoly = tmp
+			continue
+		}
```
jeffallen: I prefer to read loops which initialize an accumulator when they are wrtitten with an else and no continue:
```
for {
  if acc == nil {
    acc = tmp
  } else {
   var err error
    acc, err = add.Add(tmp)
    if err != nil {
      return nil, err
    }
  }
} //end for
```

share/poly.go
```diff
@@ -234,29 +234,13 @@ func RecoverPriPoly(g kyber.Group, shares []*PriShare, t, n int) (*PriPoly, erro
 
 	var accPoly *PriPoly
 	var err error
-	den := g.Scalar()
+	//den := g.Scalar()
```
jeffallen: Please clean this up.

share/poly.go
```diff
@@ -426,3 +410,69 @@ func RecoverCommit(g kyber.Group, shares []*PubShare, t, n int) (kyber.Point, er
...
+	}
+
+	return accPoly, nil
+
+}
+
+// LagrangeBasis returns a PriPoly containing the Lagrange coefficients for the
+// i-th position. xs is a mapping between the indices and the values that the
+// interpolation is using, computed with xScalar().
+func LagrangeBasis(g kyber.Group, i int, xs map[int]kyber.Scalar) *PriPoly {
```
jeffallen: Is this function useful on it's own? Or should it be lower-case, so not exported?

nikkolasg: Good point, it'll be enough if we don't export it. I'll do that.

share/poly.go
nikkolasg: Mhhhh. I understand your preference, mine is more along the lines of clarity of code: 
+ It's simpler / easier to read without an additional indentation and if/else case to mentally remember.
+ It clearly indicates a breaking condition in the loop, same as returning early from a function.  

For example, we never do this:
```go
if err := myDangerousOperation(); err != nil {
   return err
} else {
   mySafeOperation()
}
```
but simply
```go
if err := myDangerousOperation(); err != nil {
   return err
} 
mySafeOperation()
```
What's your reasonings for not liking `continue` ? C/C++ background ? ;)

PR #323, opened 2018-09-07
Refresh DKG Shares Test

Daeinar:
This PR contains two things:
- It fixes a small bug in the secret sharing code where too many values were taken during Lagrange interpolation.
- It adds a test showing how to refresh DKG shares as presented in the paper "Verifiable Secret Redistribution for Archive Systems".

### DIFF
--- a/share/poly_test.go
+++ b/share/poly_test.go
@@ -3,6 +3,7 @@ package share
 import (
 	"testing"
 
+	"github.com/dedis/kyber"
 	"github.com/dedis/kyber/group/edwards25519"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


share/poly_test.go
```diff
@@ -317,3 +318,132 @@ func TestPriPolyCoefficients(test *testing.T) {
 	require.Equal(test, a.coeffs, b.coeffs)
 
 }
+
+func TestRefreshDKG(test *testing.T) {
```
jeffallen: s/test/t/ to keep with go standards, please.

share/poly_test.go
```diff
@@ -317,3 +318,132 @@ func TestPriPolyCoefficients(test *testing.T) {
...
+
+	// Verify VSS shares
+	for i := 0; i < n; i++ {
+		for j := 0; j < n; j++ {
+			sij := priShares[i][j]
+			// s_ij * G
+			sijG := g.Point().Base().Mul(sij.V, nil)
+			// Eval poly
+			prod := pubShares[i][j]
+			require.Equal(test, sijG.String(), prod.V.String())
```
jeffallen: This would seem to be a bit closer to what you actually want to test: `require.True(t, sijG.Equal(prod.V))`

And the other places below like this.

share/poly_test.go
```diff
@@ -317,3 +318,132 @@ func TestPriPolyCoefficients(test *testing.T) {
...
+			_, c := subPubPolys[j].Info()
+			pubShares[j] = &PubShare{I: j, V: c[i]}
+		}
+		com, err := RecoverCommit(g, pubShares, t, n)
+		require.NoError(test, err)
+		newDKGCommits[i] = com
+	}
+
+	// Check that the old and new DKG public keys are the same
+	require.Equal(test, dkgCommits[0].String(), newDKGCommits[0].String())
```
jeffallen: Would it be interesting/useful to check here that none of the old private shares exist in the set of new private shares?

share/poly_test.go
```diff
@@ -317,3 +318,132 @@ func TestPriPolyCoefficients(test *testing.T) {
 	require.Equal(test, a.coeffs, b.coeffs)
 
 }
+
+func TestRefreshDKG(test *testing.T) {
+	g := edwards25519.NewBlakeSHA256Ed25519()
+	n := 10
+	t := n/2 + 1
```
jeffallen: Would it be interesting/useful to test with other values for t? Where is the code that says, "error t is too small, can't do what you are asking for"?

Daeinar:
@jeffallen Sorry for the overlap with #322. I've branched off of Nicolas' PR which is why there is code duplication.

share/poly_test.go
Daeinar: We are using `t` as the threshold variable name which is the standard notation in the literature. To avoid conflicts we use `test` as the parameter here. That okay?

share/poly_test.go
Daeinar: Yeah true, I'll fix that.

share/poly_test.go
Daeinar: Yep good point, I will add a check.

share/poly_test.go
Daeinar: Actually now that you mention it: The refreshing phase should be done with new values for `t` and `n` as this might happen in real use cases.

share/poly_test.go
jeffallen: Oh yeah, that makes sense. Forgot about that particularity of these files.

PR #326, opened 2018-09-10
[DKG] Resharing protocol !

nikkolasg:
This PR brings the resharing protocol implementation. Essentially, having a set of initially-DKG-generated shares, to re-shares those shares to a potentially disconnected new set of nodes of arbitrary size, with a potentially new threshold. It re-uses the same DKG structure and messages but augment it with a `Config` structure which gives all the information needed to use a `DistKeyGenerator`. 

Additionally, it fixes a bug in `poly.go` which sorts the shares given to `RecoverXXX`. In some cases, especially in the resharing case, different ordering can lead to different recovered items in the end. Very subtle nasty bug ><.

**Questions**: I have some namings problem here upon which I would love to have your opinion.
I now have this `Config` that is given to create a `DistKeyGenerator` (the structure that handles the logic and the messages). To keep compatibility, I have kept the current `NewDistKeyGenerator` constructor as it is, but I also added a `NewDistKeyHandler` that takes this new `Config`. There's clearly some naming issues there but I'm not sure about the right approach yet. I'll think about it more but in the meantime, if you have any idea, you're most welcomed !


share/dkg/pedersen/dkg.go
```diff
@@ -224,16 +371,30 @@ func (d *DistKeyGenerator) ProcessDeal(dd *Deal) (*Response, error) {
 // If the response designates a deal this dkg has issued, then the dkg will process
 // the response, and returns a justification.
 func (d *DistKeyGenerator) ProcessResponse(resp *Response) (*Justification, error) {
+	/*fmt.Printf("\t response index %d, dkg index %d\n", resp.Index, d.oidx)*/
```
jeffallen: remove debugging?

share/dkg/pedersen/dkg.go
```diff
@@ -208,9 +319,45 @@ func (d *DistKeyGenerator) ProcessDeal(dd *Deal) (*Response, error) {
 		return nil, err
 	}
 
+	reject := func() (*Response, error) {
+		idx, present := findPub(d.c.NewNodes, pub)
+		if present {
+			d.verifiers[uint32(idx)].UnsafeSetResponseDKG(uint32(idx), vss.StatusComplaint)
+		}
+		// indicate to VSS that the new status is complaint, since the check is
+		// done outdone VSS package control.
```
jeffallen: done outdone?

share/dkg/pedersen/dkg.go
```diff
@@ -244,17 +405,60 @@ func (d *DistKeyGenerator) ProcessResponse(resp *Response) (*Justification, erro
...
+func (d *DistKeyGenerator) processResharingResponse(resp *Response) (*Justification, error) {
+	// check the signature authenticity
+	npub, present := getPub(d.c.NewNodes, resp.Response.Index)
+	if !present {
+		return nil, errors.New("dkg: resharing response invalid index")
+	}
+	err := schnorr.Verify(d.suite, npub, resp.Response.Hash(d.suite), resp.Response.Signature)
+	if err != nil {
+		// can't return justification with invalid signature otherwise leads to
+		// attack on other nodes. One attacker can simply put the target victimŝ
```
jeffallen: Nice decoration on your s there...

share/dkg/pedersen/dkg.go
```diff
@@ -1,143 +1,232 @@
...
+// with the current share of the node. If the node using this config is a new
+// addition and thus has no current share, the PublicCoeffs field be must be
+// filled in.
+type Config struct {
+	Suite Suite
+
+	// Longterm is the longterm secret key.
+	Longterm kyber.Scalar
+
+	// Current group of share holders. It can be nil for new DKG. These nodes
```
jeffallen:
```
	// Current group of share holders. It will be nil for new DKG. These nodes
	// will have invalid shares after the protocol has been run. To be able to issue
	// new shares to a new group, the group member's public key must be inside this
	// list and in the Share field. Keys can be disjoint or not with respect to the
        // NewNodes list.
```

share/dkg/pedersen/dkg.go
```diff
@@ -1,143 +1,232 @@
...
+	// will have invalid share after the protocol is ran. To be able to issue
+	// new fresh shares to a new group, one's public key must be inside this
+	// list alongside with the Share field. Keys can be disjoint or not with
+	// respect to the NewNodes list.
+	OldNodes []kyber.Point
+
+	// PublicCoeffs are the coefficients of the distributed polynomial needed
+	// during the resharing protocol. The first coefficient is the key.It is
```
jeffallen: add missing space after key

share/dkg/pedersen/dkg.go
```diff
@@ -1,143 +1,232 @@
...
+	// required for new share holders.  It can be nil for new DKG.
```
jeffallen: It should be nil for a new DKG.

(Is this checked someplace?)

share/dkg/pedersen/dkg.go
```diff
@@ -1,143 +1,232 @@
...
+	PublicCoeffs []kyber.Point
+
+	// Expected new group of share holders. These public-key designated nodes
+	// will be in possession of new shares after the protocol is ran. To be a
```
jeffallen: has been run

share/dkg/pedersen/dkg.go
```diff
@@ -1,143 +1,232 @@
...
+	// receiver a of new share, one's public key must be inside this list. Keys
+	// can be disjoint or not with respect to the OldNodes list.
+	NewNodes []kyber.Point
+
+	// Share to refresh. It must be nil for fresh DKG or new node wishing to
```
jeffallen: for a fresh DKG or for a new node

share/dkg/pedersen/dkg.go
```diff
@@ -208,9 +321,45 @@ func (d *DistKeyGenerator) ProcessDeal(dd *Deal) (*Response, error) {
...
+		expectedPubShare := d.dpub.Eval(int(dd.Index))
+		if !expectedPubShare.V.Equal(dealCommits[0]) {
+			return reject()
+		}
+	}
+
 	// Set StatusApproval for the verifier that represents the participant
 	// that distibuted the Deal
-	d.verifiers[dd.Index].UnsafeSetResponseDKG(dd.Index, vss.StatusApproval)
+	// This must be set at all times but when the dealer is not in the newNodes
```
Daeinar: s/but/except/ ?

share/dkg/pedersen/dkg.go
```diff
@@ -244,17 +406,60 @@ func (d *DistKeyGenerator) ProcessResponse(resp *Response) (*Justification, erro
...
 	}
 
 	return &Justification{
-		Index:         d.index,
+		Index:         uint32(d.oidx),
 		Justification: j,
 	}, nil
 }
 
+// special case when an node that is present in the old list but not in the
```
Daeinar: s/special/Special/ 
s/an/a/ 
s/that//

share/dkg/pedersen/dkg.go
```diff
@@ -244,17 +406,60 @@ func (d *DistKeyGenerator) ProcessResponse(resp *Response) (*Justification, erro
...
+// new,i.e. leaving the group, does not have any verifiers since it can't
```
Daeinar: "new, i.e., it is leaving the group, then the node does not have ..."

share/dkg/pedersen/dkg.go
```diff
@@ -244,17 +406,60 @@ func (d *DistKeyGenerator) ProcessResponse(resp *Response) (*Justification, erro
...
+// receive shares. This function makes some check on the response and returns a
```
Daeinar: s/makes some check on/checks/

share/dkg/pedersen/dkg.go
```diff
@@ -273,18 +478,19 @@ func (d *DistKeyGenerator) SetTimeout() {
...
-	return len(d.QUAL()) >= len(d.participants)
+	return len(d.QUAL()) >= len(d.c.OldNodes)
 }
 
 // QUAL returns the index in the list of participants that forms the QUALIFIED
 // set as described in the "New-DKG" protocol by Rabin. Basically, it consists
-// of all participants that are not disqualified after having  exchanged all
-// deals, responses and justification. This is the set that is used to extract
-// the distributed public key with SecretCommits() and ProcessSecretCommits().
+// of all valid deals at the end of the protocols. It does NOT take into account
```
Daeinar: s/protocols/protocol/

share/dkg/pedersen/dkg.go
```diff
@@ -273,18 +478,19 @@ func (d *DistKeyGenerator) SetTimeout() {
...
+// any malicious share holder which share may have been revealed, due to invalid
```
Daeinar: "due to an invalid complaint"

share/dkg/pedersen/dkg.go
```diff
@@ -1,143 +1,234 @@
...
-}
+	var isResharing bool
+	if c.Share != nil || c.PublicCoeffs != nil {
+		isResharing = true
+	}
+	// canReceive is true by default since in the default DKG mode everyone
+	// participates
+	var canReceive = true
+	pub := c.Suite.Point().Mul(c.Longterm, nil)
+	oidx, oldPresent := findPub(c.OldNodes, pub)
```
Daeinar: Can we use a `map[int]kyber.Point` for `OldNodes` and `NewNodes` in `Config` and avoid the `O(n)` overhead for the key search?

share/poly.go
```diff
@@ -373,20 +393,43 @@ func (p *PubPoly) Check(s *PriShare) bool {
 	return pv.V.Equal(ps)
 }
 
-// RecoverCommit reconstructs the secret commitment p(0) from a list of public
-// shares using Lagrange interpolation.
-func RecoverCommit(g kyber.Group, shares []*PubShare, t, n int) (kyber.Point, error) {
+// xyCommits is the public version of xScalars.
+func xyCommit(g kyber.Group, shares []*PubShare, t, n int) (map[int]kyber.Scalar, map[int]kyber.Point) {
+	// we are sorting first the shares since the shares may be unrelated for
+	// some applications. In this case, all participants needs to interpolate on
```
Daeinar: s/needs/need/

share/poly.go
```diff
@@ -199,18 +199,38 @@ func RecoverSecret(g kyber.Group, shares []*PriShare, t, n int) (kyber.Scalar, e
 	return acc, nil
 }
 
-func xScalar(g kyber.Group, shares []*PriShare, t, n int) map[int]kyber.Scalar {
+// xScalar returns the list of (x_i, y_i) pairs indexed. The first map returned
+// is the list of x_i and the second map is the list of y_i, both indexed in
+// their respective map at index i.
+func xScalar(g kyber.Group, shares []*PriShare, t, n int) (map[int]kyber.Scalar, map[int]kyber.Scalar) {
+	// we are sorting first the shares since the shares may be unrelated for
```
Daeinar: What do you mean by "unrelated"? Is that the case when we construct a new polynomial via Lagrange interpolation in the DKG re-sharing case?

share/poly.go
```diff
@@ -199,18 +199,38 @@ func RecoverSecret(g kyber.Group, shares []*PriShare, t, n int) (kyber.Scalar, e
...
+	// some applications. In this case, all participants needs to interpolate on
```
Daeinar: s/needs/need/

share/dkg/pedersen/dkg.go
nikkolasg: The `O(n)` overhead will have to be done by the caller in that case. The caller most often has a simple list of pubic keys and not a map, so it has to create it first. I prefer less action for the caller.

share/poly.go
nikkolasg: Yes exactly. Uncorrelated may be a better word ?

nikkolasg:
@ineiti it's still a work in progress in fact, there's some changes I need to backport to this branch. Will do as soon as I'm finished to get a working drand integration (there's a lot of back and forth between drand and these changes).

PR #327, opened 2018-09-14
Bn256 update

jeffallen:
Update our copy of Cloudflare's bn256 to include their recent changes. This includes supporting bn256 on platforms with no assembly implementation, support for Go 1.11, and a fix for a register stomping bug on Darwin ARM64 (see golang/go#26110).


nikkolasg:
I am totally not equipped with the right knowledge to be able to comment on that , so ... LGTM ;)

PR #329, opened 2018-09-26
fix small typo/bug

### LINKED ISSUE #328: potential issue: 0x3f != 127
I noticed that our `Curve.NewKey()` implementation in package edwards25519 is different from the one described in https://cr.yp.to/ecdh.html
```
mysecret[0] &= 248;
mysecret[31] &= *127;*  <== here we use 0x3f
mysecret[31] |= 64;
```

does someone know why ? or if it is an issue ?
looks suspicious to me since we have the "correct"/same_as_ cr.yp.to value in the vartime implementation (in package curve25519)
(side note, seems to me that the `NewKey()` function has nothing to do here, since this is code that defines a *curve*)

https://github.com/dedis/kyber/blob/b723602b9b0f02a52729cabf89a7774ce5be0503/group/edwards25519/curve.go#L48-L60

sarkortrantor:
fix #328

### DIFF
--- a/group/edwards25519/curve.go
+++ b/group/edwards25519/curve.go
@@ -52,7 +52,7 @@ func (c *Curve) NewKey(stream cipher.Stream) kyber.Scalar {
 	random.Bytes(buffer[:], stream)
 	scalar := sha512.Sum512(buffer[:])
 	scalar[0] &= 0xf8
-	scalar[31] &= 0x3f
+	scalar[31] &= 0x7f
 	scalar[31] |= 0x40
 
 	secret := c.Scalar().SetBytes(scalar[:32])


ineiti:
In fact the previous implementation follows

https://tools.ietf.org/html/rfc8032#page-22
```
a = int.from_bytes(h[:32], "little")
a &= (1 << 254) - 8
a |= (1 << 254)
```

But anyway, line 56 makes either way correct. We either (counting from 0)
1. (previous implementation, rfc8032) set bit 6 and 7 to 0, then set bit 6 to 1
or
2. (your proposition, cr.yp.to) set bit 7 to 0, then set bit 6 to 1

The output of either 1 or 2 is that bit 7 is always 0 and bit 6 is always 1.

sarkortrantor:
:man_facepalming: I see ^^ thank you
anyway I think that both NewKey() (the other in curve25519 package) should be harmonized

ineiti:
No, we don't want to harmonize them, because our curve-implementations are much more fundamental than EdDSA. We could perhaps change name, but the current implementation is good, because:
1. for private/public keys, you want to make the private key random, so you use sha512 like in EdDSA
2. for more fundamental cryptography, you want to have the point be only X = x * B, and not having any fancy hashing in there

So the curve25519's `NewKey` could be called `NewPair` or something, to indicate that we're not creating a key (private/public), but a pair of scalar/point. Which is mathematically the same, but from a usage point of view will differ.

sarkortrantor:
but this is two curve implementation and they have both nothing to do with EcDSA, 
hence I don't get what you are saying (and currently the only difference between these NewKey function is the 127 != 0x3f).
as of naming I agree with you NewKey has nothing to do in a group/{curve25519, edwards25519, whatever}/*

nikkolasg:
We follow the EdDSA RFC which is to be implemented everywhere so compatibility is paramount here.

For the `NewKey` you're coming back to an old discussion of years old. From what I remember, we wanted to put that as `NewKey` because it is exactly intented to be used as a keypair generation and not simply a scalar or a point on the curve. But we still want the general implementation of the edwards25519 curve so that's the best way we came up. 
+ The curve can give point and scalar non-modified
+ the `NewKey` can give keys compatible with EdDSA.

sarkortrantor:
Totally understand, but my point is that both NewKey should have same implementation (now there is a ~syntactic difference) (they already have same behavior as @ineiti pointed out, and yes that behavior should be compatible with EdDSA, which is the case).
but this is a detail/non-issue.

nikkolasg:
@sarkortrantor Well, I'm not sure about that, do you have ressources that shows curve25519 using the bit twiddling of EdDSA ? The important point is to be compatible, so if we want to do interoperable implementation of Diffie Hellman key exchange using curve25519, we have to get this specific expected bit twiddling, which might be different than the one from eddsa. But I am not sure if they are the same or not, if you do find something let us know !

sarkortrantor:
I don't have any resources on these questions as this is not my point ;),
I only say that 
1) *currently our implementations of NewKey() differs (see linked snippets)

a) Curve25519/curve.go
https://github.com/dedis/kyber/blob/b723602b9b0f02a52729cabf89a7774ce5be0503/group/curve25519/curve.go#L82-L94
VS
b) edwards25519/curve.go
https://github.com/dedis/kyber/blob/b723602b9b0f02a52729cabf89a7774ce5be0503/group/edwards25519/curve.go#L48-L60

2) the only difference is the one pointed out
3) which happens to not be a difference, the difference is only visible at the source code level, semantically there is no difference
4) hence I think that the two NewKey() should have same form in the code, but again "we don't really care"




side note: more generally the only difference between our curve implementations is that edwards25519 has constant time ops and curve25519 vartime ops (not sure about the "only part" of statement above, but this is what we quickly concluded with @ineiti) and if that is true then curve25519 should be renamed too since Curve25519 happens to be the name of a well defined/known/whatever Montgomery curve.

(and there are other potential naming issues e.g. now it is generally accepted that "Ed25519 is used to refer to EdDSA on an edwards curve birationally equivalent to Curve25519 (...)" and is not the curve's name.

 I'll open a google doc on these issues

nikkolasg:
Ok thanks for this great clarification !
So my suggestion would still be to see if there is any RFC for Curve25519, or any golden standard, and stick to that. If that turns out to be equivalent to the bit twiddling in the edwards25519 package, then fine, but it's fine also if it does not.  In my opinion, compatibility > homogeneity in the code.

Daeinar:
[RFC 7748](https://tools.ietf.org/html/rfc7748#section-5) specifies the equivalent to `scalar[31] &= 127` (see Section 5, page 7), so I would propose we stick to that too.

Issue #312, opened 2018-05-24
Consider adding property-based testing


kc1212:
A lot of our tests are performed using just a couple of fixed, arbitrary messages. But encryption and signature schemes can be seen as pure function, which is exceptionally suitable for property-based testing. In brief, property-based testing, i.e. quickcheck, will test for one or more properties that the programmer specifies, for example `Dec(Enc(m)) == m`, with random inputs, up to a user defined size/length. Consider improving our tests using the quickcheck package https://golang.org/pkg/testing/quick/

nikkolasg:
While this sounds generally reasonable, I'm actually worried about "randomized testing" for cryptographic schemes for multiple reasons:. 
+ Relying on randomized / fuzzing testing to assert that our construction is "good" is misleading and may let us believe that "if it passes this fuzzing test, it must be secure", while it's not true *at all*. Verifying with random inputs does not bring more guarantees to our cryptographic constructions, because we will never know if it is valid for all inputs anyway.

+ It might happen that we have sometimes a test failing for some input and it might be quite hard to debug if multiple developers can't get the same test base easily,i.e. can't reproduce a minimum test which fails.

Again, while in general I think it's a good idea to try out (in onet!), I'm not so sure about this in the context of kyber, but I'd be happy to hear any other thoughts on the matter !

kc1212:
Let me address your points:
- Property-based testing is not to assert that a cryptographic construction is secure, it's more to do with correctness. It's simply to catch corner cases caused by a developer bug. For example, an empty message.
- This is not what a good quickcheck implementation will do. I don't know about go's implementation but the Haskell and Rust implementations do the following. They start by trying random inputs (it's not exactly random but it tries inputs with different properties, e.g. messages of different lengths, u can also specify how you want to generate these random inputs). Upon hitting an input that causes the test to fail, it will try to reduce the input and try to reproduce the same failure such that the bad-input is of minimum size/length. Every step of the reduction is printed out, this is very useful for debugging because you can see what kind of inputs are causing failures. Finally the developer can use the smallest bad-input (or any other one in the reduction chain) to reproduce the error.

nikkolasg:
> Property-based testing is not to assert that a cryptographic construction is secure, it's more to do with correctness. It's simply to catch corner cases caused by a developer bug. For example, an empty message.

I'm not educated enough in property-based testing and even less in the quick package that Go has so bear with me please :) :  

+ What kind of corner cases are you envisionning ? Here's a few I can think of:
[...]

PR #334, opened 2018-10-19
add a quick test to Schnorr signature kyber-312

Gilthoniel:
This PR gives an example of using `testing/quick` for property-based testing for the Schnorr signature

Draft for #312

### DIFF
--- a/sign/schnorr/schnorr_test.go
+++ b/sign/schnorr/schnorr_test.go
@@ -1,6 +1,8 @@
 package schnorr
 
 import (
+	"math/rand"
+	"reflect"
 	"testing"
 	"testing/quick"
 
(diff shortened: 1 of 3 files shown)


sign/schnorr/schnorr_test.go
```diff
@@ -62,3 +65,34 @@ func TestEdDSACompatibility(t *testing.T) {
...
+	s.rand.Read(dst)
+}
+
+func (s *quickstream) Generate(rand *rand.Rand, size int) reflect.Value {
+	return reflect.ValueOf(&quickstream{rand: rand})
+}
+
+func TestQuickSchnorrSignature(t *testing.T) {
+	f := func(rand *quickstream, msg []byte) bool {
+		suite := edwards25519.NewBlakeSHA256Ed25519WithRand(rand)
```
kc1212: why do you need `rand`? can't you just use `edwards25519.NewBlakeSHA256Ed25519`?

Gilthoniel: It was in order to use the random instance provided by quick. As the key length is fixed, I could remove it and generate only the message.

kc1212:
Actually, won't merge this one yet. Will discuss with the other on Monday on whether we're committed to `testing/quick`.

PR #336, opened 2018-10-25
Test and fix a panic caused by input too short

### LINKED ISSUE #335: panic in cosi.go
While checking something using "./ftcosi check -l -g ../dedis-cothority.toml", I got a panic due to this code:

```
// Verify checks the given cosignature on the provided message using the list
// of public keys and cosigning policy.
func Verify(suite Suite, publics []kyber.Point, message, sig []byte, policy Policy) error {
        if publics == nil {
                return errors.New("no public keys provided")
        }
        if message == nil {
                return errors.New("no message provided")
        }
        if sig == nil {
                return errors.New("no signature provided")
        }
        if policy == nil {
                policy = CompletePolicy{}
        }

        lenCom := suite.PointLen()
        VBuff := sig[:lenCom]
```

It seemed that sig[:lenCom] was out of bounds. Need a bounds check on that, I guess. Don't know why sig was short though...

jeffallen:
Fixes #335.

### DIFF
--- a/sign/cosi/cosi_test.go
+++ b/sign/cosi/cosi_test.go
@@ -116,6 +116,10 @@ func testCoSi(t *testing.T, n, f int) {
 		} else {
 			p = NewThresholdPolicy(n - f)
 		}
+		// send a short sig in, expect an error
+		if err := Verify(testSuite, publics, message, sig[0:10], p); err == nil {
+			t.Fatal("expected error on short sig")
+		}
 		if err := Verify(testSuite, publics, message, sig, p); err != nil {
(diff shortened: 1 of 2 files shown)


sign/cosi/cosi_test.go
```diff
@@ -116,6 +116,10 @@ func testCoSi(t *testing.T, n, f int) {
 		} else {
 			p = NewThresholdPolicy(n - f)
 		}
+		// send a short sig in, expect an error
```
ineiti: nitpick: why not use `require.Nil` here?

PR #338, opened 2018-11-14
adds BLS signature aggregation

tobowers:
This adds BLS signature aggregation and and aggregated signature verification. The method used is ported directly from libindy-crypto: https://github.com/hyperledger/indy-crypto/blob/master/libindy-crypto/src/bls/mod.rs#L312-L323

This also adds a few benchmarks which I found useful in comparing other libraries.

kyber is great work!


sign/bls/bls.go
```diff
@@ -31,6 +31,29 @@ func Sign(suite pairing.Suite, x kyber.Scalar, msg []byte) ([]byte, error) {
...
+			return nil, err
+		}
+		sig.Add(sig, sigToAdd)
+	}
+	return sig.MarshalBinary()
+}
+
+// VerifyMultiSignature takes an aggregated signature and a slice of public Points
+// and verifies that the signature is valid
+func VerifyMultiSignature(suite pairing.Suite, Xs []kyber.Point, msg, sig []byte) error {
```
nikkolasg: Having played around with multisigs already, I find it better to expect the aggregated signature AND aggregated public key already. Usually, you select which public key and signatures to aggregate from a bitset for example. When iteratating the bitset, you can already aggregate, instead of creating a temporary slice containing only the public keys to aggregate to be able to give it to this function. Plus, it's also more consistent to either expect aggregated stuff or individual stuff separately.

tobowers: I got rid of this method, and added a AggregateKeys method

tobowers:
I got the email and replied.

PR #342, opened 2018-11-28
Make bn256 available in constant time builds

Gilthoniel:
This removes the build vartime build tag so that any suites is available. It also implements a adapter for the Bn256 suite to make it compatible with the common suites.Suite interface and still keep the pairing.Suite properties.


nikkolasg:
Hey. 

I understand the need to make bn256 available dynamically via the `suites/` package but I don't believe making the following "hacks" should be the only way:
1. Allowing variable time implementation in the constant time package: this breaks the assumption about the constant time suite implementations if we start mixing them, so this raises the question should this constant / var time separation go on ?
2. Allowing `pairing.Suite` to be a `kyber.Group`: one does not know from which group a `Point` or a `Scalar` comes from and is semantically wrong imho, since the `pairing.Suite` is NOT a `Suite`.

I would much highly prefer thinking about removing the const/var time separation which is in the end useless for us and then register the three real suites `bn256_G1` `bn256_G2` `bn256_GT`. 

But don't forget I'm not working here anymore ;D

Gilthoniel:
@jeffallen Regarding the answer of @nikkolasg the constant time is more a question for you. Now for the non-suite, one way to solve this could be to create a wrapper around the three suites so that it can be used by onet.

Gilthoniel:
@jeffallen @nikkolasg This version should be better

nikkolasg:
I like this much better, since it kinda solves the second problem I mentioned. 

I'd still remove the distinction with non-constant time, since this PR is still introducing a HACK in my opinion but maybe that's worth a separate PR.

jeffallen:
We need to get feedback from @bford on the vartime thing.

group/edwards25519/scalar.go
```diff
@@ -23,6 +23,8 @@ import (
 
 // The scalars are GF(2^252 + 27742317777372353535851937790883648493).
 
+var marshalScalarID = [8]byte{'e', 'd', '.', 's', 'c', 'a', 'l', 'a'}
```
jeffallen: Ooh, the Scala fan boys are gonna like this.

nikkolasg:
Still receiving updates on the repo so I could not resist: 
My main concern is forcing the MarshalID thing into the implementation themselves while it could (and should imho) be separated since this has nothing to do with points and scalar, but only the encoding which is separated from `kyber` even. One could register an `edwards25519.Point` with any marshalID they wish.

But please, feel free to `comment > /dev/null` if you "just wanna go ahead".

group/curve25519/ext.go
```diff
@@ -83,11 +81,11 @@ func (P *extPoint) HideDecode(rep []byte) {
 //		iff
 //	(X1*Z2,Y1*Z2) == (X2*Z1,Y2*Z1)
 //
-func (P1 *extPoint) Equal(CP2 kyber.Point) bool {
+func (P *extPoint) Equal(CP2 kyber.Point) bool {
 	P2 := CP2.(*extPoint)
 	var t1, t2 mod.Int
-	xeq := t1.Mul(&P1.X, &P2.Z).Equal(t2.Mul(&P2.X, &P1.Z))
-	yeq := t1.Mul(&P1.Y, &P2.Z).Equal(t2.Mul(&P2.Y, &P1.Z))
+	xeq := t1.Mul(&P.X, &P2.Z).Equal(t2.Mul(&P2.X, &P.Z))
```
ineiti: Question: does this modify `P.X`? If so, then this is a race condition!

Gilthoniel: From a quick look, it seems so. Would it be okay to fix that in a different PR ? I only renamed here to fix the lint warning.

pairing/adapter.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/dedis/kyber/pairing/bn256"
+)
+
+// SuiteBn256 is an adapter from the different bn256 suites to a single one
+// that can be used by onet to encode/decode key pairs. The public key comes
+// from the field G2 and the private key comes from G1 thus the Point and
+// Scalar functions return the corresponding type.
+type SuiteBn256 struct {
+	Suite
+	kyber.Group
```
kc1212: why do you need the `kyber.Group`?

pairing/adapter.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/dedis/kyber"
+	"github.com/dedis/kyber/pairing/bn256"
+)
+
+// SuiteBn256 is an adapter from the different bn256 suites to a single one
+// that can be used by onet to encode/decode key pairs. The public key comes
+// from the field G2 and the private key comes from G1 thus the Point and
+// Scalar functions return the corresponding type.
```
kc1212: I might be wrong, but I suspect that it doesn't matter which group the pk is from. Out BLS implementation I think is similar to this one https://en.wikipedia.org/wiki/Boneh%E2%80%93Lynn%E2%80%93Shacham where the pk (g^x) is in G2. But for this one https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html the pk is in G1 (G0 in their notation). I think as long as you run the other pairing operation correctly then it's fine. G1 and G2 have the same order so I think you can sign a message in both groups as long as you verify it correctly. Also, the scalar has the same order in both groups so it doesn't matter where it comes from.

Maybe there's nothing to be done here but update the comment to say this adapter is specific to our BLS implementation.

pairing/adapter.go
Gilthoniel: To implement the suites.Suite interface that has the .Point() function used to create public key points. Suite comes from the pairing suite that supports three kinds of suite but the interface assumes only one. That makes the suite compatible with the code that is using the interface.

nikkolasg: shut me down cause I can't resist: that is a hack producing wrong semantics (there's no such things as a bn256 point) because of a problem NOT in kyber ... :(
Why kyber should suffer for the wrongness in other repos while we spent time and effort making it good ... :/
One could imagine using these interface's ID @Gilthoniel is doing in protobuf for `bn256_G1` `bn256_G2` and `bn256_GT`, wouldn't it solve the problem ... ?

pairing/adapter.go
Gilthoniel: Actually it's not similar because we have different groups G_1 and G_2 with respectively GF(p) and GF(p^2). It's true the scalar has the same order though. The pk is in G2 (G1 for them) in the Standford paper and the signature in G1 (G0 for them).

I will add the comment.

pairing/adapter.go
kc1212: this is a good point.. maybe we need to re-think what the pairing suite should look like.

I think the reason we need this kind of "standard" suite is because when onet/cothority registers one, it's one suite per service. What are the consequences if we register `pairing.Suite` @Gilthoniel ? It should be possible to test the suite implementation using the casting trick in go.

Gilthoniel: Actually as @kc1212 pointed out, scalars are the same in both groups so we could use G2 as the public suite because G1 is used only internally when signing and verifying... I'm gonna try that tomorrow and see if it works.

Gilthoniel: Ok now I remember why I did an adapter. You still need functions from the parent suite (e.i. the Pair function) then the purpose of the adapter was to keep the properties of the parent suite but make it compatible with the common suite interface.

I understand your worries about the point concept but in order to make it work, you need to have a common base to generate public and private keys. Take the key pair generator that takes a suites.Suite and thus expect Point and Scalar.

I'm not saying you're wrong and that's the only way. I guess one solution could be to rethink the suite interface so that it implements basic tools required like a key pair generation, sign, verify etc ... and hide any point logic from outside. But my point here is to make something that won't deeply change kyber.

nikkolasg: > Ok now I remember why I did an adapter. You still need functions from the parent suite (e.i. the Pair function) then the purpose of the adapter was to keep the properties of the parent suite but make it compatible with the common suite interface.

It's too late now but if one is using such functionalities `Pairing` then it knows it is using a pairing suite, therefore could it not retrieve the corresponding `pairing.Suite` from a given point or scalar ? i.e. go `bn256_G1` to `bn256.Suite`. 

Anyway... :(

PR #343, opened 2018-12-06
Make it possible to reshare among DKG group members

jeffallen:


group/edwards25519/curve.go
```diff
@@ -52,7 +52,7 @@ func (c *Curve) NewKey(stream cipher.Stream) kyber.Scalar {
 	random.Bytes(buffer[:], stream)
 	scalar := sha512.Sum512(buffer[:])
 	scalar[0] &= 0xf8
-	scalar[31] &= 0x7f
```
kc1212: why was this change? I don't think it's copied from the other PR because I only copied the files in the share folder.

pairing/bn256/mul.h
```diff
@@ -0,0 +1,181 @@
+#define mul(a0,a1,a2,a3, rb, stack) \
```
kc1212: same question, I don't understand where this came from, perhaps a bad merge?

sign/schnorr/schnorr_test.go
```diff
@@ -65,34 +62,3 @@ func TestEdDSACompatibility(t *testing.T) {
...
-			return false
-		}
-
-		return Verify(suite, kp.Public, msg, s) == nil
-	}
-
-	if err := quick.Check(f, nil); err != nil {
-		t.Error(err)
-	}
-}
```
kc1212: i think this is a sign of a bad merge, I remember reviewing this code but here it got removed.

jeffallen: I agree. The problem I'm having is that somehow the commits you were building on have two parents, but are not merges. I'll try again a different way and see what I get.

PR #344, opened 2018-12-11
Stop testing create_stable

jeffallen:
We are gettingr eady to make major changes that we would not want
to push to v2 anyway.


Makefile
```diff
@@ -12,5 +11,6 @@ test_playground:
 	  go test -v -race -short || exit 1 ; \
 	done;
 
-# Other targets are:
-# make create_stable
+# override the test target from Makefile.base because we do not want
+# to use test_stable_build right now.
+test: test_fmt test_lint test_goveralls
```
Gilthoniel: It doesn't override but add more commands

PR #345, opened 2018-12-11
checking out redist2

ineiti:
Merged all commits - but losing history


jeffallen:
Thanks for the experiment. :)

PR #346, opened 2018-12-11
Remove the vartime build tag

Gilthoniel:
This removes entirely the concept of vartime builds and thus
all the suites are available when building the project


ineiti:
I'm still somewhat hesitating to remove the `vartime` flag. It remains an important indicator of the quality of the suite used. Can we have some kind of `MustFindVartime` or so that includes the `vartime` suites? And `MustFind` only returns the constant-time ones? We could even keep it backward-compatible by adding a different `MustFind` method when using `vartime`.

Is it difficult to find which projects are using kyber with `vartime`?

group/nist/profile.tmp
```diff
@@ -0,0 +1 @@
+mode: atomic
```
kc1212: remove this file?

jeffallen:
This change will be included in #342.

PR #347, opened 2018-12-20
A mechanism for callers to opt-in to a limit on suites

jeffallen:
Previously, vartime suites were not even compiled in. We found this to
be difficult to work with when we need to mix vartime and constant
time algorithms (i.e. in Cothority). This is a compromise: by having
callers add one line to their init() function in package main,
they can opt-in to a different implementation that still provides
the same safety as the old implementation.

### DIFF
--- a/suites/suites_test.go
+++ b/suites/suites_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Test_Find(t *testing.T) {
+func TestSuites_Find(t *testing.T) {
 	ss := []string{
 		"ed25519",
 		"bn256.G1",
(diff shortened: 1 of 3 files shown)


suites/suites_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestSuites_Find(t *testing.T) {
```
Gilthoniel: it's useful to have a prefix for the tests so that you can:
```
go test -run ^TestSuites
```
If one day there are multiple files in the package

jeffallen: Just thought it was cleaner... my instinct is always the smallest identifier that is unique. But I get your point and fixed it.

kc1212:
Can we also update the README to mention this feature?

PR #348, opened 2018-12-20
batch verification of BLS signature/msg pairs

tobowers:
This adds batch verification of a number of signatures where the messages are distinct. Generally this is just used for a performance enhancement. The benchmarks show that the verification is time for the 100 sigs here takes roughly the same amount as 50 signature verifications individually.

### DIFF
--- a/sign/bls/bls.go
+++ b/sign/bls/bls.go
@@ -5,7 +5,9 @@ package bls
 
 import (
 	"crypto/cipher"
+	"crypto/sha256"
 	"errors"
+	"fmt"
 
 	"github.com/dedis/kyber"
 	"github.com/dedis/kyber/pairing"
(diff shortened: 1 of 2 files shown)


sign/bls/bls.go
```diff
@@ -80,3 +114,15 @@ func hashToPoint(suite pairing.Suite, msg []byte) kyber.Point {
 	x := suite.G1().Scalar().SetBytes(h.Sum(nil))
 	return suite.G1().Point().Mul(x, nil)
 }
+
+func distinct(msgs [][]byte) bool {
+	m := make(map[[32]byte]bool)
+	for _, msg := range msgs {
+		h := sha256.Sum256(msg)
```
tobowers: this takes about 1-2ms for the 100 sigs... I wonder if there's something faster here.

jeffallen: a comment explaining why distinctness is required would be helpful here. And a test that confirms that repeated messages are correctly refused.

tobowers: Good ideA

tobowers:
No worries on the timing. Thanks! Happy holidays! :party:

jeffallen:
@kc1212 , do you think that we should be implementing masking here, in the same way you did for BLS sigs in Java?

kc1212:
I think the mask can be kept independent. In kyber, the mask is in the cosi package, which is actually not the right place because it's independent of cosi. We actually use cosi.NewMask as the BLS signature mask in skipchain. So the batch verification of bls can stay because it does some bls specific things like computing the pairing etc. If someone wants to use the mask to have fine-grained control, they'll use the mask method to aggregate the public keys and signatures (not possible for cosi).

sign/bls/bls.go
tobowers: I added a comment and a test

PR #352, opened 2019-01-17
Removing the Hiding interface

### LINKED ISSUE #117: AGL's Elligator Code Might be Flawed
@isislovecruft and @hdevalence notified me that AGL's Elligator code might be flawed and thus should not be used. We should follow up with them on the exact details. A good documentation for Elligator that could be used for a reimplementation is https://whispersystems.org/docs/specifications/xeddsa/#hashing-to-a-point-with-elligator-2.

jeffallen:
This is a good idea (and good design), and we hope to see
it come back into Kyber in the future.

However, the current implementation may have problems
and we do not want to be shipping code we are not 100%
confident in.

Fixes #117.


cretz:
Is there an issue to track when this might be reintroduced?

PR #353, opened 2019-01-17
Properly marshal/unmarshal zero G2 points

jeffallen:
After commit 09d88ea559cff, a round-trip encode/decode
of a zero point was failing. TestG2Marshal would have
discovered this 1 time in 2^256 runs, but in order to speed
up testing, add a new TestG2MarshalZero which always fails.

The solution is to follow the previous code patch from where a zero point
was detected, i.e. run p.g.SetInfinity(). This is still compatible
with golang.org/x/crypto/bn256, which also marshals infinity
as all zeros. (Checked this via printf-debugging on it.)

### DIFF
--- a/pairing/bn256/point.go
+++ b/pairing/bn256/point.go
@@ -1,6 +1,7 @@
 package bn256
 
 import (
+	"bytes"
 	"crypto/cipher"
 	"crypto/subtle"
 	"errors"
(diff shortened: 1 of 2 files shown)


kc1212:
Interesting, can you also check whether the java one marshals infinity G2 points as zeros?

jeffallen:
Yes:
https://github.com/dedis/cothority/blob/master/external/java/src/main/java/ch/epfl/dedis/lib/crypto/bn256/BN.java#L363

But then Java unmarshaling does not have any special case code, which is what I needed to add into Go to get it to behave correctly. Hmm.

jeffallen:
Hang on, I see a difference in Go and Java I want to explore more.

Issue #337, opened 2018-10-26
documentation: README.md


pschlump:
The test with "go test -v" (the first one in the readme) should be removed as there is no test file at the top level.

Output With Error:
```
go test -v
?   	github.com/dedis/kyber	[no test files]
```

Expected Output:

```
PASS
```

pierluca:
I believe the README has been updated since.
Closing this issue as part of our clean-up effort.
Thanks for reporting it.

PR #364, opened 2019-02-24
Remove test command as it does not have any usage #337

claucece:
This removes an instruction that does not work, as there are no test on the main part, as defined in issue #337 .


claucece:
Hi @jeffallen ! Thanks so much! Is it now able to be merged?

PR #365, opened 2019-02-25
Hash data to point properly

kc1212:
Previously BLS was using a naive way to hash data to a point which is
insecure. This work fixes the issue and adds a Hash interface to Point.

### DIFF
--- a/group.go
+++ b/group.go
@@ -110,6 +110,9 @@ type Point interface {
 	// Multiply point p by the scalar s.
 	// If p == nil, multiply with the standard base point Base().
 	Mul(s Scalar, p Point) Point
+
+	// Hash a message to a point.
+	Hash([]byte) Point
 }
 
 // AllowsVarTime allows callers to determine if a given kyber.Scalar
--- a/group.go
+++ b/group.go
@@ -110,9 +110,6 @@ type Point interface {
 	// Multiply point p by the scalar s.
 	// If p == nil, multiply with the standard base point Base().
 	Mul(s Scalar, p Point) Point
-
-	// Hash a message to a point.
-	Hash([]byte) Point
 }
 
 // AllowsVarTime allows callers to determine if a given kyber.Scalar
--- a/group/curve25519/ext.go
+++ b/group/curve25519/ext.go
@@ -253,6 +253,10 @@ func (P *extPoint) Mul(s kyber.Scalar, G kyber.Point) kyber.Point {
 	return P
 }
 
+func (P *extPoint) Hash(msg []byte) kyber.Point {
+	panic("not implemented")
+}
+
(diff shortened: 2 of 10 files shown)


kc1212:
I'm not sure what the best way to "organise" the hash function. The hash functionality needs to be implemented in the `bn256` package because it needs access to the internal `curvePoint` structure. OTOH the `bls` package treats the G1 point as `kyber.Group` so it cannot easily access the hash functionality in `bn256`. For this PR I added a new interface: `Point.Hash`. But the result is that we have to implement it for all the other curves otherwise we panic. Here are all the options I can think of:

- Add `Point.Hash` interface, describe above, downside is that we have to panic when we don't have an implementation
- Add `Suite.HashToG1Point` to the pairing suite, just like `Suite.Pair`, we treat it as a special thing for the pairing suite, but in theory the hash functionality is not specific to the pairing suite 
- Make `pointG1` public and cast it in the bls package

Gilthoniel:
The option 2 sounds better to me for the simple fact that we don't have an implementation for the others. It avoids all the panics.

What if you add the `PairingSuite.Hash` instead of _HashToG1Point_ so that if we need one day to have it implemented everywhere, and we simply add this function to the global `Suite` ? The one in the pairing suite would be compatible so no change to do.

kc1212:
`PairingSuite.Hash` would be confusing though, because `PairingSuite` has three point types `G1`, `G2` and `GT`. I suggested `HashToG1Point` because it's clear what kind of point it's using.

Gilthoniel:
I guess that would be fine for a beginning. In the worst case we can still add `Hash` later that uses `HashToG1Point` (in the case other suites must implement Hash)

kc1212:
Another solution is to create a small interface:
```
type hashablePoint struct {
  Hash(msg []byte) Point
}
```

`G1` will implement `hashablePoint`. Then we simply cast it to this interface when we need to hash something. The interface can stay private, this way none of the existing kyber API changes, which opens us to more options later on (e.g., in `v3.1+`) to add to the interface when we're sure that the API needs to be extended to support hashing to point.

kc1212:
I think if you don't update the kyber version in go.mod then this change won't affect onet or cothority

Gilthoniel:
Sure but when you will want to test the Java implementation, you will need to disable the JS ones until I fixed it I guess.

kc1212:
> Sure but when you will want to test the Java implementation, you will need to disable the JS ones until I fixed it I guess.

You're welcome to push to my branch - https://github.com/dedis/cothority/pull/1739 I don't think there will be any big change

Gilthoniel:
Very well then, let's wait for @Daeinar to confirm.

PR #370, opened 2019-03-22
Correct implementation of commonSuite

ineiti:
After my last try I understand better what is needed to implement the
`commonSuite` in a convenient way. This PR reverts the previous patch
and makes it so that a `bn256.NewSuite` cannot be used for `New` or `Read`
on a `Point` or `Scalar`. Only `bn256.NewSuiteG[12G]` can do so.

It also makes `NewSuiteG[1T]` a correct InterfaceMarshaler.

### DIFF
--- a/pairing/bn256/point.go
+++ b/pairing/bn256/point.go
@@ -12,7 +12,9 @@ import (
 	"go.dedis.ch/kyber/v3/group/mod"
 )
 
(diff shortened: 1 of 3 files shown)


ineiti:
Don't merge yet...

ineiti:
Yes, I have this one waiting: https://github.com/dedis/onet/pull/530

PR #372, opened 2019-04-09
Implement modified BLS signature algorithm

Gilthoniel:
In the context of the rogue public-key attack, we need to implement a more robust version of the BLS signature scheme that is using coefficient derived from the public keys.

This PR adds this new scheme without replacing the old one and it also adds a new mask implementation that will help the future implementation of the BLSCoSi protocol using the robust version of BLS. In short, the current mask implementation aggregates the public keys but we will need to specify how to aggregate.

Note: I was wrong, the current integer can handle a 128 bits one with the right modulus.

### DIFF
--- a/sign/asmbls/asmbls.go
+++ b/sign/asmbls/asmbls.go
@@ -1,7 +1,11 @@
-// Package bls2 implements the robust BLS scheme that prevents rogue public-key
-// to be used to forge signatures
+// Package asmbls implements the Accountable-Subgroup Multi-BLS scheme which is
+// a robust implement of the bls package against rogue public-key attacks. Those
+// attacks could allow an attacker to forge a public-key and then make a verifiable
+// signature for an aggregation of signatures. It basically fixes the situation by
+// adding coefficients to the aggregate.
+//
 // See the paper: https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html
-package bls2
+package asmbls
 
 import (
 	"crypto/cipher"
@@ -11,36 +15,54 @@ import (
 	"go.dedis.ch/kyber/v3"
 	"go.dedis.ch/kyber/v3/group/mod"
 	"go.dedis.ch/kyber/v3/pairing"
+	"go.dedis.ch/kyber/v3/sign"
 	"go.dedis.ch/kyber/v3/sign/bls"
-	"golang.org/x/crypto/blake2b"
+	"golang.org/x/crypto/blake2s"
 )
 
+// modulus128 can be provided to the big integer implementation to create numbers
+// over 128 bits
 var modulus128 = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 128), big.NewInt(1))
 
 // For the choice of H, we're mostly worried about the second preimage attack. In
-// other words, find H(m) == H(m')
+// other words, find m' where H(m) == H(m')
 // We also use the entire roster so that the coefficient will vary for the same
 // public key used in different roster
-func hashPointToR(point []byte, roster [][]byte) (kyber.Scalar, error) {
-	h, err := blake2b.New(16, nil)
+func hashPointToR(pubs []kyber.Point) ([]kyber.Scalar, error) {
+	peers := make([][]byte, len(pubs))
+	for i, pub := range pubs {
+		peer, err := pub.MarshalBinary()
+		if err != nil {
+			return nil, err
+		}
+
+		peers[i] = peer
+	}
+
+	h, err := blake2s.NewXOF(blake2s.OutputLengthUnknown, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	_, err = h.Write(point)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, peer := range roster {
+	for _, peer := range peers {
 		_, err := h.Write(peer)
 		if err != nil {
 			return nil, err
 		}
 	}
 
-	r := mod.NewIntBytes(h.Sum(nil), modulus128, mod.LittleEndian)
-	return r, nil
+	out := make([]byte, 16*len(pubs))
+	_, err = h.Read(out)
+	if err != nil {
+		return nil, err
+	}
+
+	coefs := make([]kyber.Scalar, len(pubs))
+	for i := range coefs {
+		coefs[i] = mod.NewIntBytes(out[i*16:(i+1)*16], modulus128, mod.LittleEndian)
+	}
+
+	return coefs, nil
 }
 
 // NewKeyPair creates a new BLS signing key pair. The private key x is a scalar
@@ -64,13 +86,13 @@ func Verify(suite pairing.Suite, x kyber.Point, msg, sig []byte) error {
 }
 
 // AggregateSignatures aggregates the signatures using a coefficient for each
-// one of them where c = H(pk) and H: G2 -> R{1, ..., 2^128}
-func AggregateSignatures(suite pairing.Suite, sigs [][]byte, mask *bls.Mask) (kyber.Point, error) {
+// one of them where c = H(pk) and H: G2 -> R with R = {1, ..., 2^128}
+func AggregateSignatures(suite pairing.Suite, sigs [][]byte, mask *sign.Mask) (kyber.Point, error) {
 	if len(sigs) != mask.CountEnabled() {
 		return nil, errors.New("length of signatures and public keys must match")
 	}
 
-	peers, err := marshalPublicKeys(mask.Publics())
+	coefs, err := hashPointToR(mask.Publics())
 	if err != nil {
 		return nil, err
 	}
@@ -84,18 +106,13 @@ func AggregateSignatures(suite pairing.Suite, sigs [][]byte, mask *bls.Mask) (ky
 			return nil, errors.New("couldn't find the index")
 		}
 
-		c, err := hashPointToR(peers[peerIndex], peers)
-		if err != nil {
-			return nil, err
-		}
-
 		sig := suite.G1().Point()
 		err = sig.UnmarshalBinary(buf)
 		if err != nil {
 			return nil, err
 		}
 
-		sigC := sig.Clone().Mul(c, sig)
+		sigC := sig.Clone().Mul(coefs[peerIndex], sig)
 		// c+1 because R is in the range [1, 2^128] and not [0, 2^128-1]
 		sigC = sigC.Add(sigC, sig)
 		agg = agg.Add(agg, sigC)
@@ -104,10 +121,11 @@ func AggregateSignatures(suite pairing.Suite, sigs [][]byte, mask *bls.Mask) (ky
 	return agg, nil
 }
 
-// AggregatePublicKeys aggregates the same way as for the signatures using
-// the same H: G2 -> R{1, ..., 2^128} as the hash function.
-func AggregatePublicKeys(suite pairing.Suite, mask *bls.Mask) (kyber.Point, error) {
-	peers, err := marshalPublicKeys(mask.Publics())
+// AggregatePublicKeys aggregates a set of public keys (similarly to
+// AggregateSignatures for signatures) using the hash function
+// H: G2 -> R with R = {1, ..., 2^128}.
+func AggregatePublicKeys(suite pairing.Suite, mask *sign.Mask) (kyber.Point, error) {
+	coefs, err := hashPointToR(mask.Publics())
 	if err != nil {
 		return nil, err
 	}
@@ -121,30 +139,11 @@ func AggregatePublicKeys(suite pairing.Suite, mask *bls.Mask) (kyber.Point, erro
 			return nil, errors.New("couldn't find the index")
 		}
 
-		c, err := hashPointToR(peers[peerIndex], peers)
-		if err != nil {
-			return nil, err
-		}
-
 		pub := mask.Publics()[peerIndex]
-		pubC := pub.Clone().Mul(c, pub)
+		pubC := pub.Clone().Mul(coefs[peerIndex], pub)
 		pubC = pubC.Add(pubC, pub)
 		agg = agg.Add(agg, pubC)
 	}
 
 	return agg, nil
-}
-
-func marshalPublicKeys(pubs []kyber.Point) ([][]byte, error) {
-	peers := make([][]byte, len(pubs))
-	for i, pub := range pubs {
-		peer, err := pub.MarshalBinary()
-		if err != nil {
-			return nil, err
-		}
-
-		peers[i] = peer
-	}
-
-	return peers, nil
 }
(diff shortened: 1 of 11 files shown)


sign/cosi/cosi.go
```diff
@@ -372,7 +381,7 @@ func AggregateMasks(a, b []byte) ([]byte, error) {
 // the operation relying on the collective signature is) in determining whether
 // the collective signature was produced by an acceptable set of cosigners.
 type Policy interface {
-	Check(m *Mask) bool
+	Check(m ParticipationMask) bool
```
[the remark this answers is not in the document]
Gilthoniel: The Mask in cosi is not deprecated because _ftcosi_ and derived are still using it.

Basically the main difference is that cosi.Mask is handling the aggregation when enabling/disabling keys but in the case of BLS, the aggregate can vary depending if you use the old or the new scheme so _bls.Mask_ is a subset of _cosi.Mask_ that only takes care of the participation mask.

Why _bls_ instead of _bls2_, because I use this mask to make the current protocol modular so that we can change the way it's aggregated. I was not sure if it would be worth to make another Mask type in _bls2_ that is using the _bls_ one.

Gilthoniel:
Going back to WIP because I need to change the aggregate functions to support aggregation of a subset of signatures where the full set of public keys is still needed.

sign/bls/bls.go
```diff
@@ -1,6 +1,13 @@
 // Package bls implements the Boneh-Lynn-Shacham (BLS) signature scheme which
 // was introduced in the paper "Short Signatures from the Weil Pairing". BLS
 // requires pairing-based cryptography.
+//
+// Deprecated: This version is vulnerable to rogue public-key attack and the
+// new version of the protocol should be used to make sure a signature
+// aggregate cannot be verified by a forged key. You can find the protocol
+// in kyber/sign/bls2
+//
+// See the paper: https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html
```
Daeinar: Hmm, I think this comment should be reformulated. Using regular BLS signing without aggregation is perfectly fine and secure.  Meaning we only become a problem when regular BLS signatures are aggregated. I would probably remove all of the BLS aggregation code from this package, only expose regular BLS signing and verification, and then also remove the "deprecated flag".

Gilthoniel: Keep in mind we need to stay backwards compatible for the moment. Eventually we won't have two packages but one. That's only for v4 though.

sign/bls2/bls2.go
```diff
@@ -0,0 +1,150 @@
+// Package bls2 implements the robust BLS scheme that prevents rogue public-key
```
Daeinar: I would probably rename this package to something that is more "meaningful" as the 2 in BLS2 doesn't tell me much. Maybe one of `AggregateBLS`, `AggBLS`, `CoBLS`, `MultiSigBLS`, `MultiBLS`, `MBLS` (short form of MultiBLS)?

sign/bls/mask.go
```diff
@@ -0,0 +1,172 @@
+package bls
```
Daeinar: Shouldn't all of this code go into the new BLS package that supports secure signature aggregation?

sign/bls2/bls2.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+var modulus128 = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 128), big.NewInt(1))
+
+// For the choice of H, we're mostly worried about the second preimage attack. In
+// other words, find H(m) == H(m')
+// We also use the entire roster so that the coefficient will vary for the same
+// public key used in different roster
+func hashPointToR(point []byte, roster [][]byte) (kyber.Scalar, error) {
+	h, err := blake2b.New(16, nil)
```
Daeinar: Out of curiosity: Is there a specific reason why BLAKE2 instead of SHA{256,512} is used here? The current IETF draft standard seems to suggest usage of SHA512: https://www.ietf.org/id/draft-boneh-bls-signature-00.txt (but it also uses the curve BLS12-381)

sign/bls2/bls2.go
```diff
@@ -0,0 +1,150 @@
...
+	"math/big"
+
+	"go.dedis.ch/kyber/v3"
+	"go.dedis.ch/kyber/v3/group/mod"
+	"go.dedis.ch/kyber/v3/pairing"
+	"go.dedis.ch/kyber/v3/sign/bls"
+	"golang.org/x/crypto/blake2b"
+)
+
+var modulus128 = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 128), big.NewInt(1))
```
Daeinar: Can you please add a comment on the meaning of this modulus?

sign/bls2/bls2.go
```diff
@@ -0,0 +1,150 @@
...
+// Verify checks the given BLS signature S on the message m using the public
+// key X by verifying that the equality e(H(m), X) == e(H(m), x*B2) ==
+// e(x*H(m), B2) == e(S, B2) holds where e is the pairing operation and B2 is
+// the base point from curve G2.
+func Verify(suite pairing.Suite, x kyber.Point, msg, sig []byte) error {
+	return bls.Verify(suite, x, msg, sig)
+}
+
+// AggregateSignatures aggregates the signatures using a coefficient for each
+// one of them where c = H(pk) and H: G2 -> R{1, ..., 2^128}
```
Daeinar: `H: G2 -> R with R = {1,...,2^128}`

sign/bls2/bls2.go
Gilthoniel: I followed the Stanford paper where they're saying the hash function should map to the set [1; 2^128] so I took BLAKE2 because of the length of the digest. (and supposedly better than MD5)

sign/bls2/bls2.go
```diff
@@ -0,0 +1,150 @@
...
+		// c+1 because R is in the range [1, 2^128] and not [0, 2^128-1]
+		sigC = sigC.Add(sigC, sig)
+		agg = agg.Add(agg, sigC)
+	}
+
+	return agg, nil
+}
+
+// AggregatePublicKeys aggregates the same way as for the signatures using
+// the same H: G2 -> R{1, ..., 2^128} as the hash function.
```
Daeinar:
```
AggregatePublicKeys aggregates a set of public keys (similarly to AggregateSignatures for signatures) 
using the hash function H: G2 -> R with R = {1,...,2^128}.
```

sign/bls/bls.go
Daeinar: Okay, then perhaps make it clear that only the signature aggregation is broken/deprecated (and will be phased out in a future version of kyber). The regular BLS signing and verification is still fine (and thus will remain in kyber).

sign/bls2/bls2.go
Daeinar: According to the BLAKE2b documentation the output should be at least 32 bytes, i.e., 256 bits, if BLAKE2b is used as a hash function (see `New` in https://godoc.org/golang.org/x/crypto/blake2b) to get to the minimum security level of 128 bit.

It's not clear to me how this maintains a 128-bit security level, as BLAKE2b (or any hash function fwiw) achieves only 64 bits of security for a 128-bit output (due to the birthday bound).

Gilthoniel: My reasoning behing this, and pls correct me if I'm wrong, is that we want to be safe against a 2nd preimage attack because the attacker needs to find `t*` that is equal to `t` where `t` is the coefficient of the honest public key. Isn't the birthday attack only right for the collision resistance ?

In any case, finding a 2nd preimage doesn't mean you can forge the public key because you still need to find the private key associated with it: `g^alpha * (1/pk) == 2nd preimage` meaning you still need to solve the DL problem.

sign/bls2/bls2.go
Daeinar: Maybe call it `ASMBLS` for simply `ASM` for Accountable-Subgroup Multi-Signatures as defined in https://eprint.iacr.org/2018/483.pdf ?

Daeinar:
Overall looks good to me modulo some other changes discussed offline with @Gilthoniel (e.g., using [BLAKE2b in XOF mode](https://godoc.org/golang.org/x/crypto/blake2b#XOF) to compute `t1 || ... || tn = XOF(pk1 || ... || pkn)` instead of deriving the scalars via domain-separated iterated hashing as `ti = BLAKE2(pki || pk1 || ... || pkn)`).

sign/bls/mask.go
Gilthoniel: Moved it to `sign` as suggested offline

Gilthoniel:
@Daeinar comments are fixed. I went for BLAKE2s instead of b because we have only an implementation of this one in Javascript.

sign/asmbls/asmbls.go
```diff
@@ -0,0 +1,149 @@
+// Package asmbls implements the Accountable-Subgroup Multi-BLS scheme which is
+// a robust implement of the bls package against rogue public-key attacks. Those
+// attacks could allow an attacker to forge a public-key and then make a verifiable
+// signature for an aggregation of signatures. It basically fixes the situation by
```
jeffallen:
```suggestion
// signature for an aggregation of signatures. It fixes the situation by
```

sign/asmbls/asmbls.go
```diff
@@ -0,0 +1,149 @@
+// Package asmbls implements the Accountable-Subgroup Multi-BLS scheme which is
+// a robust implement of the bls package against rogue public-key attacks. Those
```
jeffallen:
```suggestion
// an implementation of the bls package which is robust against rogue public-key attacks. Those
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
+// Package sign contains useful tools for the different signin algorithms
```
jeffallen:
```suggestion
// Package sign contains useful tools for the different signing algorithms.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// Publics returns a copy of the list of public keys
+func (m *Mask) Publics() []kyber.Point {
+	pubs := make([]kyber.Point, len(m.publics))
+	copy(pubs, m.publics)
+	return pubs
+}
+
+// Participants returns the list of public keys participating
```
jeffallen:
```suggestion
// Participants returns the list of public keys participating.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+	return m.forEachBitEnabled(func(i, j, n int) int {
+		if i*8+int(j) == idx {
+			return n
+		}
+
+		return -1
+	})
+}
+
+// Publics returns a copy of the list of public keys
```
jeffallen:
```suggestion
// Publics returns a copy of the list of public keys.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+
+				n++
+			}
+		}
+	}
+
+	return -1
+}
+
+// IndexOfNthEnabled returns the index of the nth enabled bit or -1 if out of bound
```
jeffallen:
```suggestion
// IndexOfNthEnabled returns the index of the nth enabled bit or -1 if out of bounds.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+	if enable {
+		m.mask[byteIndex] ^= mask
+	} else {
+		m.mask[byteIndex] ^= mask
+	}
+	return nil
+}
+
+// forEachBitEnabled is a helper to iterate over the bits set to 1 in the mask
+// and to return the result of the callback only if it is positive
```
jeffallen:
```suggestion
// and to return the result of the callback only if it is positive.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+func (m *Mask) SetMask(mask []byte) error {
+	if m.Len() != len(mask) {
+		return fmt.Errorf("mismatching mask lengths")
+	}
+
+	m.mask = mask
+	return nil
+}
+
+// SetBit turns on or off the bit at the given index
```
jeffallen:
```suggestion
// SetBit turns on or off the bit at the given index.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+	copy(clone[:], m.mask)
+	return clone
+}
+
+// Len returns the length of the byte array necessary to store the bitmask
+func (m *Mask) Len() int {
+	return (len(m.publics) + 7) / 8
+}
+
+// SetMask replaces the current mask by the new one if the length matches
```
jeffallen:
```suggestion
// SetMask replaces the current mask by the new one if the length matches.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// Mask returns the bitmask as a byte array
+func (m *Mask) Mask() []byte {
+	clone := make([]byte, len(m.mask))
+	copy(clone[:], m.mask)
+	return clone
+}
+
+// Len returns the length of the byte array necessary to store the bitmask
```
jeffallen:
```suggestion
// Len returns the length of the byte array necessary to store the bitmask.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+		}
+
+		return nil, errors.New("key not found")
+	}
+
+	return m, nil
+}
+
+// Mask returns the bitmask as a byte array
```
jeffallen:
```suggestion
// Mask returns the bitmask as a byte array.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+// Mask is a bitmask of the participation to a collective signature
+type Mask struct {
+	mask    []byte
+	publics []kyber.Point
+}
+
+// NewMask creates a new mask from a list of public keys. If a key is provided, it
+// will set the bit of the key to 1 or return an error if it is not found
```
jeffallen:
```suggestion
// will set the bit of the key to 1 or return an error if it is not found.
```

sign/mask.go
```diff
@@ -0,0 +1,173 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"go.dedis.ch/kyber/v3"
+	"go.dedis.ch/kyber/v3/pairing"
+)
+
+// Mask is a bitmask of the participation to a collective signature
```
jeffallen:
```suggestion
// Mask is a bitmask of the participation to a collective signature.
```

sign/cosi/cosi.go
```diff
@@ -225,6 +225,15 @@ func Verify(suite Suite, publics []kyber.Point, message, sig []byte, policy Poli
 	return nil
 }
 
+// ParticipationMask is an interface to get the total number of candidates
+// and the number of participants
```
jeffallen:
```suggestion
// and the number of participants.
```

PR #383, opened 2019-04-30
Embedding on bn256

taybart:
I added embedding on G1 for the bn256 package. I think a discussion about the adapters is required since I don't think what I have done here will maintain compatibility, namely, the adapter is returning a G1 for Point() and not a G2.


Gilthoniel:
@taybart why is the change in the adapter necessary ? Do you need to create G1 points ? The adapter is just a hacky way to fit the Suite interface but you have access to a pairing suite that let you do something like `suite.G1().Point()`.

taybart:
@Gilthoniel I was confused about how to actually use the kyber suite when I started. Sorry about that, the hack is removed!

taybart:
Does it look OK otherwise? I am using this code to test it (mostly lifted from the ElGamal example in ed25519):

```go
package main

import (
	"fmt"

	"go.dedis.ch/kyber/v3"
	"go.dedis.ch/kyber/v3/pairing"
	"go.dedis.ch/kyber/v3/util/random"
)

func elGamalEncrypt(suite pairing.Suite, pubkey kyber.Point, message []byte) (
	K, C kyber.Point, remainder []byte) {

	// Embed the message (or as much of it as will fit) into a curve point.
	M := suite.G1().Point().Embed(message, random.New())
	max := suite.G1().Point().EmbedLen()
	if max > len(message) {
		max = len(message)
	}
	remainder = message[max:]
	// ElGamal-encrypt the point to produce ciphertext (K,C).
	k := suite.G1().Scalar().Pick(random.New()) // ephemeral private key
	K = suite.G1().Point().Mul(k, nil)          // ephemeral DH public key
	S := suite.G1().Point().Mul(k, pubkey)      // ephemeral DH shared secret
	C = suite.G1().Point().Add(S, M)            // message blinded with secret
	return
}

func elGamalDecrypt(suite pairing.Suite, prikey kyber.Scalar, K, C kyber.Point) (
	message []byte, err error) {

	// ElGamal-decrypt the ciphertext (K,C) to reproduce the message.
	S := suite.G1().Point().Mul(prikey, K) // regenerate shared secret
	M := suite.G1().Point().Sub(C, S)      // use to un-blind the message
	message, err = M.Data()                // extract the embedded data
	return
}

func main() {
	suite := pairing.NewSuiteBn256()

	// Create a public/private keypair
	a := suite.G1().Scalar().Pick(suite.RandomStream()) // Alice's private key
	// a := suite.Scalar().Pick(random.New()) // Alice's private key
	A := suite.G1().Point().Mul(a, nil) // Alice's public key

	// ElGamal-encrypt a message using the public key.
	m := []byte("The quick brown fox")
	// m := []byte("The")
	K, C, _ := elGamalEncrypt(suite, A, m)

	// Decrypt it using the corresponding private key.
	mm, err := elGamalDecrypt(suite, a, K, C)

	// Make sure it worked!
	if err != nil {
		fmt.Println("decryption failed: " + err.Error())
	} else if string(mm) != string(m) {
		fmt.Println("decryption produced wrong output: " + string(mm))
	} else {
		fmt.Println("Decryption succeeded: " + string(mm))
	}
}
```

taybart:
I included `examples/bn256_enc_test.go`. Not sure what you mean about the test being weak. I am willing to add more tests if you point me in the right direction.

Gilthoniel:
Sorry I wasn't clear enough. I was talking about the test that makes sure a point is actually on the curve (you subtract the point and check if it is null). Another guy will have a look at that before we move forward.

taybart:
Sounds good, I didn't leave in the check that the point is on the curve because it is by definition of calculation, I couldn't find a good source to check if the point is in the group or not. The other method I saw was to multiply the point by the order of the group and check if it is at infinity.

pairing/bn256/point.go
```diff
@@ -62,7 +62,34 @@ func (p *pointG1) Clone() kyber.Point {
 }
 
 func (p *pointG1) EmbedLen() int {
-	panic("bn256.G1: unsupported operation")
+	return (255 - 8 - 8) / 8
+}
+
+func gfpToBig(gfp *gfP) *big.Int {
```
Gilthoniel: Could you move this as a function of `gfP` ? So you can do do `curveB.BigInt()`

pairing/bn256/point.go
```diff
@@ -62,7 +62,34 @@ func (p *pointG1) Clone() kyber.Point {
...
+	bigInt := new(big.Int)
+	decoded := new(gfP)
+	montDecode(decoded, gfp)
+	buf := make([]byte, 32)
+	decoded.Marshal(buf)
+	bigInt.SetBytes(buf)
+	return bigInt
+}
+
+func bigToGfp(bigX *big.Int) *gfP {
```
Gilthoniel: Could you move this to a function in `gfp.go` ? Something like `newGFpFromBigInt(...)`

pairing/bn256/point.go
```diff
@@ -72,11 +99,61 @@ func (p *pointG1) Embed(data []byte, rand cipher.Stream) kyber.Point {
...
+
+		t := new(big.Int).Add(xxx, intCurveB)
+		y := new(big.Int).ModSqrt(t, Order)
+		if y != nil {
+			p.g.x = *bigToGfp(x)
+			p.g.y = *bigToGfp(y)
+			p.g.z = *newGFp(1)
+			q := newPointG1()
+			t := q.Sub(p, p)
+			if t.Equal(q.Null()) {
```
Gilthoniel: You can actually use the already implemented function `p.g.IsOnCurve()`

pairing/bn256/point.go
```diff
@@ -62,7 +62,34 @@ func (p *pointG1) Clone() kyber.Point {
 }
 
 func (p *pointG1) EmbedLen() int {
-	panic("bn256.G1: unsupported operation")
+	return (255 - 8 - 8) / 8
```
Gilthoniel: Can you add a comment explaining the length ?

taybart: Added what I thought was a reasonable comment.

pairing/bn256/point.go
taybart: I added this and realized Order was the incorrect number to modulo operations by so its in mod P now. I also had to change the constant p to P because the receivers on those points is p.

pairing/bn256/point.go
taybart: Good idea! It should be included now.

pairing/bn256/point.go
taybart: Another good idea, this should also be in.

pairing/bn256/point.go
```diff
@@ -62,21 +62,65 @@ func (p *pointG1) Clone() kyber.Point {
...
+		rand.XORKeyStream(b[:], b[:])
+		if data != nil {
+			b[0] = byte(dl)       // Encode length in low 8 bits
+			copy(b[1:1+dl], data) // Copy in data to embed
+		}
+		x := new(big.Int).SetBytes(b[:])
+
+		xxx := new(big.Int).Mul(x, x)
+		xxx.Mul(xxx, x)
+		xxx.Mod(xxx, P)
```
Gilthoniel: I would prefer to keep that prime unchanged, also you can see that this piece of code is similar to `hashToPoint` (well it's doing the same thing basically). One simple trick to avoid the prime `p` to change is to declare a function `deriveY`:

```go
// deriveY compute the y coordinate of x for x^3 + b = y^2
func deriveY(x *big.Int, curveB *big.Int) (*big.Int) {
  xxx := new(big.Int).Mul(x, x)
  xxx.Mul(xxx, x)
  xxx.Mod(xxx, p)

  t := new(big.Int).Add(xxx, intCurveB)
  return new(big.Int).ModSqrt(t, p)
}
```

You can then hook that up in your Embed and in the hashToPoint.

taybart: You da boss! I thought about doing it like this but decided against it for some reason. Should be in now

taybart:
Any word on this?

Gilthoniel:
One last thing I need to ask you is to sign this [CLA](https://github.com/dedis/cothority/blob/master/CLAI). This is required because of our partnership with the Industry. If you agree to sign, you can send it to me at gaylor.bosson[at]epfl.ch.

armfazh:
The method implemented in this PR doesnt run in constant time, and could bring security issues in the future if not used appropriately. (for example, Dragonbleed attack on wpa3.0)

It would be better to implement a constant-time method, for example using the one in https://godoc.org/github.com/cloudflare/bn256#HashG1
which is aligned to the WIP specification of hash to curve Internet standard.

taybart:
@armfazh embedding a hash on a curve kinda defeats El Gamal right? The data needs to be recoverable and hashing it would destroy the original message

armfazh:
> @armfazh embedding a hash on a curve kinda defeats El Gamal right? The data needs to be recoverable and hashing it would destroy the original message

Right, the hash to curve is a one-way function, for ElGamal, it's required a reversible function.

taybart:
Would it be possible to get this merged?

pierluca:
Hi @taybart , 
Sorry for the long delay. As you can see, a number of issues and PRs have remained open in the last months without much response. This was due to a temporary organizational situation. That's changing, so we're going to resume addressing the issues, reviewing and merging the PRs, etc.  Please hang in there, we'll get back to you in the coming weeks.
Thank you for your understanding.

ineiti:
> Would it be possible to get this merged?

As a first step, please make sure that all tests pass.

taybart:
@ineiti Do you have the SONAR_TOKEN? Is that being set with github secrets? Sonarcloud scan is the thing that is failing

ineiti:
Hmm - that would be for @nkcr or @pierluca to know - any idea?

taybart:
I could merge this into a branch in this org/repo and then make a new pr with that branch? That should make the secrets available

pierluca:
According to GitHub:
> Secrets are not passed to workflows that are triggered by a pull request from a fork.

We use secrets to provide the Sonar token to the GitHub Action.

For many years, there was no solution, but apparently we should be able to do this now:
https://community.sonarsource.com/t/how-to-use-sonarcloud-with-a-forked-repository-on-github/7363/30

I haven't personally got time to look into improving the GitHub Actions configuration right now.
Anyone keen to fix this ? @ineiti @nkcr @jbsv

ineiti:
What @taybart proposes is perhaps the quickest solution, until somebody has the time to do it correctly: create a branch on this repo with the content of @taybart 's master branch, and then create a PR with that.

taybart:
If you make a branch I can update the PR for that new branch

ineiti:
Sorry it took so long. I added a new PR here: #475 - can you look if that is the correct, latest version?

Shall we close this PR?

taybart:
Thanks!

Issue #377, opened 2019-04-16
Proposal: kyber.Group wrapper for libsecp256k1


coventry:
I'm thinking of wrapping `libsecp256k1` in the `kyber.Group` interface. I would use essentially [the same structure as `go-ethereum`](https://github.com/ethereum/go-ethereum/tree/master/crypto/secp256k1) for this.

1. Would this be a welcome addition to `kyber`?
2. If there are any reasons why this is an obviously bad idea, I'd be glad to learn of them. My plan is to use it with the `share/dkg` and `share/dss` packages for ethereum crypto, where it's possible to do reasonably cheap Schnorr signatures. The Alt-BN128 curve, which the EVM precompiled contracts target, won't work for my purposes, and I can use `ECRECOVER` to do the arithmetic I need.

jeffallen:
We welcome pull requests, but I cannot guarantee that this will be merged.

First, we need to respect a backwards compatibility commitment to those using kyber v3. If secp256k1 needs modifications to the main abstractions of kyber (Scalar and Point), that's a no-go. I would be open to discussing making Kyber v4 to support this, but it would need to be pretty compelling to justify that.

Next, we would like Kyber to remain pure Go, to make sure that building is still fast and easy for users (many of whom are students who are new to programming and new to Go). Thus other secp256k1's might be better choices:
* https://github.com/tendermint/tendermint/tree/master/crypto/secp256k1
* https://github.com/btcsuite/btcd/tree/master/btcec

Kyber will never be the fastest crypto library, but it should be fast enough while still being very easy to work with the prototype and develop novel applied crypto projects.

Finally, we'll need to look at license compatibility. You'll need to sign a CLAI or CLAC so that your contribution can be licensed like the rest of Kyber. (See https://github.com/dedis/cothority/blob/master/CONTRIBUTION) If your submitted code is simply a wrapper to adapt a dependency like tendermint's implementation to the Kyber API, easy. If you need to make modifications to the secp256k1 implementation in order to get access to the internals (we had to do this for both ed25519 and bn256), then you'll be forking their code and checking it into Kyber, in which case their code needs have a license that is compatible with Kyber's MPLv2 license.

coventry:
[...]

PR #385, opened 2019-05-17
Wrap btcec's secp256k1 implementation for kyber.{Group,Point,Scalar}

coventry:
This is a go-only, variable time implementation of the `kyber.{Group,Point,Scalar}` interfaces for secp256k1 (bitcoin/ethereum curve.) Please see issue #377 for discussion.

With minor changes to the test suites, this passes the `vss/rabin`, `dkg/rabin`, and `dss` tests. I've left those changes out, but included some adjustments to `vss_test.go` which helped me when debugging.


go.mod
```diff
@@ -1,6 +1,7 @@
 module go.dedis.ch/kyber/v3
 
 require (
+	github.com/btcsuite/btcd v0.0.0-20190410025418-9bfb2ca0346b
```
jeffallen: btcd is a rather large dependency, since all we really want is btcec. Their license (the ISC license) would allow us to fork their code, and bring in btcec directly into Kyber if we wanted. btcec is fairly stable, having received just a couple commits in 2 years.

Did you consider incorporating btcec as source instead of as a dependency?

coventry: No, I hadn't, but I'd be happy to do that if the licence allows, and expect it would be a straightforward modification.

jeffallen: Yes, please.

group/secp256k1/curve.go
```diff
@@ -0,0 +1,41 @@
+// Package secp256k1 is an implementation of the kyber.{Group,Point,Scalar}
+// interfaces, based on btcd/btcec and kyber/group/mod
+//
+// XXX: NOT CONSTANT TIME!
+package secp256k1
+
+import (
+	"math/big"
+
+	secp256k1BTCD "github.com/btcsuite/btcd/btcec"
```
jeffallen: Why do you rename the package here? Anyway, with the inclusion of the btcec code, this will all be local and not imported.

You should include the minimum of the btcec code you can, for example not ciphering.go, etc. Probably just ScalarBaseMult and ScalarMult and their dependencies, including the pre-computed point table. Furthermore, you can leave gensecp256k1.go and genprecomps.go out, just change the comment in secp256k1.go to explain where it came from (i.e. the real btcec package).

group/secp256k1/suite.go
```diff
@@ -0,0 +1,76 @@
+// Package secp256k1 is an implementation of the kyber.{Group,Point,Scalar}
```
jeffallen: This suite should be registered in package go.dedis.ch/kyber/v3/suites.

group/secp256k1/scalar.go
```diff
@@ -0,0 +1,195 @@
...
+// Pick sets s to a random value mod groupOrder sampled from rand, and returns
+// it
+func (s *secp256k1Scalar) Pick(rand cipher.Stream) kyber.Scalar {
+	return s.Set((*secp256k1Scalar)(random.Int(groupOrder, rand)))
+}
+
+// MarshalBinary returns the big-endian byte representation of s, or an error on
+// failure
+func (s *secp256k1Scalar) MarshalBinary() ([]byte, error) {
+	b := toInt(s.modG()).Bytes()
```
jeffallen: Why is that modG in there? It has the consequence that a MarshalBinary can change the underlying data, which is surprising to callers. Compare to pairing/bn256/point.go, where MarshalBinary uses Clone when it knows it's about to call MakeAffine and modify the point.

group/secp256k1/scalar_test.go
```diff
@@ -0,0 +1,127 @@
+package secp256k1
```
jeffallen: I was hoping to see test cases that compare operations in this library to those in other libraries. Can you find out if there are a standard set of inputs and outputs that other secp256k1 implementations use to check compatibility between each other? I'm not saying that we have to have this, just that if such a thing is normally used, then we should be using it as well. Or you could be the first: generate a few scalars and points using libsecp256k1 and then confirm that the same results come from this library.

group/secp256k1/scalar_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"encoding/hex"
+	"math/big"
```
jeffallen: group/nist/curve.go uses group/mod to do scalar modular math, why did you decide not to use it? I have to say, I'm not totally clear on the difference, what would be the benefits, etc. But I generally understand that the SECG group's curves are like the NIST curves, but with constants that are more transparent ("no numbers up my sleeves"), and so I'm wondering why the implementation in this directory would be different from ../nist.

group/secp256k1/scalar_test.go
coventry: I don't think there's anything standard, though of course there are test vectors we could borrow from other suites: [Python's cryptography package](https://github.com/pyca/cryptography/blob/c574e7554c7aa27c56f6478258a4e18f79457652/vectors/cryptography_vectors/asymmetric/ECDSA/SECP256K1/SigGen.txt), [some random vectors](https://chuckbatson.wordpress.com/2014/11/26/secp256k1-test-vectors/).

group/secp256k1/scalar_test.go
coventry: I just didn't see much value in the abstraction group/mod provided. Plus it carries two unnecessary fields.

group/secp256k1/scalar.go
coventry: This is defensive programming, since a secp256k1 should already be reduced mod the groupOrder. I can take a clone first, though.

group/secp256k1/curve.go
coventry: I just found the name btcec obscure.

coventry:
Thanks for the feedback. I'll revise accordingly.

jeffallen:
Ping?

coventry:
Sorry, Jeff. It's been hard to find time to work on this. The folks [I've been developing it for](https://github.com/smartcontractkit/chainlink/pull/1272) don't see the value in ripping the minimal subset of btcec out of btcd.

jeffallen:
OK, understand. But we cannot accept a giant growth in dependencies for this. Kyber is used in systems that need to stay as small and tight as possible. I am closing for now, please re-open if you find a way to move forward on this.

Thanks again for your interest in Kyber.

PR #386, opened 2019-05-31
Fix ed25519 secret key generation w.r.t multiple of 8

### LINKED ISSUE #351: [edwards25519] NewKey related (potential) issue
NewKey (part of key generator interface) generates secrets of the correct form
(lying in 2^254 + 8{0,1,2,...,2^251-1}) before calling scalar.SetBytes to retrieve the scalar.
However SetBytes reduces modulo q (order of subgroup generated by base point (9,)).
Since q=2^252+... < 2^254 IMHO we just destroy what we have done.
(or if turns out that deeper understanding shows this is not an issue, deserve some documentation)

Gilthoniel:
This fixes the generation by making sure the scalar is not reduced
modulo the prime of the field.

Fixes #351

### DIFF
--- a/group/edwards25519/curve.go
+++ b/group/edwards25519/curve.go
@@ -45,16 +45,32 @@ func (c *Curve) Point() kyber.Point {
 	return P
 }
 
+// NewKeyAndSeedWithInput returns a formatted Ed25519 key (avoid subgroup attack by
+// requiring it to be a multiple of 8). It also returns the input and the digest used
+// to generate the key.
+func (c *Curve) NewKeyAndSeedWithInput(buffer []byte) (kyber.Scalar, []byte, []byte) {
+	digest := sha512.Sum512(buffer[:])
+	digest[0] &= 0xf8
+	digest[31] &= 0x7f
+	digest[31] |= 0x40
+
+	secret := c.Scalar().(*scalar)
+	copy(secret.v[:], digest[:])
+	return secret, buffer, digest[32:]
+}
+
+// NewKeyAndSeed returns a formatted Ed25519 key (avoid subgroup attack by requiring
+// it to be a multiple of 8). It also returns the seed and the input used to generate
+// the key.
+func (c *Curve) NewKeyAndSeed(stream cipher.Stream) (kyber.Scalar, []byte, []byte) {
+	var buffer [32]byte
+	random.Bytes(buffer[:], stream)
+	return c.NewKeyAndSeedWithInput(buffer[:])
+}
+
 // NewKey returns a formatted Ed25519 key (avoiding subgroup attack by requiring
 // it to be a multiple of 8). NewKey implements the kyber/util/key.Generator interface.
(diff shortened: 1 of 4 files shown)


ineiti:
This is a tricky one. It seemed to me (not sure), that in `kyber`, every step already returns a reduced scalar. Other libraries handle this differently, this is why in the java-kyber, I added a reduction step after each operation.

This is very important whenever scalars are used in a hash, as the reduced scalar and the non-reduced scalars will produce different hashes.

I'm not aware of a place where the dedis/kyber library explicitly states that it always works on reduced scalars, but from what I know, it does always reduce the scalars, and should state this clearly. Or you would need to make sure that everywhere, where scalars are used in hashes, they have been reduced before.

Is there a way to create reduced scalars in kyber? `scalar.Add(scalar, 0)`?

So I'm afraid that this PR might break some of the hashes somewhere, but sometimes work. Can you please verify if that is the case or not? I think of cosi-signing and schnorr-signing, but others might apply, too.

Gilthoniel:
The secret is still an `ed25519` scalar so any operation on it will reduce it to `q` so Schnorr and CoSi signatures will basically remain equivalent. `EdDSA` is using the prefix (in the contrary of the secret key in other implementations...) which is then not reduced.

I'm actually wondering if we should have a different key generation for `EdDSA` versus `Schnorr` and `CoSi` because the lastest require a secret key between [1;q[.

Gilthoniel:
@Daeinar any advice on this ?

Gilthoniel:
Moving on on this, it seems to me that this fixes the issue and doesn't break any compatibility as the response computed during the signing is reduced anyway. The only thing that is going to change is the resulting secret key.

The interesting fact is that `s = (h*s + r) mod q` which is equivalent to `s = (h mod q)*(s mod q) + (r mod q)` so the actual issue was not an issue as far as I understand correctly.

kc1212:
The problem that this PR tries to solve is that `SetBytes` is doing reduction maybe when it shouldn't. I'm wondering if there's something more fundamental to fix. The RFC https://tools.ietf.org/html/rfc7748#section-5 has this `decodeScalar25519` function which doesn't appear to do any kind of reduction. 

From the description:
```
   For X25519, in
   order to decode 32 random bytes as an integer scalar, set the three
   least significant bits of the first byte and the most significant bit
   of the last to zero, set the second most significant bit of the last
   byte to 1 and, finally, decode as little-endian.  This means that the
   resulting integer is of the form 2^254 plus eight times a value
   between 0 and 2^251 - 1 (inclusive).
```

Gilthoniel:
After meeting with @Daeinar we concluded that there was no issue at the first place because the secret is reduced modulo q in the signing process anyway. We were just doing the reduction one step earlier than the actual RFC but it is producing the same result.

One test that I'm going to write out of curiosity is about the response that is supposed to have three leading zeros for the most significant bytes.

Gilthoniel:
The PR is fixing the bit twidling in the eddsa file (0x7f instead of 0x3f) and is cleaning the code a bit.

PR #390, opened 2019-09-10
New randomStream struct to allow use of user-specified entropy source

PizzaWhisperer:
MixedRandStream holds an array of `io.Reader` and creates the random stream by reading from each source and applying XOF from blake2xb.
Perdersen's dkg code is modified to use it. In the dkg config there are two new fields: 
- `reader` which may hold the user's random data. It is not mandatory for the user to specify one, but if they did, it will be used as additional entropy for the generation of the dkg's secret coefficient
- and a boolean `userReaderOnly` which forces the use of the user's random data only (by default the user's data is always combined with randomness from `crypto/rand`). This is useful for reproducibility and debugging purposes


share/dkg/pedersen/dkg.go
```diff
@@ -10,11 +10,14 @@
...
 import (
+	"crypto/cipher"
+	"crypto/rand"
 	"errors"
 	"fmt"
+	"io"
 
-	"go.dedis.ch/kyber/v3"
-
+	"github.com/PizzaWhisperer/kyber/util/random"
```
nikkolasg: That's the thing with Go, it's not easy to fork ^^ Normally with the Go module I think you should be able to use the kyber path...
If it's not working, then try to do 2 PR: first for that random package and then for DKG

share/dkg/pedersen/dkg.go
```diff
@@ -222,12 +243,14 @@ func NewDistKeyHandler(c *Config) (*DistKeyGenerator, error) {
 
 // NewDistKeyGenerator returns a dist key generator ready to create a fresh
 // distributed key with the regular DKG protocol.
-func NewDistKeyGenerator(suite Suite, longterm kyber.Scalar, participants []kyber.Point, t int) (*DistKeyGenerator, error) {
+func NewDistKeyGenerator(suite Suite, longterm kyber.Scalar, participants []kyber.Point, t int, r io.Reader, userOnly bool) (*DistKeyGenerator, error) {
```
nikkolasg: Let's keep the API as it is for now, we don't need to make API-breaking changes for this. This function will use the default values.

share/dkg/pedersen/dkg_test.go
```diff
@@ -3,11 +3,15 @@ package dkg
 import (
 	"crypto/rand"
 	"fmt"
+	"io"
+  "io/ioutil"
```
nikkolasg: formatting

util/random/rand.go
```diff
@@ -76,3 +81,51 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
+	// create the XOF output, with hash of collected data as seed
+	h := sha512.New()
+	h.Write(b.Bytes())
+	seed := h.Sum(nil)
+	hash := blake2xb.New(seed)
+	hash.XORKeyStream(dst, src)
+}
+
+// NewMixedStream returns a new cipher.Stream that gets random data from the specified
+// readers. If no reader was given, Go's crypto/rand package is used.
```
nikkolasg: Can you add more comments as to what it does specifically (mixing entropy sources via hashing and then using blake2 to do the XOR)

util/random/rand.go
```diff
@@ -76,3 +81,51 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
+	}
+
+	// we are ok with few sources being insecure (i.e., providing less than
+	// READER_BYTES bytes), but not all of them
+	if nerr == len(r.Readers) {
+		panic("all readers failed")
+	}
+
+	// create the XOF output, with hash of collected data as seed
+	h := sha512.New()
```
nikkolasg: Can you create a variable in this package with a clear name related to mixed random stream to indicate the choice of the hash function ?

PizzaWhisperer:
Closing in favour of 2 separate PRs

PR #391, opened 2019-09-20
New randomStream struct to allow use of user-specified entropy source

PizzaWhisperer:
MixedRandStream holds an array of io.Reader that are provided by the user. The associated `XORKeyStream` function works as follow: entropy from each source is mixed via hashing and used as seed to create a blake2bx instance, which perfoms the final XOR.


util/random/rand.go
```diff
@@ -76,3 +81,59 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
+	l := len(dst)
+	if len(src) != l {
+		panic("XORKeyStream: mismatched buffer lengths")
+	}
+
+// try to read 32 bytes from all readers and write them in a buffer
+	var b bytes.Buffer
+	var nerr int
+	for _, reader := range r.Readers {
+		buff := make([]byte, READER_BYTES)
```
nikkolasg: no need to allocate a new buffer each time, you can reuse the same throught the iteration

util/random/rand.go
```diff
@@ -76,3 +81,59 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
 func New() cipher.Stream {
 	return &randstream{}
 }
+
+// READER_BYTES is how many bytes we expect from each source
+const READER_BYTES = 32
+
+// MIXEDRAND_HASH is the hash used in XORKeyStream to mix all entropy sources together.
+// sha512 is chosen as its digest is 64 bytes long, matching the length of blake2xb's seed.
+var MIXEDRAND_HASH = sha512.New
```
nikkolasg: I'm guessing one reason Travis fails is because in Go we don't use UPPER case for constants, we use regular CamelCase.

util/random/rand.go
```diff
@@ -76,3 +81,59 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
+	blake2 := blake2xb.New(seed)
+	blake2.XORKeyStream(dst, src)
+}
+
+// NewMixedStream returns a new cipher.Stream that gets random data from the given
+// readers. If no reader was provided, Go's crypto/rand package is used.
+// In order to use every source, it tries to read 32 bytes from each, and mix it
+// all together into a string of valid length via hashing. This string is then
+// used as a seed for blake2, which is used to perform the final XOR.
+func NewMixedStream(readers ...io.Reader) cipher.Stream {
```
nikkolasg: I'm still more in favor of keeping the same API as before `New()` so we don't introduce a new struct and API for having the same behavior basically. -> what I'm saying is I dont think we need to have another struct for that feature.

But Jeff's opinion should weight more in at this point cc/ @jeffallen

util/random/rand.go
```diff
@@ -76,3 +81,59 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
+
+// MIXEDRAND_HASH is the hash used in XORKeyStream to mix all entropy sources together.
+// sha512 is chosen as its digest is 64 bytes long, matching the length of blake2xb's seed.
+var MIXEDRAND_HASH = sha512.New
+
+type mixedrandstream struct {
+	Readers []io.Reader
+}
+
+func (r *mixedrandstream) XORKeyStream(dst, src []byte) {
```
nikkolasg: Even if it can be time-consuming a few tests would be more than welcome due to the "big responsibility" this feature has :D

util/random/rand.go
PizzaWhisperer: This would be in order to avoid having something like 
`func New(readers ...io.Reader) cipher.Stream {
	if len(readers) == 0 {
		return &randstream{}
	} else {
	        return &mixedrandstream{readers}
        }
}` ?

jeffallen: We normally prefer simpler APIs, even if that requires a more complex internal implementation. So yes, I would like to see `func New(readers ...io.Reader)`, that's a nice backwards compatible, and easy to use API.

PizzaWhisperer: And do you want to keep the two structs separated or should I merge them ?

go.sum
```diff
@@ -16,3 +20,7 @@ golang.org/x/crypto v0.0.0-20190123085648-057139ce5d2b h1:Elez2XeF2p9uyVj0yEUDqQ
 golang.org/x/crypto v0.0.0-20190123085648-057139ce5d2b/go.mod h1:6SG95UA2DQfeDnfUPMdvaQW0Q7yPrPDi9nlGo2tz2b4=
 golang.org/x/sys v0.0.0-20190124100055-b90733256f2e h1:3GIlrlVLfkoipSReOMNAgApI0ajnalyLa/EZHHca/XI=
 golang.org/x/sys v0.0.0-20190124100055-b90733256f2e/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v2 v2.2.2 h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=
+gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
```
jeffallen: There are unexpected new dependencies. Please use `go mod tidy` to clean this up.

util/random/rand.go
```diff
@@ -71,8 +76,60 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
 	}
 }
 
-// New returns a new cipher.Stream that gets random data from Go's crypto/rand package.
+// READER_BYTES is how many bytes we expect from each source
+const READER_BYTES = 32
```
jeffallen: I don't see what this adds to the public interface of this package, so please just move it close to where it is used, and not in package global scope.

util/random/rand.go
```diff
@@ -71,8 +76,60 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
+
+// MixedrandHash is the hash used in XORKeyStream to mix all entropy sources together.
+// sha512 is chosen as its digest is 64 bytes long, matching the length of blake2xb's seed.
+var MixedrandHash = sha512.New
```
jeffallen: It is unclear why this is here; I can see that a caller could choose to modify it, but the docs don't explain that. And don't explain the thread-safety implications of modifying it. I don't see the interest in being able to modify this, let's just hard code it for now.

util/random/rand.go
```diff
@@ -71,8 +76,60 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
+	h := MixedrandHash()
+	h.Write(b.Bytes())
+	seed := h.Sum(nil)
+	blake2 := blake2xb.New(seed)
+	blake2.XORKeyStream(dst, src)
+}
+
+// New returns a new cipher.Stream that gets random data from the given
+// readers. If no reader was provided, Go's crypto/rand package is used.
+// In order to use every source, it tries to read 32 bytes from each, and mix it
```
jeffallen: Change to: Otherwise, for each source, 32 bytes are read. They are concatenated and then hashed, and the resulting hash is used as a seed to a PRNG.

(I don't think the callers need to know about all the details, they can read the source to know.)

util/random/rand.go
nikkolasg: In an ideal world, I'd merge them. Problem is that I think if we merge them, since we change the output of the regular `random.New()` struct, a lot of tests / examples  in kyber might fail now. 
If you can try that, that'd be awesome. If it doesn't fail, then let's merge. If it does, then let's not merge.

util/random/rand.go
nikkolasg: It adds information about how much data will be read from the different readers. The user of this API needs to know how many bytes should its reader have. 
Just to be clear, I explicitely asked @PizzaWhisperer to put it as a variable there so blame it on me ;) 
No worries to change it to a local variable but then we need to explicitely mention that same number in the comment, otherwise the API is un-specified.

util/random/rand.go
nikkolasg: * Good point about the thread safety. 
* About the doc mentionning the change, well, the doc can be changed ;) 
* It's not so uncommon to have variables that can be changed globally in Go code but have a good default purpose. See for example the `crypto/rand.Reader` variable: https://golang.org/pkg/crypto/rand/#pkg-variables . It's even worse in that case because the global variable is mutable. 
* In any case, after some thinking, I would much prefer to stick with `SHA256` (or even better  Blake2s). No need to inccur additional performance hit for no reason (SHA3 is much slower than the two other).

PizzaWhisperer: So you would prefer this line https://github.com/dedis/kyber/blob/93efc4f605eccd9d74cb6de04a9f67b0bb5e0ab3/util/random/rand.go#L116 to simply be `h := sha256.New()` ?

util/random/rand.go
PizzaWhisperer: travis is happy when I merge the two :) so should we keep it that way ?

PR #393, opened 2019-10-01
Fix multiplication for bn256 arm64

AdoAdoAdo:
This pull requests fixes a problem with multiplication on arm64 that is causing a **bn256.G2: malformed point** when running it on a raspberry pi4 with a build for arm64.


ineiti:
+1 from here - so it is the same as the code from cloudflare:

https://github.com/cloudflare/bn256/blob/master/mul_arm64.h

I did a quick check with the other files in the directory, but they seem to be OK.

PR #396, opened 2019-10-14
DKG uses the new random stream structure

PizzaWhisperer:
PR adds two fields to perdersen dkg's configuration, both will be used to the instantiate a new randomStream to pick the secret coefficient:
- `Reader` can hold a user specified entropy source
- `userReaderOnly` boolean forces the use of the user's random data only for the secret coefficient generation (by default the user's data is always combined with randomness from crypto/rand). This is useful for reproducibility and debugging purposes

### DIFF
--- a/share/dkg/pedersen/dkg.go
+++ b/share/dkg/pedersen/dkg.go
@@ -83,7 +83,10 @@ type Config struct {
 	// the number of deals required is less than what it is supposed to be.
 	OldThreshold int
 
-	// Reader holds a user-specified entropy source used to create a random stream
+	// Reader is an optional field that can hold a user-specified entropy source.
+	// If it is set, Reader's data will be combined with random data from crypto/rand
+	// to create a random stream which will pick the dkg's secret coefficient. Otherwise,
+	// the random stream will only use crypto/rand's entropy.
 	Reader io.Reader
 
 	// UserReaderOnly forces the dkg's secretCoeff to be picked with randomness comming
--- a/share/dkg/pedersen/dkg.go
+++ b/share/dkg/pedersen/dkg.go
@@ -89,8 +89,8 @@ type Config struct {
 	// the random stream will only use crypto/rand's entropy.
 	Reader io.Reader
 
-	// UserReaderOnly forces the dkg's secretCoeff to be picked with randomness comming
-	// from the user only, allowing reproducibility
+	// When UserReaderOnly it set to true, it forces the dkg's secretCoeff to be
+	// picked with randomness comming from the user only, allowing reproducibility.
 	UserReaderOnly bool
 }
 
--- a/share/dkg/pedersen/dkg.go
+++ b/share/dkg/pedersen/dkg.go
@@ -10,7 +10,6 @@
 package dkg
 
 import (
-	"crypto/cipher"
 	"crypto/rand"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


share/dkg/pedersen/dkg.go
```diff
@@ -78,6 +82,13 @@ type Config struct {
 	// absent) when doing a resharing to avoid a downgrade attack, where a resharing
 	// the number of deals required is less than what it is supposed to be.
 	OldThreshold int
+
+	// Reader holds a user-specified entropy source used to create a random stream
```
nikkolasg: We should specify what happens when it is set and when it is not set. In particular that when set, dkg still uses the additional crypto reader as well.

share/dkg/pedersen/dkg_test.go
```diff
@@ -1241,3 +1242,53 @@ func TestDKGResharingPartialNewNodes(t *testing.T) {
...
+		NewNodes:       partPubs,
+		Threshold:      defaultT,
+		Reader:         r2,
+		UserReaderOnly: true,
+	}
+	dkg2, err := NewDistKeyHandler(c2)
+	require.Nil(t, err)
+	require.NotNil(t, dkg2.dealer)
+
+	require.True(t, dkg1.dealer.PrivatePoly().Secret().Equal(dkg2.dealer.PrivatePoly().Secret()))
```
nikkolasg: Can we test that if we dont set that flag, the keys are different as well ?

share/dkg/pedersen/dkg.go
```diff
@@ -164,7 +178,18 @@ func NewDistKeyHandler(c *Config) (*DistKeyGenerator, error) {
...
+		if c.Reader == nil {
+			randomStream = random.New()
+		} else { // use the reader they gave, alone or combined
+			if c.UserReaderOnly {
+				randomStream = random.New(c.Reader)
+			} else {
+				randomStream = random.New(c.Reader, rand.Reader)
+			}
+		}
+		secretCoeff := c.Suite.Scalar().Pick(randomStream)
```
Daeinar: Just a minor nitpick regarding style: Can't we initialize `randomStream` by default with `random.New()` and then re-initalize it depending on the options the user provided? Meaning something like that:

```
randomStream := random.New()
if c.Reader != nil && c.UserReaderOnly {
    randomStream = random.New(c.Reader)
}
if c.Reader != nil && !c.UserReaderOnly {
    randomStream = random.New(c.Reader, rand.Reader)
}
```

share/dkg/pedersen/dkg.go
```diff
@@ -164,7 +177,14 @@ func NewDistKeyHandler(c *Config) (*DistKeyGenerator, error) {
 		canIssue = true
 	} else if !isResharing && newPresent {
 		// fresh DKG case
-		secretCoeff := c.Suite.Scalar().Pick(c.Suite.RandomStream())
+		randomStream := random.New()
+		// if the user provided a reader, use it alone or combined with crypto/rand
+		if c.Reader != nil && !c.UserReaderOnly {
```
Gilthoniel: I'm not sure to get why you use the `UserReaderOnly` combined with the actual reader so that it doesn't mix with _crypto/random_. Wouldn't it be better to use _crypto/random_ by default when the reader is nil and use *only* the reader when it is defined ? Is there a security concern to mix it with _crypto/random_ ?

PizzaWhisperer: We are not sure that the entropy given by the user is "good enough". So for security, the reader is mixed with crypto/rand by default to avoid any issue coming from bad randomness.

Gilthoniel: As far as you provide your own source of entropy, it's your job to make sure it's a good one IMHO.

PizzaWhisperer: I agree with you, but we cannot test that the randomness is good, so what we do here if that we make sure to always obtain good randomness no matter what the user supplies

nikkolasg: @Gilthoniel  The goal is to be able to always use a safe source of entropy when doing the DKG. It could have been done in drand, but that would have required (1) a method to mix a reader with crypto/rand - and there's a huge design space to explore so not necessarily trivial and (2) an additional action from the user of DKG before using it. 
I proposed that solution to @PizzaWhisperer because it doesn't involve any additional API to be safe (simply omitting the fields in the config will give you  a safe default) while allowing reproducible tests and external entropy sources.
At this point, we're on the realm of opinions in my opinion, but I'd like to think that this way of doing has more advantages than consequences. Happy to be proven wrong though !

share/dkg/pedersen/dkg.go
```diff
@@ -78,6 +81,16 @@ type Config struct {
...
 	OldThreshold int
+
+	// Reader is an optional field that can hold a user-specified entropy source.
+	// If it is set, Reader's data will be combined with random data from crypto/rand
+	// to create a random stream which will pick the dkg's secret coefficient. Otherwise,
+	// the random stream will only use crypto/rand's entropy.
+	Reader io.Reader
+
+	// When UserReaderOnly it set to true, it forces the dkg's secretCoeff to be
+	// picked with randomness comming from the user only, allowing reproducibility.
```
jeffallen:
```suggestion
	// picked with randomness coming from the user only, allowing reproducibility.
```

PR #405, opened 2019-12-09
bn256: sync stringification with typescript

tharvik:
I propose to add that PR anyway, even if not for the reasons cited in here ;)

TLDR: this makes golang's `bn256.Point.String()` behave the same as javascript's `bn256.point.toString()`.

A follow-up PR should handle #413 

## Previous comment

Points can be one way transformed to String, but the format is not standarized/coherent. As it turn out, Typescript's kyber and Golang's kyber are making subtle differences (adding some separator), which is quite important when generating an UUID out of it, such as in `onet.NewServerIdentity`.

So, currently, **crossing {go,ts}land with identity using bn256 doesn't yield the same ID**, which is breaking quite an important assumption.

As `@dedis/kyber` contains tests related to stringification, I supposed that the issue was to be found in golang kyber. You might want to run `grep -A3 -rE 'func\s+.*\s+String\('` to find *other* issue.

Update: also the width of the string is different, such as `bn256.G1(0,1)` in typescript vs `bn256.G1(0000000000000000000000000000000000000000000000000000000000000000,0000000000000000000000000000000000000000000000000000000000000001)` in golang.


ineiti:
Hmm - usually you should avoid taking hashes or using string-representations when working with crypto. One of the reasons I don't like toml/json-files and then taking the hash of the those files as an identification - toml/json are so brittle and a space can change everything.

jeffallen:
I agree with Linus. The fact that the strings are different is worth cleaning up. But the real bug you are chasing is that NewServerIdentity is using the wrong serialisation to hash technique.

Moving this ti WIP while we discuss the correct fix.

tharvik:
@jeffallen I think we should add a new field to `Point` (or `Marshaling`).

IMO, an ID need to have the next features
* zeroadic, as it is instance based, not parameters based
* vary for every instance, such that changing a bit yield a different one
* easily comparable (better even to have it constant time)
* stable, such that creating an instance with the same parameters yield the same

For now, Point doesn't offer any ID-upgradable field, except for
 * `String()` but that's meant for humans, not really stable (and the why of this PR)
 * `MarshalSize()` and `EmbedLen()` but theses aren't changing much (if at all) per instance
 * `Data()` would be pretty fit, but if I understood it correctly, it works with only some instances (which ID for no-embedding `Point`?)

And now, for some more comments
  * `{Un,}MarshalBinary` are two sides of the coin, we would force these to be constant if we use it directly as an ID
 * `Null()`, `Base()`, `Clone()` all give back a `Point`, not helpful

ineiti:
I think I don't understand what you're trying to achieve. From my point of view, `MarshalBinary` gives back a binary representation of the point that doesn't change in time and that is the same for two equal points.

I tried to understand what your feature-list is, but I'm lost at `zeroadic`, and Google cannot help me, neither. But the other three are fulfilled with `MarshalBinary`. Or, better yet, `MarshalID() + MarshalBinary()`.

tharvik:
> From my point of view, `MarshalBinary` gives back a binary representation of the point that doesn't change in time and that is the same for two equal points.

And that where we disagree: semantically, `MarshalBinary` and `UnmarshalBinary` are not required to be stable through time, the first only need to generate a serialisation that the latter can use. To explicit that it needs to stay the same, we need a field having this semantic specification.

We already have this semantic problem with `String()`, thinking that there is "only one way" to implement it; as it turns out, there was more. Having a dedicated field explicits the stability requirement.

> I tried to understand what your feature-list is, but I'm lost at `zeroadic`, and Google cannot help me, neither.

Hum, that's the first result I: it's a function with zero arguments (or as it is golang, only self).

> But the other three are fulfilled with `MarshalBinary`. Or, better yet, `MarshalID() + MarshalBinary()`.

So theses are zeroadic, yep, but not stable enough IMO.

ineiti:
Last time I looked, `MarshalBinary` followed https://tools.ietf.org/html/rfc8032#section-5.1.2 - and this is supposed _not_ to change. So `MarshalBinary` _is_ stable.

`String()` is quite OK to change. Some time back it didn't follow `fmt.Sprintf("%x", p.MarshalBinary())`, but I think now it does, so you can copy/paste the string from one language in the other. But that's not guaranteed, and I'm OK with that. It's a string, after all. 

See also https://github.com/NathanaelA/nativescript-websockets/issues/75#issuecomment-545973198 for some string-fun ;)

tharvik:
> Last time I looked, `MarshalBinary` followed https://tools.ietf.org/html/rfc8032#section-5.1.2 - and this is supposed _not_ to change. So `MarshalBinary` _is_ stable.

The *implementation of `MarshalBinary` for some `Point`* is stable, but there isn't a *semantic* requirement that it has to.

For example¸ a foreign implementation could implement `{Un,}MarshalBinary` a certain way, then, after some time, uses varint encoding instead of fixedint for a field, thus respecting the requirement of `{Un,}MarshalBinary` but not of time stability.

> See also [NathanaelA/nativescript-websockets#75 (comment)](https://github.com/NathanaelA/nativescript-websockets/issues/75#issuecomment-545973198) for some string-fun

Thanks for the link, that's exactly the issue I'm trying to avoid now :P
By thinking that `String` (read `{Un,}MarshalBinary`) is stable, it introduced this hard-to-find bug.

If you really really want to have only `{Un,}MarshalBinary`, at least, add a big warning that is *has* to be stable through time (but that breaks the golang semantic of `{Un,}MarshalBinary`).

nikkolasg:
I would argue that `time stability` does not exist and any piece of code is evolving (*has* to evolve) through time, and if you want to refer to a certain point in time in the past, that's where versioning comes into place. Even more for a piece of cryptographic code, you should not assume any function to be stable at all points in time, because bugs / vulnerabilities may be found at any point in time. You should assume a given major version to be stable though.

PR #419, opened 2020-04-23
Adds a dkg/pedersen example scenario

nkcr:
Adds an example (test) scenario for dkg/pedersen.

Doing a `ProcessJustification(justification)`  at line 87 yields a panic, is it an intended behaviour ?

### DIFF
--- a/examples/dkg_test.go
+++ b/examples/dkg_test.go
@@ -108,15 +108,16 @@ func Test_Example_DKG(t *testing.T) {
 		shares[i] = distrKey.PriShare()
 		publicKey = distrKey.Public()
 		node.secretShare = distrKey.PriShare()
-		t.Log("distributed public key:", publicKey)
-		t.Log("distributed private share:", shares[i])
+		t.Log("new distributed public key:", publicKey)
 	}
 
-	// 8. Encrypt a secret with the public key and decrypt it with the
-	// reconstructed shared secret key. Reconstructing the shared secret key in
-	// not something we should do as it gives the power to decrpy any further
-	// messages encrypted with the shared public key. Step 9 shows how to
-	// decrypt the message by gathering partial decryptions from the nodes.
+	// 8. Variant A - Encrypt a secret with the public key and decrypt it with
+	// the reconstructed shared secret key. Reconstructing the shared secret key
+	// in not something we should do as it gives the power to decrypt any
+	// further messages encrypted with the shared public key. For this we show
+	// in variant B how to make nodes send back partial decryptions instead of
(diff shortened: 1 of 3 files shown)


nkcr:
In this example I compute the shared secret key once with all the secret shares of the node, which allows me to decrypt the message.
Is it possible to decrypt the message without gathering all the secret shares ? ie. with each node sending back only a partial decryption of the encrypted message, instead of its secret share ?

examples/dkg_test.go
```diff
@@ -106,17 +107,45 @@ func Test_Example_DKG(t *testing.T) {
...
+	// decrypt the message by gathering partial decryptions from the nodes.
 	message := []byte("Hello world")
 	secretKey, err := share.RecoverSecret(suite, shares, n, n)
 	require.NoError(t, err)
 	K, C, remainder := ElGamalEncrypt(suite, publicKey, message)
 	require.Equal(t, 0, len(remainder))
 	decryptedMessage, err := ElGamalDecrypt(suite, secretKey, K, C)
 	require.Equal(t, message, decryptedMessage)
+
+	// 9. Second version, each node provide only a partial decryption
```
ineiti, line 128:
```suggestion
	// 9. Second version, each node provides only a partial decryption
```

nkcr:
I still have this panic at line 87 if I try to call `ProcessJustification`. Is it normal?

examples/dkg_test.go
```diff
@@ -21,8 +23,24 @@ the "happy" path where each node does its job correctly.
...
+	// should illustrate an exponential growth.
+	//
+	// for (( i=1; i<30; i++ )); do
+	//   start=`gdate +%s%N`
+	//   NUM_NODES=$i go test -run Test_Example_DKG >/dev/null
+	//   duration=$(( `gdate +%s%N` - start ))
+	//   echo $duration
+	// done
+	//
+	var nStr = os.Getenv("NUM_NODES")
```
ineiti, line 37: `os.Getenv` in a test? Sounds really strange to me. What about a benchmark? Or just a for-loop in the test?

The only thing I do sometimes is to check for `testing.Short()`. In this case, it could be that the full test has `n := 7`, but using `testing.Short()` it uses only `n := 3`.

nkcr: This provides a quick way to do a benchmark, as described in the comment above. As this scenario is a small demonstration of DKG I find it also interesting to include a word about the complexity. This is an opt-in functionality and the test is stable in its default behaviour. 

For the record:

Time vs number of nodes

PR #422, opened 2020-07-01
Fix a Pedersen DKG check to empty list

Gilthoniel:
This fixes a check that was ignoring empty array.


ineiti:
Nice catch...

TIL:

```
var a []byte
b := make([]byte, 0)
fmt.Println(a == nil, len(a))
fmt.Println(b == nil, len(b))
```

Will print
```
true 0
false 0
```

PR #423, opened 2020-07-03
G2negfix

AkshayaMani:
Sets t properly in the G2 Neg function (in pairing/bn256/twist.go). 
Also added tests for computing pairings with a point output by G2's Neg function.

### DIFF
--- a/pairing/bn256/suite_test.go
+++ b/pairing/bn256/suite_test.go
@@ -365,3 +365,65 @@ func Test_g2_2add_oncurve_issue400(t *testing.T) {
 	err = p.UnmarshalBinary(ma)
 	require.NoError(t, err)
 }
+
+// Test doing pairings with a point produced by G2's Neg function
+// This fails with useneg == true in kyber 3.0.11.
+func testpairG2negUseneg(t *testing.T, useneg bool) {
+	suite := NewSuite()
+	g1 := suite.G1().Point().Base()
+	g2 := suite.G2().Point().Base()
+	g2neg := suite.G2().Point()
+	if useneg {
+		// Use Neg to get the negation
+		g2neg = suite.G2().Point().Neg(g2)
+	} else {
+		// Multiply by -1 to get the negation
+		minus1 := suite.G1().Scalar().SetInt64(-1)
+		g2neg = suite.G2().Point().Mul(minus1, g2)
+	}
+	b := suite.G1().Scalar().SetInt64(2)
+	g1b := suite.G1().Point().Mul(b, g1)
+	g2negb := suite.G2().Point().Mul(b, g2neg)
+	res1 := suite.Pair(g1b, g2neg)
+	res2 := suite.Pair(g1, g2negb)
+	require.Equal(t, res1, res2)
+}
+
+// The same for G1, just for completeness (but this one does not fail on
+// 3.0.11)
+func testpairG1negUseneg(t *testing.T, useneg bool) {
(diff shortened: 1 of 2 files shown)


ineiti:
Thanks for the PR. Here some comments:
- the tests fail because the code is not properly formatted. Please use `go fmt`
- I think you mixed up `inverse` and `negate` in the comments to the tests:
  - `inverse` is with regard to the multiplication operation, and only available in `Scalar`
  - `negate` is with regard to the addition operation, and available both in `Scalar` and `Point`
- I didn't look at the other tests, but probably `G1`, `G2`, and `GT` are not tested using https://github.com/dedis/kyber/blob/master/util/test/test.go#L368 - I wonder if `test.GroupTest` would uncover other errors

So proposed next steps:
1. try if `test.GroupTest` would catch the `negate` error, if yes, use that one for `G1`, `G2`, and `GT`
2. if `test.GroupTest` doesn't catch this error, the `Neg`-test in `test.testGroup` should be modified to catch it

Even though I think `test.GroupTest` will not pass on the `G*` groups, as they are missing some implementations...

AkshayaMani:
As @ineiti suspected, test.GroupTest will not pass on the G* groups, as they are still missing some implementations. All other issues have been addressed.

ineiti:
Thanks for this new commit. How bad was the failure of the `test.GroupTest`? Was it just 1-2 failing tests, or more?

For the `testpairG2negUseneg`, please add a third argument, `kyber.Group`, so you don't need a `G1` and a `G2` version of it, but can call it with the appropriate group.

pairing/bn256/suite_test.go
```diff
@@ -365,3 +365,65 @@ func Test_g2_2add_oncurve_issue400(t *testing.T) {
 	err = p.UnmarshalBinary(ma)
 	require.NoError(t, err)
 }
+
+// Test doing pairings with a point produced by G2's Neg function
+// This fails with useneg == true in kyber 3.0.11.
+func testpairG2negUseneg(t *testing.T, useneg bool) {
```
ineiti, line 371: please add a third argument, kyber.Group, so you don't need a G1 and a G2 version of it, but can call it with the appropriate group.

AkshayaMani:
> Thanks for this new commit. How bad was the failure of the `test.GroupTest`? Was it just 1-2 failing tests, or more?

All three groups (G1, G2, GT) panic when testGroup uses testEmbed to call Embed at [https://github.com/dedis/kyber/blob/master/util/test/test.go#L42](https://github.com/dedis/kyber/blob/master/util/test/test.go#L42) .

> For the `testpairG2negUseneg`, please add a third argument, `kyber.Group`, so you don't need a `G1` and a `G2` version of it, but can call it with the appropriate group.

The suite.Pair() function requires its first argument to be in G1, and its second to be in G2; the testpairG1negUseneg test negates the first argument and the testpairG2negUseneg test negates the second. So passing in an argument of the group (or even of which group to use) would just result in a lot of code wrapped in "if"s to negate, and then pair, the correct variable.

ineiti:
Sorry for insisting, but the tests are often used as a way to know how to use the library, so we like to keep them nice and short. Please have a look at #424 and tell me if that looks OK and cleaner to you, too.

Best,

Linus

AkshayaMani:
Thank you. Looks good!

ineiti:
Closing in favour of #424

PR #424, opened 2020-07-17
G2negfix

ineiti:
Simpler test for g2negfix, else it's the same as dedis/kyber#423


jeffallen:
Ian confirmed in e-mail that he is the author of these commits, and they they arrived via his student's Github account with his blessing. He also has signed a CLAI.

PR #425, opened 2020-07-20
Schnorr signature verification with a null public key is an error

jeffallen:
Schnorr signatures are only defined for non-zero private keys because the
corresponding public key to a zero private key is the mulplicative identity,
causing the message hash to be excluded from the signature verification.

Do not allow null public keys so that if one is passed in by accident, Verify
returns an error instead of saying that the signature is good, even if the
message was modified.

It is still the caller's responsibility to use Kyber to make
valid secret keys.

### DIFF
--- a/sign/eddsa/eddsa.go
+++ b/sign/eddsa/eddsa.go
@@ -122,6 +122,10 @@ func (e *EdDSA) Sign(msg []byte) ([]byte, error) {
 // Verify uses a public key, a message and a signature. It will return nil if
 // sig is a valid signature for msg created by key public, or an error otherwise.
 func Verify(public kyber.Point, msg, sig []byte) error {
+	if public.Equal(group.Point().Null()) {
+		return errors.New("invalid public key")
+	}
+
 	if len(sig) != 64 {
 		return fmt.Errorf("signature length invalid, expect 64 but got %v", len(sig))
 	}
--- a/sign/schnorr/schnorr.go
+++ b/sign/schnorr/schnorr.go
@@ -31,6 +31,7 @@ type Suite interface {
 // signature when using the edwards25519 Group.
 func Sign(s Suite, private kyber.Scalar, msg []byte) ([]byte, error) {
 	var g kyber.Group = s
+
 	// create random secret k and public point commitment R
 	k := g.Scalar().Pick(s.RandomStream())
 	R := g.Point().Mul(k, nil)
@@ -60,6 +61,9 @@ func Sign(s Suite, private kyber.Scalar, msg []byte) ([]byte, error) {
 // Verify verifies a given Schnorr signature. It returns nil iff the
 // given signature is valid.
 func Verify(g kyber.Group, public kyber.Point, msg, sig []byte) error {
+	if public.Equal(g.Point().Null()) {
+		return errors.New("invalid public key")
+	}
 	R := g.Point()
 	s := g.Scalar()
 	pointSize := R.MarshalSize()
--- a/sign/eddsa/eddsa_test.go
+++ b/sign/eddsa/eddsa_test.go
@@ -120,6 +120,28 @@ func TestEdDSASigningRandom(t *testing.T) {
 	}
 }
 
+func TestInvalid(t *testing.T) {
+	suite := edwards25519.NewBlakeSHA256Ed25519()
+	ed := NewEdDSA(suite.RandomStream())
+	// Simulate the situation that somehow, due to wrong initialisation
+	// or whatever, the caller ends up signing with a zero secret key.
+	ed.Secret = suite.Scalar().Zero()
+	ed.Public = suite.Point().Null()
+
+	msg := []byte{1, 2, 3}
+	sig, err := ed.Sign(msg)
+	assert.NoError(t, err)
+
+	// Corrupt the message, will the signature still be valid?
+	msg[0]++
+
+	// Without the "is point not Nul" check in Verify, this
+	// call to Verify returns no error, even though the message
+	// has been corrupted!
+	err = Verify(ed.Public.Null(), msg, sig)
+	assert.Error(t, err)
+}
+
 type constantStream struct {
 	seed []byte
 }
--- a/sign/schnorr/schnorr_test.go
+++ b/sign/schnorr/schnorr_test.go
@@ -96,3 +96,32 @@ func TestQuickSchnorrSignature(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+// Schnorr is only defined for scalars which are not the additive identity
+// because if it was allowed, then the math works out such that the message
+// hash is ignored, and the same signature would verify for any message
+// for the public key which is the neutral identity point. The following
+// two tests check the errors we added to protect callers from this mistake.
+
+func TestBadPubKey(t *testing.T) {
+	suite := edwards25519.NewBlakeSHA256Ed25519()
+
+	// This is a good signature made on msg with private
+	// key Scalar().Zero() before that input was made into
+	// an error in Sign.
+	msg := []byte("Hello Schnorr")
+	sig := []byte{0xf4, 0x8a, 0x95, 0x90, 0xf7, 0xcc, 0x4d, 0x7b, 0xc3, 0xd3, 0x88, 0x64, 0x0, 0xd5, 0xf2, 0x9a, 0x3c, 0xde, 0x60, 0xd9, 0xe5, 0xb2, 0x4d, 0x68, 0x4c, 0x23, 0x7d, 0x6, 0x7c, 0x3, 0xcf, 0x0, 0x2c, 0xe, 0x17, 0xbf, 0xb, 0x9b, 0xa1, 0x2b, 0xa2, 0x10, 0xae, 0x59, 0x3d, 0xd, 0x34, 0xa9, 0x10, 0x31, 0x58, 0x9, 0x92, 0x40, 0x50, 0x68, 0x5a, 0x7c, 0xe7, 0x62, 0x32, 0xc7, 0xa5, 0x8}
+
+	// Modifying one byte of the message would still allow
+	// the signature to verify because the public key is Null.
+	msg[0]++
+
+	pn := suite.Point().Null()
+	err := Verify(suite, pn, msg, sig)
+
+	// Before the check, err was nil here, a dangerous situation where a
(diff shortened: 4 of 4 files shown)


sign/eddsa/eddsa.go
```diff
@@ -69,6 +73,9 @@ func (e *EdDSA) UnmarshalBinary(buff []byte) error {
 	e.seed = buff[:32]
 	e.prefix = prefix
 	e.Secret = secret
+	if e.Secret.Equal(group.Scalar().Zero()) {
+		return errors.New("invalid secret key")
```
ineiti, line 77: Again, I'm not sure if the test should be here... Do you really want to make sure that you cannot unmarshal a zero? Perhaps as it's only for EdDSA it might be OK.

sign/schnorr/schnorr.go
```diff
@@ -60,6 +65,9 @@ func Sign(s Suite, private kyber.Scalar, msg []byte) ([]byte, error) {
 // Verify verifies a given Schnorr signature. It returns nil iff the
 // given signature is valid.
 func Verify(g kyber.Group, public kyber.Point, msg, sig []byte) error {
+	if public.Equal(g.Point().Null()) {
+		return errors.New("invalid public key")
```
ineiti, line 69: Giving an error if you want to sign with a zero key makes sense. Just asking - does it make sense to forbid checking the signature with a Null point? It's always true...

nikkolasg:
Just chiming in here to say that these checks are not exhaustive and shouldn't be at this level I think. What a good API should do is to enable to generate good keys, and sign/verify with those keys. If a user wants to use something else, and generate a scalar which is 0, or 1, or 2 ... (non exhaustive list!) then it's his problem and its signature will be invalid. 
What the API needs to make sure is that honest users can't generate insecure keys, and attacker can't derive signatures with their keys.

ineiti:
OK, votes are in - add/remove as wished ;)

- `schnorr.Sign`: jeffallen, ineiti
- `schnorr.Verify`: jeffallen
- `Group.NewKeyAndSeed`: ineiti, nikkolasg
- `Group.NewKey`: ineiti, nikkolasg
- `EdDSA.Unmarshal`: jeffallen

sign/eddsa/eddsa.go
```diff
@@ -34,6 +34,10 @@ func NewEdDSA(stream cipher.Stream) *EdDSA {
 	}
 
 	secret, buffer, prefix := group.NewKeyAndSeed(stream)
+	if secret.Equal(group.Scalar().Zero()) {
+		panic("random stream caused private key to be zero")
```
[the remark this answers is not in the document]
jeffallen: If you retry, how many times? In line 33, there's already a panic for a holly unsuitable random stream. One that returns a zero private key is also unsuitable.

jeffallen:
schnorr.Verify absolutely has to have the protection, because without it, if the public key accidentally (or via incorrect input from an attacker) ends up the Null point, then schnorr.Verify ends up a no-op. Like @nikkolasg , I don't really care how/where problematic scalars come from, as long as Verify never lies and says a message which was not signed was signed.

Putting that argument into words helps me see I got totally off track in eddsa.go: there I need one single check in Verify and not all the other stuff.

ineiti:
I think it might be best to have some real-time chat about this, else this discussion will get very long ;)

Funny also that we both think @nikkolasg supports us ;) From what I understand he writes, we should only verify for `0` or low values of the scalar in the `NewKey*` methods.

I would also be OK to verify against `0` in the `schnorr.Sign` method, but that's less important to me.

I'm trying to come up with a good example of your attack using the Null public key that would be caught by `schnorr.Verify`. If a message has been signed by a non-zero scalar, then it will not pass verification with a `Null` point. This is why I think that the `schnorr.Sign` should check for `0` scalar, or, even more importantly, the `NewKey*` methods.

Then the `schnorr.Sign` and `schnorr.Verify` would accept just whatever you throw at them, which makes sense for a basic crypto library.
Only the `NewKey*` methods would return a somewhat 'secure' private key.

This also reminds me of one of the code-reviewers comments on our use of `Scalar` and `Point` for `Private` and `Public` keys... Mathematically it's the same, but conceptually not, as seen here: a `Scalar` can be `0`, while a `Private` should never be `0`.

nikkolasg:
> schnorr.Verify absolutely has to have the protection, because without it, if the public key accidentally (or via incorrect input from an attacker) ends up the Null point, then schnorr.Verify ends up a no-op

What ? No schnorr verification should fail if the public key does not correspond to the private key used to making the signature, that's it. There should NOT be any need to do any "manual check" in schnorr.Verify/Sign that's what I meant.

>From what I understand he writes, we should only verify for 0 or low values of the scalar in the NewKey* methods.

We could but I would simply make sure we take from a correct random stream - that's it. I never saw any keygen API that checked "manually" the secret / public key, but they all do make sure that they fetched enough random bytes from a secure source. 
(hence my suggestion is to do nothing :D unless the keygen can be improved for that ^ )

jeffallen:
Before having a very long conversation in any medium, please take a fresh look at this new commit.

Please think from the perspective of a caller to Verify who expects it to NEVER EVER lie to them, but the caller has a mistake in their program and though they think they are using a good private key, they actually have a zero private key. (This is exactly how I found this situation.)

Which is better: that they go on not realising that they are calling Sign and Verify with less that result in signatures never failing? Or that Verify fails?

Now think abut a correct-enough implementation where somehow a bad guy still manages to get the secret set to zero... now he's able to pass any message he wants as signed. With this change, his "always true" signature is useless to him, because it is now an "always invalid public key" signature.

nikkolasg:
> Please think from the perspective of a caller to Verify who expects it to NEVER EVER lie to them, but the caller has a mistake in their program and though they think they are using a good private key, they actually have a zero private key. (This is exactly how I found this situation.)

What's the reason you ended up with a zero key ? Because you used a custom `cipher.Stream` ? 
BTW For me this thing is non-idiomatic and should be replaced via a simple `io.Reader` as what `crypto/rand` provides. 

> Which is better: that they go on not realising that they are calling Sign and Verify with inputs that result in signatures never failing no matter how the message is corrupt? Or that Verify fails?

That the key generation never outputs a zero key, which is normally the case if you use a secure source of randomness: i.e. the API should only expose creating a key from a secure source of randomness.

> 

> Now think abut a correct-enough implementation where somehow a bad guy still manages to get the secret set to zero... now he's able to pass any message he wants as signed. With this change, his "always true" signature is useless to him, because it is now an "always invalid public key" signature.

If a an attacker is able to change the private key of an honest user, he doesn't need to set it to zero in the first place, he can do whatever he wants with it. I never heard of any attacks where corrupting the key to zero is helping anything. 
Moreover, that are **consideration of the higher level protocols** : if you give a secret = 0 to schnorr.Sign and then verify with G, then the verification should pass. If a user got his key changed by an attacker, then the protocol above need to detect it if we want to prevent that (like you always register your latest key on chain no? if you change your signing key, pgp will detect this at the higher level than the crypto operation, at the filesystem / store level, when it sees it is not the hash that it has saved, etc)

Of course, you can always implement manual check but i find them useless because secret key = 1, 2 3 4 etc is as dangerous as 0, and more prone to errors / complication down the line.

ineiti:
I see where you're coming from and that you would like to have the library tell you what you did is wrong. So you protect against a wrong verification, which protects your use case. But currently you still allow the creation of useless signatures and for dangerous keys. 

>  I never saw any keygen API that checked "manually" the secret / public key

At least Wikipedia has an algorithm that uses [1..n-1] for the private key: https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm#Signature_generation_algorithm
But RFC8032 only mentions the hash and the bit-twiddling, which will also make sure that the private key is != 0...

So I still keep thinking that the order of importance of checking the `0` private key is:
- `NewKey*` - this is the closest in the library to something like a `Private` key, as opposed to a `Scalar`, which can really be anything
- `schnorr.Sign` - if we decide that we never want to sign using a `0` scalar - but it's not taking a `Private` key, so a `Scalar` would be OK to me
- `schnorr.Verify` - that's the least of my worries, as it's at the very end of the chain. If you came this far with a `0`/`Null` key, it's probably intended

Perhaps the `It will need an independent security review before it should be considered ready for use in security-critical applications. ` should be made more prominent...

ineiti:
> BTW For me this thing is non-idiomatic and should be replaced via a simple `io.Reader` as what `crypto/rand` provides.

I think that is mostly historic, as the first random sources were block ciphers - also now we have XOF ciphers as random sources. And then you have this XOR interface...

jeffallen:
Ok, let me ask a new question. If you put personal preferences for where checks happen in the code aside (or if they even should happen at all), is this change correct? Does it correctly save a caller who accidentally ended up trying to verify a message signed with an invalid key?

Because that's the goal of this PR: to add a lightweight guard against a real failure case that we really saw when a lazy user was trying to use Kyber by writing his unit test wrong ("guilty as charged, your honour").

Kyber is used in teaching, experimental, and research environments, so we need to observe how people fail at using Kyber and fix that.

jeffallen:
Ok, you've made your case, expect a new version of this patch tomorrow.

nikkolasg:
>If you put personal preferences for where checks happen in the code aside (or if they even should happen at all), is this change correct?

It's not personal preference, it's "best practices" :) 
Yes the changes are correct functionally.

ineiti:
Also, regarding the PR #427, I wonder if we should not simply define `schnorr.*` as `experimental - don't use in production`, and concentrate on the `eddsa.*` implementations.

Which means that all verifications in `byzcoin` are in fact moot and should be replaced asap :( Seems we got caught in our own experiments and forgot to correctly separate `Scalar` and `Private Key`...

jeffallen:
So on further investigation, protecting against a zero private key is pointless, the seed is already hashed. I have decided just not to do anything here; the problem I stumbled over is real, but the way I fell into it was by not using the API correctly.

ineiti:
> not using the API correctly.

this sucks - and definitely reminds me of the code-review where they complained about our usage of `Scalar` instead of `Private`...

So, yes, this _is_ an experimental library that allows you to shoot yourself in the foot :(

Btw, I updated my wish-list for darcs: https://github.com/dedis/cothority/issues/2186 - in case they are compatible between cothority and dela, it would be nice to extract them and use them for both projects...

Issue #426, opened 2020-07-20
Insecure signature verification algorithm for ed25519


calctopian:
As currently implemented, the verification algorithm is not strongly unforgeable under chosen message attacks (SUF-CMA), that is, an adversary can construct an alternative signature for a given signed message; moreover, it doesn't provide guarantees of malicious strong universal exclusive ownership (M-S-UEO) to prevent key substition attacks. Furthermore, it doesn't guarantee Message-Bound Signatures (MBS), i.e., there exist no two distinct messages for which the same signature would verify with respect to a given (potentially maliciously generated) public key.

Warning: failing to patch this vulnerability could lead to double-spending attacks on cothority.

Fixed by pull request #427
Supersede issue #311

calctopian:
Fixed by #427

ineiti:
Thanks for the warning for double-spending attacks ;) I'd have to look into our verifications to make sure this is not possible.

calctopian:
About the exploitability of this issue:

If a malicious attacker adds one of points of the subgroups of small order (i.e., non-identity torsion points) to a key image, up to 7 double spends could be performed for each authentic spend on the cothority: this is now prevented by the Ge25519HasSmallOrder function.

This happens because multiplying the sum of a torsion point by a multiple of 8 and a "normal" point will return the same point, because multiplying small subgroup points by 8 returns the identity element. Without the function Ge25519HasSmallOrder, signatures verify on malicious key images derived from the authentic ones.

ineiti:
Just for naming: `cothority` is the big project that includes a lot of smaller projects. The biggest of the smaller projects is `byzcoin`, which implements a blockchain using a BFT algorithm. So the attack would be against `byzcoin`.

I wonder how feasible such an attack would be, given that each transaction that is signed also has a `nonce` in it. So it is not possible to 'double spend' or 'replay' a transaction - you would need to be able to increase the `nonce` and still have a valid signature, which, if I understand the attack correctly, is not possible.

Also, can you please elaborate what you call a `key image`?

calctopian:
Byzcoin using nonces is great because nonces provide extra protection: additionally, Byzcoin is using blscosi and not EdDSA as was previously used with cosi: thus transactions seem to remain safe from this attack.

[...]

Issue #311, opened 2018-05-17
[ed25519/eddsa] Malleable signature


nikkolasg:
We should enforce that the ed25519 and eddsa packages enforce non-malleability of the signatures.
See agl's [commit]( https://github.com/golang/crypto/commit/c4a91bd4f524f10d064139674cf55852e055ad01) for example. 
Unfortunately, it is not straightforward to do that at the current state of the code, because we currently don't have a way to get the field order (to check if the scalar is in the range `[0,order[`).
See https://github.com/dedis/kyber/issues/259 for a related issue.

jeffallen:
@Daeinar could you please comment on this? To me it seems like a nice-to-have, but not critical before v3 (because it is not a backwards incompatible change)

calctopian:
Superseded by issue #426 and solved by PR #427.

Issue #402, opened 2019-11-24
Is there a plan to add support for BLS12-381?


wangyifan:
It seems many projects(e.g. zcash and ethereum 2.0) now use BLS 12-381 for BLS signature. BN256 is documented as weak [in this issue](https://github.com/golang/go/issues/19479). Is there a plan to add support for BLS12-381?

jeffallen:
No, there is no plan for that, but we do accept PRs.

wangyifan:
@nikkolasg thanks for the heads up. I will take a look.

pierluca:
This has been added in #505

PR #427, opened 2020-07-20
Fix ed25519 verification algorithm

calctopian:
The current implementation is insecure and non-compliant to RFC8032: for example, [RFC8031 Section 5.1.7](https://tools.ietf.org/html/rfc8032#section-5.1.7) requires checking sig to prevent signature malleability.

A recent paper ("[The Provable Security of Ed25519: Theory and Practice](https://eprint.iacr.org/2020/823.pdf)") teaches potential shortcomings that occur when implementing Ed25519: in short, a correct implementation must provide strong unforgeability (SUF-CMA), resilience against key substitution attacks (S-UEO and M-S-UEO) and ensuring that a signature verify a unique message (MBS), even for malicious keys.

This Pull Request fixes kyber according to the RFC8032 and the state of the art in Ed25519 implementations (all the previous properties, SUF-CMA, S-UEO, M-S-UEO and MBS).

Fixes issues #426 and #311.


ineiti:
Thank you very much for your PR. Can you please:

- add the references to RFC8032 and the paper as a comment in the code
- add tests that show how this works/fails

sign/eddsa/eddsa_test.go
```diff
@@ -101,6 +101,35 @@ func TestEdDSASigning(t *testing.T) {
 	}
 }
 
+// Test signature malleability
+func TestEdDSASigningMalleability(t *testing.T) {
+	/* 2^252+27742317777372353535851937790883648493 */
```
ineiti, line 106: is this the prime order of the base point? That would be more interesting to know here - I have to admit I only recognize the 2**252, which I know is part of the prime order. But I don't know if it's +1, -1, +1234567...

sign/eddsa/eddsa_test.go
```diff
@@ -101,6 +101,35 @@ func TestEdDSASigning(t *testing.T) {
...
+	var L []uint8 = []uint8{0xed, 0xd3, 0xf5, 0x5c, 0x1a, 0x63, 0x12, 0x58, 0xd6, 0x9c, 0xf7,
```
ineiti, line 107: as you're casting this to `uint16` anyway, might as well start with `uint16`

ineiti:
Nice first test - looking at the PR you found an error ;) I would like to have also tests for the following:
- `Ge25519HasSmallOrder`
- `Ge25519IsCanonical`
- `Sc25519IsCanonical`

You should do the tests with a loop that tests different values for each of these methods.

group/edwards25519/ge.go
```diff
@@ -547,16 +547,16 @@ func Ge25519HasSmallOrder(s []byte) int32 {
 //
 // Checks whether group element s is less than p
 func Ge25519IsCanonical(s []byte) int32 {
-	var c uint8
-	var d uint8
+	var c = 0
```
ineiti, line 550:
```suggestion
	c := 0
```

same for `var d = 0`

sign/eddsa/eddsa_test.go
```diff
@@ -130,6 +130,28 @@ func TestEdDSASigningMalleability(t *testing.T) {
 	assert.Error(t, Verify(ed.Public, msg, sig))
 }
 
+// Test non-canonical keys
+func TestEdDSASigningNonCanonical(t *testing.T) {
+	var nonCanonicalPk []byte = []byte{0xf6, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
```
ineiti, line 135: What is that number? Why is it noncanonical? Can you write a loop that creates canonical and non-canonical public keys for tests?

Also, if I understand correctly, this test will fail even with a canonical public key, no?

What I proposed was to write a test that directly tests the methods I listed, without using the `Sign` method.

sign/eddsa/eddsa_test.go
```diff
@@ -130,6 +130,28 @@ func TestEdDSASigningMalleability(t *testing.T) {
...
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f}
+	suite := edwards25519.NewBlakeSHA256Ed25519()
+	ed := NewEdDSA(suite.RandomStream())
```
ineiti, line 139:
```suggestion
	ed := NewEdDSA(random.New())
```

Then you don't need the `suite`...

sign/eddsa/eddsa_test.go
calctopian: Yes, that is l, the prime order of the base point: I'll document it on the source code.

sign/eddsa/eddsa_test.go
calctopian: OK, I'll change that.

group/edwards25519/ge.go
calctopian: OK, I'll change these.

sign/eddsa/eddsa_test.go
calctopian: I wrote that because that what is used on TestEdDSASigningRandom, thus everything remains uniform.

sign/eddsa/eddsa_test.go
calctopian: That number is a non-canonical point to be used as a public key in that function: the Verify function must fail with non-canonical public keys.

These tests are verifying the integrity of the edwards25519.Verify function, now reinforced with new functions (Ge25519IsCanonical, Sc25519IsCanonical...)

calctopian:
For the moment, TestEdDSASigningMalleability is testing Sc25519IsCanonical and TestEdDSASigningNonCanonical is testing Ge25519IsCanonical.
These tests are checking that the edwards25519.Verify function works correctly by detecting non-canonical signatures and public keys: please note that what really matters is the full integrity of the whole Verify function, not individual functions. If you write tests for the individual functions (Sc25519IsCanonical, Ge25519IsCanonical, ...) but don't test their integration on the edwards25519.Verify function, you are setting yourself up to failure because what matters is their combination.

sign/eddsa/eddsa_test.go
ineiti: OK - then I think we need to change the `TestEdDSASigningRandom` ;)

sign/eddsa/eddsa_test.go
ineiti: How did you chose that number?

How can you be sure the verification fails because it's non-canonical and not just because it is a wrong public key and thus the verification fails anyway?

ineiti:
About testing `Sc25519IsCanonical`, `TestEdDSASigningNonCanonical`, `Ge25519IsCanonical`: these methods are very difficult to read for your average programmer (me). If it's just some bit-twiddling or so I can convince myself that it's OK. But these are some cryptographic methods that should really hold up to tests. If I were to do a code review, and these methods don't have a unit-test, I would start by making some to be sure they hold up.

Probably for you it's clear why and how they work. Please write down your knowledge by creating a unit-test for each of the three methods. Each unit test should:
- show the method passing and failing
- use more than one input for passing and failing
- explain why they fail

It is also important to explain why you chose your numbers. If it's `random numbers`, please use actual random numbers, if possible even in a loop to test more than one. Personally I'm OK with different random runs for every test, while others prefer pre-initialized random number generators.

If it's off-by-one numbers, explain what they are off-by-one or off-by-two. If possible, use the constants from the group-definitions and add/subtract from those.

If it's magical numbers, explain why they are magical, and how we can be sure that your magic is reproducible (at which point the magic collapses to science).

sign/eddsa/eddsa_test.go
calctopian: It's a well-known non-canonical point, also used in other libraries: standards are full of constants and test vectors.

calctopian:
Yes, in an ideal world there would be unit tests for everything: for example, where are the unit tests for geScalarMultBase, geScalarMult, feMul, feSquare, ...? Why are there dozens of functions without unit tests on kyber/group/edwards25519/, directly translated from the "ref10" implementation of ed25519 (SUPERCOP)? 

I was following the same rules that have been applied to the current codebase. According to you: should I follow a different "double standard"?

ineiti:
About the 'double standard': code evolves, so does the knowledge of what is 'right' in code. Unfortunately we don't have the time to update all our codebases to the best coding practices we develop over time.

Also, if it's an internal committer, we know that she'll be around to fix eventual bugs. For an external committer, it's more important to be sure that her code will be correct, e.g., https://github.com/dedis/kyber/pull/427/commits/cf5b86d0c73eb9ec9c976adf0aa08a2ddd487169#diff-bd9209ce75feb92351db6c68bd8010d7R128

We appreciate your participation in the code, and would really like this commit to finish in the master branch. But to avoid errors, I need to insist on more tests. For other PRs I wrote the tests myself (#424), because it was easy to understand what's going on. For this PR I'm not sure what you're doing, my gut tells me you're right, so I follow through with the commits. But I don't have the time to read the papers, look if you implemented it correctly or not. This is what tests are for. And they're currently missing for the three main methods you want to add.

And, yes, there are a lot of tests missing in other places, too. Which bit us already more than once: #401, #411 - even code from Cloudflare was not perfect. So we'd like to avoid that in the future...

sign/eddsa/eddsa_test.go
calctopian: Don't worry, on this PR I have change them all.

calctopian:
I can't believe that you keep defending that there should double-standards, one for externals and one internals. What is written on **The Book** about this? ;)

	- Mark it down: **G-d** hates it when you demonstrate a double standard — one for “them” and one for “you.” [Proverbs 20:10](https://www.biblegateway.com/passage/?search=Proverbs+20%3A10&version=TPT)
	- **The L-rd** hates double standards — that’s hypocrisy at its worst! (Proverbs 20:23](https://www.biblegateway.com/passage/?search=Proverbs+20%3A23&version=TPT)

sign/eddsa/eddsa_test.go
calctopian: Now TestGe25519IsCanonicalChecksNonCanonicals generates that non-canonical point and all the others.

ineiti:
Let's keep this on-topic - I'm very happy to discuss the Lord's ways if you wish, but I don't think this is the right place. You can send me an email and I'd be very happy to discuss it there.

Here I am concerned about technical correctness. Personally I do not have the time to dive into this matter, and currently nobody else on the team has this time to look into this. So I see two interesting ways forward:
- you wait for somebody to look into this and write the tests in your place
- write the tests as I asked you to do and answer my questions

I would be very sad to have this PR be closed without a merge into master, but it is an option...

calctopian:
What of your questions on this PR remains unanswered? None.
What of your suggestions did I object to? None.
As you can check, I have already written two of the tests that you asked for: TestSc25519IsCanonical and TestGe25519ChecksNonCanonical for Ge25519IsCanonical.
I'm very compliant with everything that you have asked me to do.

This PR also fixes issue #311 that has been opened for more than 2 years: in fact, you can't put anything into production without fixing it. Are you really saying that nobody else on your team has had the time to look into it, _in 2 years_?

I'm very concerned about technical correctness, so concerned about it that I'm fixing your own bugs: what you have to understand is that technical correctness is a function of how a software project is managed.

ineiti:
Thanks - I did only see your comment, not the new commit. The tests look much better now, and I can better follow your reasoning and convince myself that the code is correct.

I still do have some questions about the tests, but I will discuss it with a a member of our team who has time to look into the papers. Once we made sure that everything is correct and we're happy with the code, the PR will be merged.

calctopian:
You [latest commit](https://github.com/drand/kyber/commit/e24f401bec68bf3dd0b13c748d921553b42d5c13) to drand/kyber  introduced the interface SubGroupElement with the function IsIncorrectGroup(), but _they are not implemented anywhere_.
Therefore, that **NOT** "allows to check for non malleability during the verification".

nikkolasg:
It's implemented in another repo that's why: https://github.com/drand/kyber-bls12381/blob/master/kyber_g2.go#L149

General tip: please try to ask if you don't know before judging too fast :)

ineiti:
Hmm - took me quite some time to find it ;) 

https://github.com/kilic/bls12-381/blob/59713d4652a1851eb2ad0fb980533a5472294fe0/g1.go#L280

But I like that it's defined on the point instead of being a method aside that takes a slice of bytes.

calctopian:
General tip: when somebody asks you to review a PR, you have to review that PR, only talking about what happens in other repositories is not reviewing this PR. Linus has reviewed this PR, you haven't.

Here, we're not interested about your fix for only a specific curve (i.e., only implemented for bls12-381) because dedis/kyber [doesn't support bls12-381](https://github.com/dedis/kyber/issues/402): we appreciate your PRs in this repository, please do not create unnecessary forks.

Your "look what I'm doing in other forked repositories for other curves, ain't it cool?" is completely irrelevant here (i.e., impertinent comments).

gnarula:
Closed because of #430

PR #430, opened 2020-07-30
EdDSA Verification Checks

gnarula:
This PR builds on #427 with the following modifications:

* The `IsCanonical` and `HasSmallOrder` checks are moved to `Point` and `Scalar` implementations of `edwards25519`
* Tests use the constants defined in `edwards25519`
* Ensured bitwise operations are on unsigned types / bytes
* Added `eddsa.VerifyWithChecks(pub, msg, sig []byte)` which checks constraints as imposed by RFC8032 and group order checks as suggested in https://eprint.iacr.org/2020/823.pdf and implemented in [libsodium](https://github.com/jedisct1/libsodium/blob/4744636721d2e420f8bbe2d563f31b1f5e682229/src/libsodium/crypto_sign/ed25519/ref10/open.c#L31)


sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+	assert.Nil(t, err)
+
+	err = Verify(ed.Public, msg, sig)
+	assert.Equal(t, fmt.Errorf("public key is not canonical or has small order"), err)
+}
+
 // Test the property of a EdDSA signature
 func TestEdDSASigningRandom(t *testing.T) {
-	suite := edwards25519.NewBlakeSHA256Ed25519()
+	//suite := edwards25519.NewBlakeSHA256Ed25519()
```
ineiti, line 205:
```suggestion
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+func TestEdDSAVerifyNonCanonical(t *testing.T) {
+	var nonCanonicalPk []byte = []byte{0xec, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}
+	ed := NewEdDSA(random.New())
+
+	msg := make([]byte, 32)
+	_, err := rand.Read(msg)
+	assert.NoError(t, err)
+
```
ineiti, line 168:
```suggestion
      msg := random.Bits(256, true, random.New())
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+func TestEdDSAVerifySmallOrder(t *testing.T) {
+	var smallOrderPk []byte = []byte{0xc7, 0x17, 0x6a, 0x70, 0x3d, 0x4d, 0xd8, 0x4f, 0xba, 0x3c, 0x0b,
+		0x76, 0x0d, 0x10, 0x67, 0x0f, 0x2a, 0x20, 0x53, 0xfa, 0x2c, 0x39,
+		0xcc, 0xc6, 0x4e, 0xc7, 0xfd, 0x77, 0x92, 0xac, 0x03, 0x7a}
+	ed := NewEdDSA(random.New())
+
+	msg := make([]byte, 32)
+	_, err := rand.Read(msg)
+	assert.NoError(t, err)
+
```
ineiti, line 191:
```suggestion
      msg := random.Bits(256, true, random.New())
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+	err = ed.Public.UnmarshalBinary(publicKey)
+	assert.Nil(t, err)
+
+	err = Verify(ed.Public, msg2, sig2)
+	assert.Equal(t, fmt.Errorf("signature is not canonical"), err)
+}
+
+// Test non-canonical keys
+// TODO check with p+2
```
ineiti, line 158: Can this check be easily added?

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+		0xaf, 0xc0, 0x8d, 0x93, 0x39, 0xd4, 0xee, 0xf6, 0x76, 0x57, 0x33,
+		0x36, 0xa5, 0xc5, 0x1e, 0xb6, 0xf9, 0x46, 0xb3, 0x1d,
+	}
+	publicKey := []byte{
+		0x7d, 0x4d, 0x0e, 0x7f, 0x61, 0x53, 0xa6, 0x9b, 0x62, 0x42, 0xb5,
+		0x22, 0xab, 0xbe, 0xe6, 0x85, 0xfd, 0xa4, 0x42, 0x0f, 0x88, 0x34,
+		0xb1, 0x08, 0xc3, 0xbd, 0xae, 0x36, 0x9e, 0xf5, 0x49, 0xfa}
+
+	err = ed.Public.UnmarshalBinary(publicKey)
+	assert.Nil(t, err)
```
ineiti, line 151:
```suggestion
	require.NoError(ed.Public.UnmarshalBinary(publicKey))
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+
+	err = Verify(ed.Public, msg2, sig2)
+	assert.Equal(t, fmt.Errorf("signature is not canonical"), err)
```
ineiti, line 154:
```suggestion
	require.Equal(t, fmt.Errorf("signature is not canonical"), err)
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+		0xa2, 0xde, 0xf9, 0xde, 0x14, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10}
+	var c uint16 = 0
+
+	ed := NewEdDSA(random.New())
+
+	msg := make([]byte, 32)
+	_, err := rand.Read(msg)
+	assert.NoError(t, err)
+
```
ineiti, line 119:
```suggestion
	msg := random.Bits(256, true, random.New())
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+	sig, err := ed.Sign(msg)
+	assert.Nil(t, err)
```
ineiti, line 121:
```suggestion
	require.NoError(t, err)
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+	assert.Nil(t, Verify(ed.Public, msg, sig))
```
ineiti, line 122:
```suggestion
	require.NoError(t, Verify(ed.Public, msg, sig))
```

sign/eddsa/eddsa_test.go
```diff
@@ -101,12 +102,110 @@ func TestEdDSASigning(t *testing.T) {
...
+
+	// Add l to signature
+	for i := 0; i < 32; i++ {
+		c += uint16(sig[32+i]) + L[i]
+		sig[32+i] = byte(c)
+		c >>= 8
+	}
+
+	err = Verify(ed.Public, msg, sig)
+	assert.Equal(t, fmt.Errorf("signature is not canonical"), err)
```
ineiti, line 132:
```suggestion
	require.Equal(t, fmt.Errorf("signature is not canonical"), err)
```

gnarula:
> I would replace all `assert` with `require`, as most of the time it doesn't make sense to continue the test, e.g., if the `msg := random.Bits...` fails, no reason to continue the test.
> 
> Also, there are some `assert.Nil` which should be `require.NoError`.
> 

Yeah, I'm yet to work my way through `eddsa_test`. Realised it quite late yesterday after I created the PR and hence the TODO commit/messages in haste :)

> I thought that `HasSmallOrder` is defined on the point and doesn't need to take a slice of bytes, no?

I used the byte slice because `point.MarshalBinary()` returns values `modulo prime` and we have `prime` and `prime+1` in weak keys.

> 
> Also, `IsCanonical` could be defined only in the eddsa package, as it doesn't make sense in the `edwards25519` package, given that `Point`s and `Scalar`s always are canonical.

👍

ineiti:
> > I thought that `HasSmallOrder` is defined on the point and doesn't need to take a slice of bytes, no?
> 
> I used the byte slice because `point.MarshalBinary()` returns values `modulo prime` and we have `prime` and `prime+1` in weak keys.

Yes - I still don't really understand what this check does with all the XORs :( This is where I would've liked to have an input from Simone, to know what makes most sense. From an engineering point of view `HasSmallOrder([]byte)` is a static method, which doesn't belong in a structure, but should be kept in the package. But as it only applies to `ed25519.point`s, it is cleaner to be defined on the structure, but then it shouldn't be a static method, so it should use the structure.
But then, as you write, we will never have the `prime` and `prime+1` as weak keys.

Is it correct to replace `prime` and `prime+1` with `0` and `1`? As we're always counting modulo `prime`?

gnarula:
> But then, as you write, we will never have the `prime` and `prime+1` as weak keys.

We might run into that at https://github.com/dedis/kyber/pull/430/files#diff-bd9209ce75feb92351db6c68bd8010d7R154

sign/eddsa/eddsa.go
```diff
@@ -119,12 +139,24 @@ func (e *EdDSA) Sign(msg []byte) ([]byte, error) {
...
+// VerifyWithChecks uses a public key buffer, a message and a signature.
+// It will return nil if sig is a valid signature for msg created by
+// key public, or an error otherwise. Compared to `Verify`, it performs
+// additional checks around the canonicality and ensures the public key
+// does not have a small order.
+func VerifyWithChecks(pub, msg, sig []byte) error {
 	if len(sig) != 64 {
 		return fmt.Errorf("signature length invalid, expect 64 but got %v", len(sig))
 	}
+	if (sig[63]&240) > 0 && !group.Scalar().(edDSAScalar).IsCanonical(sig[32:]) {
```
ineiti, line 151: The `(sig[63]&240) > 0` should also go in `IsCanonical`.

ineiti:
> > But then, as you write, we will never have the `prime` and `prime+1` as weak keys.
> 
> We might run into that at https://github.com/dedis/kyber/pull/430/files#diff-bd9209ce75feb92351db6c68bd8010d7R154

This doesn't make sense: if the keys are supposed to be `< prime`, then how can you test against `prime` and `prime+1` weak keys?

OK, crypto.stackexchange to the rescue:

https://crypto.stackexchange.com/questions/55632/libsodium-x25519-and-ed25519-small-order-check

So, if I get that right, the second-last paragraph has a nice summary:

```
The blacklist is the list of all points that will give zero as the ‘shared secret’. It does not list all illegitimate public keys—there are too many to list, of the form 𝐵+𝑄 where 𝐵 is a legitimate public key (of which there are 𝑝1 possibilities) and 𝑄 is on the blacklist. The only way to detect legitimate public keys is to confirm that [𝑝1]𝐵=0.
``` 

So if we test for `p` and `p+1` in non-reduced numbers, they will fall on `0` and `1` anyway, which are already tested. Which makes me quite convinced that we can:

- from an engineering point of view: define `HasSmallOrder` on `ed25519.point`
- from a mathematical point of view: only check the first 5 weak keys, as `p` and `p+1` will be caught by the reduction and the check with `0` and `1`
- from a verification point of view: by `UnmarshalBinary`ing the `[]byte` in `eddsa.VerifyWithChecks` _after_ we check the `IsCanonical`

But I don't see yet where we should/could check for the correct subgroup with `prime * Point = 0`, like the following does: https://github.com/kilic/bls12-381/blob/59713d4652a1851eb2ad0fb980533a5472294fe0/g1.go#L280

That's probably for another PR.

gnarula:
> This doesn't make sense: if the keys are supposed to be `< prime`, then how can you test against `prime` and `prime+1` weak keys?

Aha, I was under the impression that `R = sig[:32]` could be `> prime` but I read [RFC8032](https://tools.ietf.org/html/rfc8032#section-5.1.7) again and it does indeed mention that verification should fail if R cannot be decoded. [Decoding rules](https://tools.ietf.org/html/rfc8032#section-5.1.3) require failure for non canonical points. This is what you suggest in the third bullet point as well. 👍 Working on the changes now

PR #432, opened 2020-08-06
Fixing schnorr.Verify

calctopian:
This PR fixes the [latest PR](https://github.com/dedis/kyber/pull/430) because it didn't fix schnorr.Verify as previously [suggested in PR #427](https://github.com/dedis/kyber/pull/427):

- Added schnorr.VerifyWithChecks(g kyber.Group, pub, msg, sig []byte) which checks constraints as imposed by RFC8032 and group order checks
- Expanded the test TestEdDSACompatibility to check the reverse compatibility (i.e., an EdDSA signature can be checked by schnorr.Verify).
- Refactored tests and functions to leave them in their proper place

Fixes #431

PR #434, opened 2020-08-12
Schnorr Verification Checks

### LINKED ISSUE #431: schnorr.Verify must be fixed
As it was [previously](https://github.com/dedis/kyber/issues/426#issuecomment-663552533) [pointed out](https://github.com/dedis/kyber/pull/427/commits/ef5d2755799454565b7a647b15b3d4b4c453c467), schnorr.Verify must be fixed but the [previous commit](https://github.com/dedis/kyber/commit/88245de13231e54fa5863b219deae246fe0e7961)  didn't fix it.

But why is it so important?

 - EdDSA is Schnorr instantiated with the Ed25519 curve: if you fix EdDSA, then you must also fix Schnorr when instantiated with the Ed25519 curve.
 - eddsa.Verify is hardly used (IdentityProxy), however schnorr.Verify is used everywhere: dedis/cothority (darcs, viewchanges, roster extension, authproxy, evoting, ...) and dedis/dela. The vulnerability is on schnorr.Verify, leaving it unpatched is irresponsible.
 - schnorr.Verify could verify signatures generated by third-party libraries that don't conform to the keys/signatures generated by this library: a defensive approach must be taken to prevent malicious signatures.
 - in kyber, the implementations of schnorr.Verify and eddsa.Verify are [identical by design](https://github.com/dedis/kyber/blob/master/sign/schnorr/schnorr.go#L9) (i.e., what schnorr signs is verified by eddsa, et vice versa) except for eddsa.Sign having [deterministic nonce generation](https://github.com/dedis/kyber/blob/master/sign/eddsa/eddsa.go#L97) and schnorr.Sign[ being random](https://github.com/dedis/kyber/blob/master/sign/schnorr/schnorr.go#L35): as pointed out in the paper, "Ed25519 uses deterministic signing which removes the need for fresh random numbers during the signing process. **This choice does not lead to any particular consequences for our security analysis** since we model the key derivation function as a random oracle." ( [Section 4.2.5](https://eprint.iacr.org/2020/823.pdf)
 - in fact, there is a [test checking the compatibility between schnorr and eddsa signatures](https://github.com/dedis/kyber/blob/master/sign/schnorr/schnorr_test.go#L53): thus, they both should receive equal treatment.

gnarula:
Added `schnorr.VerifyWithChecks` which checks if the scalars and
points are canonical and ensures the points do not have a small
order for Edwards25519 Group.

Builds on top of #432 and closes #431

Asserts the group type and uses EdDSA verification for `*edwards25519.SuiteEd25519` group.

### DIFF
--- a/group/edwards25519/point.go
+++ b/group/edwards25519/point.go
@@ -263,3 +263,30 @@ func (P *point) HasSmallOrder() bool {
 
 	return (k>>8)&1 > 0
 }
+
+// IsCanonical determines whether the group element is canonical
+//
+// Checks whether group element s is less than p, according to RFC8032§5.1.3.1
+// https://tools.ietf.org/html/rfc8032#section-5.1.3
+//
+// Taken from
+// https://github.com/jedisct1/libsodium/blob/4744636721d2e420f8bbe2d563f31b1f5e682229/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c#L1113
+//
+// The meethod accepts a buffer instead of calling `MarshalBianry` on the receiver
+// because that always returns a value modulo `prime`.
+func (P *point) IsCanonical(s []byte) bool {
+	if len(s) != 32 {
+		return false
+	}
+
+	c := (s[31] & 0x7f) ^ 0x7f
+	for i := 30; i > 0; i-- {
+		c |= s[i] ^ 0xff
+	}
+
+	// subtraction might underflow
+	c = byte((uint16(c) - 1) >> 8)
+	d := byte((0xed - 1 - uint16(s[0])) >> 8)
+
+	return 1-(c&d&1) == 1
+}
--- a/group/edwards25519/point.go
+++ b/group/edwards25519/point.go
@@ -272,7 +272,7 @@ func (P *point) HasSmallOrder() bool {
 // Taken from
 // https://github.com/jedisct1/libsodium/blob/4744636721d2e420f8bbe2d563f31b1f5e682229/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c#L1113
 //
-// The meethod accepts a buffer instead of calling `MarshalBianry` on the receiver
+// The method accepts a buffer instead of calling `MarshalBinary` on the receiver
 // because that always returns a value modulo `prime`.
 func (P *point) IsCanonical(s []byte) bool {
 	if len(s) != 32 {
--- a/group/edwards25519/scalar.go
+++ b/group/edwards25519/scalar.go
@@ -2230,3 +2230,37 @@ func scReduce(out *[32]byte, s *[64]byte) {
 	out[30] = byte(s11 >> 9)
 	out[31] = byte(s11 >> 17)
 }
+
+// IsCanonical whether scalar s is in the range 0<=s<L as required by RFC8032, Section 5.1.7.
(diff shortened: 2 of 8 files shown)


nikkolasg:
This PR seems to induce an uncessary unmarhsalling of the public point and as well it seems that now you're binding the signature scheme to the underlying curve - so any other curve will not benefit from a potential malleability check which is imo suboptimal.
But i guess if it's cothority focused it's fine :D

group/edwards25519/scalar.go
```diff
@@ -2230,3 +2230,37 @@ func scReduce(out *[32]byte, s *[64]byte) {
 	out[30] = byte(s11 >> 9)
 	out[31] = byte(s11 >> 17)
 }
+
+// IsCanonical whether scalar s is in the range 0<=s<L as required by RFC8032, Section 5.1.7.
```
jeffallen, line 2234:
```suggestion
// IsCanonical whether the scalar in sb is in the range 0<=s<L as required by RFC8032, Section 5.1.7.
```

group/edwards25519/point.go
```diff
@@ -263,3 +263,30 @@ func (P *point) HasSmallOrder() bool {
...
+
+// IsCanonical determines whether the group element is canonical
+//
+// Checks whether group element s is less than p, according to RFC8032§5.1.3.1
+// https://tools.ietf.org/html/rfc8032#section-5.1.3
+//
+// Taken from
+// https://github.com/jedisct1/libsodium/blob/4744636721d2e420f8bbe2d563f31b1f5e682229/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c#L1113
+//
+// The meethod accepts a buffer instead of calling `MarshalBianry` on the receiver
```
jeffallen, line 275:
```suggestion
// The method accepts a buffer instead of calling `MarshalBinary` on the receiver
```

group/edwards25519/scalar.go
```diff
@@ -2230,3 +2230,37 @@ func scReduce(out *[32]byte, s *[64]byte) {
...
+// Also provides Strong Unforgeability under Chosen Message Attacks (SUF-CMA)
+// See paper https://eprint.iacr.org/2020/823.pdf for definitions and theorems
+// See https://github.com/jedisct1/libsodium/blob/4744636721d2e420f8bbe2d563f31b1f5e682229/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c#L2568
+// for a reference.
+// The method accepts a buffer instead of calling `MarshalBinary` on the receiver since that
+// always returns values modulo `primeOrder`
```
jeffallen, line 2240:
```suggestion
// always returns values modulo `primeOrder`.
```

jeffallen:
We need to test this change with Cothority. We need to replay the Byzcoin chain, because it uses these signatures in view changes, and evoting unit tests, which use these signatures in shuffles.

jeffallen:
Update: @gnarula has found some ByzCoin transactions with signer identities of "ed25519:0100000000000000000000000000000000000000000000000000000000000000", which is a public key with small order. It's also obviously not a valid public key. We are investigating how those transactions got posted. We need a solution to make sure that ByzCoin stays stable even after we check this change in to Kyber.

ineiti:
What is the reason `schnorr.Verify` does the verification now, instead of `eddsa.Verify`?

PR #449, opened 2020-11-26
Add go.mod usage example to front page

jeffallen:


README.md
```diff
@@ -32,20 +32,31 @@ to time. If you need something that doesn't change in a backward-compatible
...
-
-	go get go.dedis.ch/kyber
-
-You can recursively test all the packages in the library as follows:
-
-	go test -v ./...
+Using the module
+----------------
+
+Kyber support Go modules, and currently have a major version of 3, which means that
```
pierluca, line 38:
```suggestion
Kyber supports Go modules, and currently has a major version of 3, which means that
```

pierluca (approved):
I would argue Kyber is singular, other than that LGTM :)

PR #452, opened 2021-04-26
Support bls12-381 in pairing

wangyifan:
1. The curve code is from https://github.com/kilic/bls12-381
2. This is based on the work from https://github.com/drand/kyber-bls12381/ but modified to make the coding style consistent between bn256 and bls12381. 
3. Tests in sign/bls and sign/tbls are updated to support testing both bn256 and bls12381
4. This PR is related to issue https://github.com/dedis/kyber/issues/402

### DIFF
--- a/pairing/bls12381/group.go
+++ b/pairing/bls12381/group.go
@@ -1,0 +1,120 @@
+package bls12381
+
+import (
+	"crypto/cipher"
+	"crypto/sha256"
+	"hash"
+
+	"go.dedis.ch/kyber/v3"
+	"go.dedis.ch/kyber/v3/util/random"
+	"go.dedis.ch/kyber/v3/xof/blake2xb"
+)
+
+// GroupChecker allows to verify if a Point is in the correct group or not. For
+// curves which don't have a prime order, we need to only consider the points
+// lying in the subgroup of prime order. That check returns true if the point is
+// correct or not.
+type GroupChecker interface {
+	kyber.Point
+	IsInCorrectGroup() bool
+}
+
+type common struct {
+	isPrime bool
+}
+
+func (g *common) Scalar() kyber.Scalar {
+	return NewKyberScalar()
+}
+
+func (g *common) ScalarLen() int {
+	return g.Scalar().MarshalSize()
+}
+
+func (g *common) IsPrimeOrder() bool {
+	return g.isPrime
+}
+
+func (g *common) Hash() hash.Hash {
+	return sha256.New()
+}
+
+// XOF returns a newlly instantiated blake2xb XOF function.
+func (g *common) XOF(seed []byte) kyber.XOF {
+	return blake2xb.New(seed)
+}
+
+// RandomStream returns a cipher.Stream which corresponds to a key stream from
+// crypto/rand.
+func (g *common) RandomStream() cipher.Stream {
+	return random.New()
+}
+
+type groupG1 struct {
+	common
+	*commonSuite
+}
+
+func (g *groupG1) String() string {
+	return "bls12381.G1"
+}
+
+func (g *groupG1) Point() kyber.Point {
+	return NullpointG1()
+}
+
+func (g *groupG1) PointLen() int {
+	return g.Point().MarshalSize()
+}
+
+type groupG2 struct {
+	common
+	*commonSuite
+}
+
+func (g *groupG2) String() string {
+	return "bls12381.G2"
+}
+
+func (g *groupG2) Point() kyber.Point {
+	return NullpointG2()
+}
+
+func (g *groupG2) PointLen() int {
+	return g.Point().MarshalSize()
+}
+
+type groupGT struct {
+	common
+	*commonSuite
+}
+
+func (g *groupGT) String() string {
+	return "bls12381.GT"
+}
+
+func (g *groupGT) Point() kyber.Point {
+	return newEmptyGT()
+}
+
+func (g *groupGT) PointLen() int {
+	return g.Point().MarshalSize()
+}
+
+func NewGroupG1() kyber.Group {
+	return &groupG1{
+		common: common{isPrime: true},
+	}
+}
+
+func NewGroupG2() kyber.Group {
+	return &groupG2{
+		common: common{isPrime: false},
+	}
+}
+
+func NewGroupGT() kyber.Group {
+	return &groupGT{
+		common: common{isPrime: false},
+	}
+}
--- a/pairing/bls12381/group.go
+++ b/pairing/bls12381/group.go
@@ -60,7 +60,7 @@ func (g *groupG1) String() string {
 }
 
 func (g *groupG1) Point() kyber.Point {
-	return NullpointG1()
+	return nullpointG1()
 }
 
 func (g *groupG1) PointLen() int {
@@ -77,7 +77,7 @@ func (g *groupG2) String() string {
 }
 
 func (g *groupG2) Point() kyber.Point {
-	return NullpointG2()
+	return nullpointG2()
 }
 
 func (g *groupG2) PointLen() int {
@@ -101,20 +101,14 @@ func (g *groupGT) PointLen() int {
 	return g.Point().MarshalSize()
 }
 
-func NewGroupG1() kyber.Group {
+func newGroupG1() kyber.Group {
 	return &groupG1{
 		common: common{isPrime: true},
 	}
 }
 
-func NewGroupG2() kyber.Group {
+func newGroupG2() kyber.Group {
 	return &groupG2{
-		common: common{isPrime: false},
-	}
-}
-
-func NewGroupGT() kyber.Group {
-	return &groupGT{
 		common: common{isPrime: false},
 	}
 }
--- a/pairing/bls12381/point.go
+++ b/pairing/bls12381/point.go
@@ -16,7 +16,7 @@ type pointG1 struct {
 	p *bls12381.PointG1
 }
 
-func NullpointG1() *pointG1 {
+func nullpointG1() *pointG1 {
 	var p bls12381.PointG1
 	return newPointG1(&p)
 }
@@ -88,7 +88,7 @@ func (k *pointG1) Neg(a kyber.Point) kyber.Point {
 
 func (k *pointG1) Mul(s kyber.Scalar, q kyber.Point) kyber.Point {
 	if q == nil {
-		q = NullpointG1().Base()
+		q = nullpointG1().Base()
 	}
 	bls12381.NewG1().MulScalarBig(
 		k.p,
@@ -154,7 +154,7 @@ type pointG2 struct {
 	p *bls12381.PointG2
 }
 
-func NullpointG2() *pointG2 {
+func nullpointG2() *pointG2 {
 	var p bls12381.PointG2
 	return newPointG2(&p)
 }
@@ -226,7 +226,7 @@ func (k *pointG2) Neg(a kyber.Point) kyber.Point {
 
 func (k *pointG2) Mul(s kyber.Scalar, q kyber.Point) kyber.Point {
 	if q == nil {
-		q = NullpointG2().Base()
+		q = nullpointG2().Base()
 	}
 	bls12381.NewG2().MulScalarBig(
 		k.p,
--- a/pairing/bls12381/point.go
+++ b/pairing/bls12381/point.go
@@ -1,0 +1,403 @@
+package bls12381
+
+import (
+	"crypto/cipher"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
+
+	bls12381 "github.com/kilic/bls12-381"
+	"go.dedis.ch/kyber/v3"
+	"go.dedis.ch/kyber/v3/group/mod"
+)
+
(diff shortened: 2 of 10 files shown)


pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,365 @@
...
+
+// Code extracted from kyber/utils/test
+// TODO: expose API in forked drand/kyber
+// Apply a generic set of validation tests to a cryptographic Group,
+// using a given source of [pseudo-]randomness.
+//
+// Returns a log of the pseudorandom Points produced in the test,
+// for comparison across alternative implementations
+// that are supposed to be equivalent.
+func testGroup(t *testing.T, g kyber.Group, rand cipher.Stream) []kyber.Point {
```
ineiti, line 21: Long time I didn't look at the code, so I might be missing something obvious: why did you copy the test here, instead of adding the new suite to the list of suites to be tested?

wangyifan: Can you point me to where the list of suites to be tested? It's my first time submitting code to this repo. I might miss something obvious as well.

ineiti: OK, got it wrong: in fact the suites themselves call the tests:

- group/edwards25519/curve_test.go
- group/nist/group_test.go

But I'm not sure it works with pairing-based suites, as some of the methods of the `Point` and `Scalar` interface might not be defined.

So what I'm proposing is something like this: instead of copying the lines from the `util/test/test.go` file, do something like:

```
func TestG1(t *testing.T) {
	test.GroupTest(t, newGroupG1())
}
```

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+	return 48
+}
+
+func (k *pointG1) String() string {
+	b, _ := k.MarshalBinary()
+	return "bls12-381.G1: " + hex.EncodeToString(b)
+}
+
+func (k *pointG1) Hash(m []byte) kyber.Point {
+	p, _ := bls12381.NewG1().HashToCurve(m, Domain)
```
nkcr, line 138: @ineiti should we `panic()` if there is an error ?

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+
+}
+
+func (k *pointG1) IsInCorrectGroup() bool {
+	return bls12381.NewG1().InCorrectSubgroup(k.p)
+}
+
+// Domain comes from the ciphersuite used by the RFC of this name compatible
+// with the paired library > v18
+var Domain = []byte("BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_")
```
nkcr, line 150: should be placed at the top

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+	buf := make([]byte, k.MarshalSize())
+	n, err := io.ReadFull(r, buf)
+	if err != nil {
+		return n, err
+	}
+	return n, k.UnmarshalBinary(buf)
+}
+
+func (k *pointG2) MarshalSize() int {
+	return 96
```
nkcr, line 267: should be in a constant (at the top)

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+	return &pointGT{
+		f: f,
+	}
+}
+
+func (k *pointGT) Equal(kk kyber.Point) bool {
+	return k.f.Equal(kk.(*pointGT).f)
+}
+
+const gtLength = 576
```
nkcr, line 308: should be at the top

pairing/bls12381/suite_test.go
nkcr: I tries and that works. So yes, please use the functions from `util/test/`.

nkcr (changes requested):
Thank you for your contribution and this valuable work !
We would be happy to merge your PR, here you will find some comments more related to the "engineering" side. @si-co will take a closed look from the crypto perspective.

In the mean time if you could address the issues mentioned that would be awesome.

pairing/bls12381/group.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (g *common) ScalarLen() int {
+	return g.Scalar().MarshalSize()
+}
+
+func (g *common) IsPrimeOrder() bool {
+	return g.isPrime
+}
+
+func (g *common) Hash() hash.Hash {
```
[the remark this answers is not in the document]
ineiti: From what I remember, the `Suite` should have these, and not the `Group`.

pairing/bls12381/point.go
ineiti: Hmm - good question. I don't see the `Hash` defined in `/group.go`, so I don't know where the function signature comes from. 

But if it's not possible to return an error, and the point cannot be created, then I'd call `panic()` here, yes. Better than continuing with an invalid Point!

sign/bls/bls_test.go
```diff
@@ -6,23 +6,49 @@ import (
...
+	testBLSFailBatchVerify(t, suite)
+	testBinaryMarshalAfterAggregationIssue400(t, suite)
+}
+
+func TestBLS12381(t *testing.T) {
+	suite := bls12381.NewSuite()
+	testBLS(t, suite)
+	testBLSFailSig(t, suite)
+	testBLSFailKey(t, suite)
+	testBLSAggregateSignatures(t, suite)
```
calctopian, line 33: Thank you very much for your interest and contribution.

However, there is a security issue: BLS12-381 is also vulnerable to rogue public-key attacks. You shouldn't use AggregateSignatures() and AggregatePublicKeys() from sign/bls as you are currently testing here.

Instead, you should move these testBLSAggregateSignatures(), testBLSFailAggregatedSig() and testBLSFailAggregatedKey() to sign/bdn

sign/bls/bls_test.go
```diff
@@ -6,23 +6,49 @@ import (
...
+	testBLSFailAggregatedSig(t, suite)
+	testBLSFailAggregatedKey(t, suite)
+	testBLSBatchVerify(t, suite)
+	testBLSFailBatchVerify(t, suite)
+	testBinaryMarshalAfterAggregationIssue400(t, suite)
+}
+
+func testBLS(t *testing.T, suite pairing.Suite) {
+	msg := []byte("Hello Boneh-Lynn-Shacham")
 	private, public := NewKeyPair(suite, random.New())
```
calctopian, line 43: This comment is about a general design decision affecting the whole library, not just this Pull Request.

In BLS curves, G1 is small and fast and G2 is large and slow. Why? G1 are pairs of integers and G2 are pairs of complex integers, thus G2 takes twice the storage and requires more computation.
In practice, points are compressed for storage and transmission by dropping the y-coordinate, halving the amount of data. For BLS12-381, G1 points are halved from 96 bytes to 48 bytes, and G2 points are halved from 192 bytes to 96 bytes.

In this library, G1 is used for signatures and G2 is used for public keys: in other words, in dedis/kyber signatures require 48 bytes and public keys require 96 bytes.
This is the reverse of how BLS signatures are used in other blockchains, which swap your choice of G1 for G2. It's recommended in [draft-boneh-bls-signature-00](https://datatracker.ietf.org/doc/html/draft-boneh-bls-signature-00#section-2):

     2.  (minimizing public key size) Put public keys in G1 and signatures
       in G2.  This latter case comes up when we do signature
       aggregation, where most of the communication costs come from
       public keys.  **This is particularly relevant in applications such
       as blockchains** and compressing certificate chains, where the goal
       is to minimize the total size of multiple public keys and
       aggregated signatures.

and in a later edition of [draft-irtf-cfrg-bls-signature-04](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.1):

    2.  Minimal-pubkey-size: public keys are points in G1, signatures are
       points in G2.

       **Implementations using signature aggregation SHOULD use this
       approach, since the size of (PK_1, ..., PK_n, signature) is
       dominated by the public keys even for small n.**

For example, in Ethereum 2.0 they choose to minimise the public key size as recommended.

calctopian: This affects not only this Pull Request, but also ByzCoin, skipchain, and dedis/dela.

Your blockchains will be faster and lighter if you swap G1 for G2.

nikkolasg: Just want to point out that drand's fork (the one from which this work is derived) is modified to allow that flexibility; there's a general signing interface and BLS can be used on G1 or G2. See https://github.com/drand/kyber/blob/master/sign/bls/bls.go#L37 for example. (same for threshold BLS).

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+func (k *pointG1) Add(a, b kyber.Point) kyber.Point {
+	aa := a.(*pointG1)
+	bb := b.(*pointG1)
+	bls12381.NewG1().Add(k.p, aa.p, bb.p)
+	return k
+}
+
+func (k *pointG1) Sub(a, b kyber.Point) kyber.Point {
```
si-co, line 76: Should be changed accordingly to @nkcr's suggested change for `Add`. Same for the remaining functions.

pairing/bls12381/point.go
si-co: The same function signature is used for `bn256`, so no error is returned. I would `panic` in case of error, as it is done in `bn256` to avoid carrying around an invalid point.

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+
+func newPointG2(p *bls12381.PointG2) *pointG2 {
+	return &pointG2{p: p}
+}
+
+func (k *pointG2) Equal(k2 kyber.Point) bool {
+	return bls12381.NewG2().Equal(k.p, k2.(*pointG2).p)
+}
+
+func (k *pointG2) Null() kyber.Point {
```
si-co, line 170: As before, this and the subsequent functions should modify the point in place.

pairing/bls12381/point.go
si-co: Instead of using an additional constant, this can be simply computed as 2 times the `MarshalSize` on G1.

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+
+func (k *pointGT) Sub(a, b kyber.Point) kyber.Point {
+	aa := a.(*pointGT)
+	bb := b.(*pointGT)
+	bls12381.NewGT().Sub(k.f, aa.f, bb.f)
+	return k
+}
+
+func (k *pointGT) Neg(q kyber.Point) kyber.Point {
+	panic("bls12-381: GT is not a full kyber.Point implementation")
```
si-co, line 350: This and the next function are available in `bls12-381`, so I would implement them.

pairing/bls12381/point.go
```diff
@@ -0,0 +1,403 @@
...
+	buf := make([]byte, k.MarshalSize())
+	n, err := io.ReadFull(r, buf)
+	if err != nil {
+		return n, err
+	}
+	return n, k.UnmarshalBinary(buf)
+}
+
+func (k *pointGT) MarshalSize() int {
+	return 576
```
si-co, line 385: As before, this should be a constant.

pairing/bls12381/suite.go
```diff
@@ -0,0 +1,175 @@
...
+func NewSuiteG1() *Suite {
+	s := NewSuite()
+	s.commonSuite.Group = &groupG1{commonSuite: &commonSuite{}, common: common{isPrime: true}}
+	return s
+}
+
+// NewSuiteG2 returns a G2 suite.
+func NewSuiteG2() *Suite {
+	s := NewSuite()
+	s.commonSuite.Group = &groupG2{commonSuite: &commonSuite{}, common: common{isPrime: true}}
```
si-co, line 44: `isPrime` should be false.

pairing/bls12381/suite.go
```diff
@@ -0,0 +1,175 @@
...
+	return s
+}
+
+// NewSuiteGT returns a GT suite.
+func NewSuiteGT() *Suite {
+	s := NewSuite()
+	s.commonSuite.Group = &groupGT{commonSuite: &commonSuite{}, common: common{isPrime: true}}
```
si-co, line 51: `isPrime` should be false.

pairing/bls12381/suite.go
```diff
@@ -0,0 +1,175 @@
...
+func (c *commonSuite) Write(w io.Writer, objs ...interface{}) error {
+	return fixbuf.Write(w, objs)
+}
+
+// Hash returns a newly instantiated sha256 hash function.
+func (c *commonSuite) Hash() hash.Hash {
+	return sha256.New()
+}
+
+// XOF returns a newlly instantiated blake2xb XOF function.
```
si-co, line 155: Typo: newlly --> newly

AnomalRoil:
Note that a maintained Kyber wrapper for Kilic/bls12-381 is already available here:
https://github.com/drand/kyber-bls12381

Furthermore, https://github.com/kilic/bls12-381 has been archived and therefore effort to support BLS12-381 will probably be put towards supporting Circl as in #487

PR #455, opened 2021-11-19
Fix a bug in method: String() of PriShare

seafooler:
It is weird that method `String()` of `PriShare` returns the pointer to V currently. 
Maybe, it would be better to reflect the data contents. Thus, we can compare if two variables of PriShare are equal, by simply comparing the results of `String()` method.

### DIFF
--- a/share/poly.go
+++ b/share/poly.go
@@ -40,7 +40,7 @@ func (p *PriShare) Hash(s kyber.HashFactory) []byte {
 }
 
 func (p *PriShare) String() string {
-	return fmt.Sprintf("{%d:%p}", p.I, p.V)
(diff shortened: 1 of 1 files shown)


ineiti:
As `p.V` is a `kyber.Scalar`, I'm not sure that the `String` method is always meaningful. Probably the best would be to use a `p.V.MarshalBinary` and convert it to a hex-string, and display that.

seafooler:
> As `p.V` is a `kyber.Scalar`, I'm not sure that the `String` method is always meaningful. Probably the best would be to use a `p.V.MarshalBinary` and convert it to a hex-string, and display that.

That is true. Thanks for your great suggestion. I'll amend it and repropose a new `pull request`.

PR #456, opened 2021-11-19
Fix a bug in method: String() of PriShare

seafooler:
Fix a bug to make `String()` method of `PriShare` more useful.
I have modified the codes based on @ineiti 's comments in `Pull Request` #455.

### DIFF
--- a/share/poly.go
+++ b/share/poly.go
@@ -41,7 +41,10 @@ func (p *PriShare) Hash(s kyber.HashFactory) []byte {
 }
 
 func (p *PriShare) String() string {
-	data, _ := p.V.MarshalBinary()
+	data, err := p.V.MarshalBinary()
+	if err != nil {
+		panic("couldn't marshal point")
+	}
 	return fmt.Sprintf("{%d:%s}", p.I, hex.EncodeToString(data))
 }
 
--- a/share/poly.go
+++ b/share/poly.go
@@ -13,6 +13,7 @@ import (
 	"crypto/cipher"
 	"crypto/subtle"
 	"encoding/binary"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"sort"
@@ -40,7 +41,8 @@ func (p *PriShare) Hash(s kyber.HashFactory) []byte {
 }
 
 func (p *PriShare) String() string {
-	return fmt.Sprintf("{%d:%p}", p.I, p.V)
+	data, _ := p.V.MarshalBinary()
+	return fmt.Sprintf("{%d:%s}", p.I, hex.EncodeToString(data))
 }
 
 // PriPoly represents a secret sharing polynomial.


share/poly.go
```diff
@@ -40,7 +41,8 @@ func (p *PriShare) Hash(s kyber.HashFactory) []byte {
 }
 
 func (p *PriShare) String() string {
-	return fmt.Sprintf("{%d:%p}", p.I, p.V)
+	data, _ := p.V.MarshalBinary()
```
ineiti, line 44:
```suggestion
	data, err := p.V.MarshalBinary()
	if err != nil {
		panic("couldn't marshal point")
	}
```

ineiti (changes requested):
Always do error-checking.

ineiti (approved):

ineiti:
@pierluca, @nkcr this looks OK to me. Do you want to merge this?

nkcr:
@ineiti I'm wondering if doing a `panic` in a `String` function is a good idea. As a user I would prefer that the returned string contains an ugly error instead of the function to panic.

share/poly.go
```diff
@@ -40,7 +41,11 @@ func (p *PriShare) Hash(s kyber.HashFactory) []byte {
 }
 
 func (p *PriShare) String() string {
-	return fmt.Sprintf("{%d:%p}", p.I, p.V)
+	data, err := p.V.MarshalBinary()
+	if err != nil {
+		panic("couldn't marshal point")
+	}
+	return fmt.Sprintf("{%d:%s}", p.I, hex.EncodeToString(data))
```
ineiti, line 48:
```suggestion
		return "wrong V-value"
	}
	return fmt.Sprintf("{%d:%x}", p.I, data)
```

seafooler: Both are also fine to me.

nkcr: In fact there is already a `String()` function for a `kyber.Scalar`.

So why not doing: ? 

```
return fmt.Sprintf("{%d:%s}", p.I, p.V)
```

seafooler: I have already talked about this with @ineiti before, as seen in #455. 

> As p.V is a kyber.Scalar, I'm not sure that the String method is always meaningful. Probably the best would be to use a p.V.MarshalBinary and convert it to a hex-string, and display that.

ineiti: Even better - I _did_ look in the `type Scalar`, but didn't follow to the `type Marshaling` where it's defined...

ineiti, line 48: Sorry - should've thought of this 1 week ago...

```suggestion
	return fmt.Sprintf("{%d:%s}", p.I, p.V)
```

ineiti:
OK, so I guess I messed this up...
- TIL: `%v` _does_ call `String()` if it is available - but only way down in https://pkg.go.dev/fmt. So usually I prefer to indicate the conversion I want directly.
- some of the `Point` implementations print the x/y co-ordinates, which might not be what you want
- I still prefer the `MarshalBinary` display, but probably it's easier to just go with the `%s` (which I prefer over `%v` as you explicitly state what you want and don't let the compiler decide)

One excuse I have is that I spent many days debugging passing of `Scalar`s and `Point`s between go, java, and javascript. And my favorite output ended up being the marshalled one, because it's the most interoperable and comparable.

So to say that I still prefer the `MarshalBinary` one, as it would've spared me some time during debugging!

share/poly.go
nkcr: Fine for me 👍

nkcr: Marked as resolved, but the commit wasn't included..? I'm fine in using the simple `%s` solution and think this should be the chosen approach.

seafooler: Using the simple `%s` is fine with me too.

nkcr: Ok. So please update the PR to just use `%s` and let's merge it.

seafooler: Thx a lot ! I'll close this PR and repropose a new one.

seafooler: I have proposed a new PR #459, plz help to merge it. Many thanks.

PR #457, opened 2021-11-23
Adds neff shuffling of sequences

nkcr:
Based on the work from @lhmerino (https://github.com/dedis/votegral/blob/master/apps/Cothority/votegral/lib/crypto/shuffle.go)


examples/neff_shuffle_test.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.dedis.ch/kyber/v3"
+	kproof "go.dedis.ch/kyber/v3/proof"
+	"go.dedis.ch/kyber/v3/shuffle"
+)
+
+/*
```
ineiti, line 12: We have all comments as `//` and don't use the `/*` syntax. So for uniformity I propose to change the comment format.

examples/neff_shuffle_test.go
```diff
@@ -0,0 +1,106 @@
...
+This example illustrates how to use the Neff shuffle protocol with simple,
+single pairs.
+*/
+func Test_Example_Neff_Shuffle_Simple(t *testing.T) {
+	numPairs := 3
+
+	// generate random pairs
+
```
ineiti, line 20: Uniformity of comments: don't have an empty line between the comment an the code.

examples/neff_shuffle_test.go
```diff
@@ -0,0 +1,106 @@
...
+	X := make([][]kyber.Point, numSequences)
+	Y := make([][]kyber.Point, numSequences)
+
+	// generate random sequences
+
+	for i := 0; i < numSequences; i++ {
+		xs := make([]kyber.Point, sequenceLen)
+		ys := make([]kyber.Point, sequenceLen)
+
+		for i := 0; i < sequenceLen; i++ {
```
ineiti, line 72: Using twice the same iterator `i`, please change one.

Here and in other places.

examples/neff_shuffle_test.go
```diff
@@ -0,0 +1,106 @@
...
+	e := make([]kyber.Scalar, NQ)
+	for j := 0; j < NQ; j++ {
+		e[j] = suite.Scalar().Pick(suite.RandomStream())
+	}
+
+	prover, err := getProver(e)
+	require.NoError(t, err)
+
+	proof, err := kproof.HashProve(suite, "SequencesShuffle", prover)
+
```
ineiti, line 97: Add error check

shuffle/sequences.go
```diff
@@ -0,0 +1,155 @@
...
+	// Pick a random permutation used in ALL k ElGamal sequences. The permutation
+	// (π) of an ElGamal pair at index i always outputs to the same index
+	pi := make([]int, k)
+	for i := 0; i < k; i++ {
+		pi[i] = i
+	}
+
+	// Fisher–Yates shuffle
+	for i := k - 1; i > 0; i-- {
+		j := int(randUint64(rand) % uint64(i+1))
```
ineiti, line 50: As the modulo introduces a small bias, it is better to use the `random.Int` method. Even though this only matters if k is big.

```suggestion
	for i := int64(k - 1); i > 0; i-- {
		j := random.Int(big.NewInt(i+1), rand).Int64()
```

shuffle/sequences.go
```diff
@@ -0,0 +1,155 @@
...
+		if j != i {
+			pi[i], pi[j] = pi[j], pi[i]
+		}
+	}
+
+	// Pick a fresh ElGamal blinding factor β(j, i) for each ElGamal sequence
+	// and each ElGamal pair
+	beta := make([][]kyber.Scalar, NQ)
+	for j := 0; j < NQ; j++ {
```
ineiti, line 59: Confusing to me: I usually start with `i` in the outer loop, and then go through `j`, `k`, ... in the inner loops. Except if the paper uses the same notation.

Here and in other places.

shuffle/shuffle_test.go
```diff
@@ -64,3 +70,82 @@ func shuffleTest(suite Suite, k, N int) {
...
+	H := suite.Point().Mul(h, nil)
+
+	// Create a set of ephemeral "client" keypairs to shuffle
+	c := make([]kyber.Scalar, k)
+	C := make([]kyber.Point, k)
+
+	for i := 0; i < k; i++ {
+		c[i] = suite.Scalar().Pick(rand)
+		C[i] = suite.Point().Mul(c[i], nil)
+		//		fmt.Println(" "+C[i].String())
```
ineiti, line 88: Remove debugging strings.

shuffle/shuffle_test.go
```diff
@@ -64,3 +70,82 @@ func shuffleTest(suite Suite, k, N int) {
...
+		// Do a key-shuffle
+		XX, YY, getProver := SequencesShuffle(suite, nil, H, X, Y, rand)
+
+		e := make([]kyber.Scalar, NQ)
+		for j := 0; j < NQ; j++ {
+			e[j] = suite.Scalar().Pick(suite.RandomStream())
+		}
+
+		prover, err := getProver(e)
+		if err != nil {
```
ineiti, line 132: Here and other tests for err: pass the `*testing.T` in this method, then do the 'normal' `require.NoError(t, err)`

shuffle/shuffle_test.go
```diff
@@ -64,3 +70,82 @@ func shuffleTest(suite Suite, k, N int) {
 		}
 	}
 }
+
+func sequenceShuffleTest(suite Suite, k, NQ, N int) {
```
ineiti, line 74: Pass the `*testing.B` in this method. Then you can use the `b.N`, and `require.NoError(b, err)`.

shuffle/shuffle_test.go
```diff
@@ -64,3 +70,82 @@ func shuffleTest(suite Suite, k, N int) {
...
+	r := suite.Scalar() // temporary
+	for j := 0; j < NQ; j++ {
+		for i := 0; i < k; i++ {
+			r.Pick(rand)
+			X[j][i] = suite.Point().Mul(r, nil)
+			Y[j][i] = suite.Point().Mul(r, H) // ElGamal blinding factor
+			Y[j][i].Add(Y[j][i], C[i])        // Encrypted client public key
+		}
+	}
+
```
ineiti, line 119: You might want to add a `b.ResetTimer()` here - depending on which part you want to benchmark this.

examples/neff_shuffle_test.go
```diff
@@ -0,0 +1,106 @@
...
+	// shuffle sequences
+
+	XX, YY, getProver := shuffle.SequencesShuffle(suite, nil, nil, X, Y, suite.RandomStream())
+
+	// compute the proof
+
+	NQ := len(X)
+	e := make([]kyber.Scalar, NQ)
+	for j := 0; j < NQ; j++ {
```
ineiti, line 89: Why is this `j` and not `i`? Except if the paper does otherwise, I would keep the top-level for-loops using `i`, 2nd-level `j`, aso. It makes it easier to read.

ineiti (changes requested):
Mostly style issues, but one cryptographic issue with the choice of the random number for the shuffle.

examples/neff_shuffle_test.go
nkcr: That's not the case for the majority of files in this package, which is why I used this comment format.

examples/neff_shuffle_test.go
nkcr: It follows the paper.

shuffle/sequences.go
nkcr: Mmm but then the function would use an external source of randomness, instead of the provided `rand`. Would that be okay ?

shuffle/sequences.go
nkcr: Yes, it follows the paper's notation. But still we could iterate over the rows first, I agree it's quite confusing, and probably not optimized.

shuffle/shuffle_test.go
nkcr: It can't. This function is used for the unit and benchmark tests, which is also why we use `panic` and not `require.NotError`. Nothing new though, it follows the same pattern as the existing tests.

shuffle/sequences.go
nkcr: Yes, it follows the paper's notation and I agree it's confusing. We choose to stick to the paper's notation.

shuffle/sequences.go
ineiti: Are you sure? From what I understand, it's using the `rand` here as an argument to `random.Int`. So it should be OK. Or are I'm missing something?

shuffle/sequences.go
ineiti: OK - didn't look at the paper. You could even add an URL to arxive, if it's in there.

examples/neff_shuffle_test.go
ineiti: True - but I'm really not used to see this in go. So keep it like this.

shuffle/sequences.go
nkcr: My bad, I overlooked your example.

ineiti (approved):

si-co (commented):
I don't see any error in the crypto, mostly stylistic comments: feel free to integrate them or ignore them if they aren't coherent with the code base.

si-co (approved):

PR #460, opened 2022-03-16
Test for 2/3 in privates order

jadeydi:


sign/tbls/tbls_test.go
```diff
@@ -28,4 +29,47 @@ func TestTBLS(test *testing.T) {
...
+	// tests for privates in order
+	n = 3
+	t = n/2 + 1
+	priv1 := suite.G2().Scalar().Pick(suite.RandomStream())
+	priv2 := suite.G2().Scalar().Pick(suite.RandomStream())
+
+	g2 := suite.G2()
+	base := suite.G2().Point().Base()
+	commits := make([]kyber.Point, t)
+	commits[0] = g2.Point().Mul(priv2, base)
```
ineiti, line 42: usually you can multiply with `nil` instead of giving the base point.

sign/tbls/tbls_test.go
```diff
@@ -15,7 +16,7 @@ func TestTBLS(test *testing.T) {
 	suite := bn256.NewSuite()
 	n := 10
 	t := n/2 + 1
-	secret := suite.G1().Scalar().Pick(suite.RandomStream())
+	secret := suite.G2().Scalar().Pick(suite.RandomStream())
```
ineiti, line 19: Are you sure? I didn't look into the code, but usually you need to 'mix' G1 and G2.

ineiti (commented):
Can you please elaborate what you're doing here and why you want to add this to the tests?

PR #461, opened 2022-06-08
Kyber Documentation

parinayc20:
I have tried to create an abstract API for the Kyber Suite, Group, Point, and Scalar. The docs are currently hosted at [https://dedis.github.io/kyber](https://dedis.github.io/kyber). I would like to get them reviewed by the engineers before merging them into the main branch.

Thanks,
Parinay Chauhan


nkcr (commented):
Looks nice !
Here are some high level comments:

- make sure to format the text at 80 chars for clarity (if you're on VS code, the `rewrap` plugin does the job for you)
- a minor detail, but also add an empty line after titles:
From...
```
# Title
## Sub title
Test
```
...to
```
# Title

## Subtitle

Text
```
- in the `Parameters` section of tables you are only repeating the name of the arguments, which we already have in the signature. Same for the return. Instead you should describe the argument(s) itself, something like:

|            |                                            |
| ---------- |--------------------------------------------|
| Function   | `Object.MarshalTo(io.Writer) (int, error)` |
| Parameters | - `io.Writer`: The destination writer to which the object will be marshalled                                |
| Output     | - `int`: Number of bytes written<br>- `error` : Contains an error if something unexpected happened, otherwise is nil                             |
- `marshalling` is missing an interface definition
- to help with navigation, I would add a `See also` section in the introduction when an interface is composed of another interface. For example:
---
The Suite interface is the sum total all suite mix-ins provided in kyber.
For example - `SuiteEd255219`, the suite for the edwards 25519 curve
```
type Suite interface {
	kyber.Encoding
	kyber.Group
	kyber.HashFactory
	kyber.XOFFactory
	kyber.Random
}
```
See also [Encoding](...), [Group](...), [HashFactory](...), ...

---

- You can update the README.md as in https://github.com/dedis/dela/tree/master/docs

PR #464, opened 2022-07-05
security fix: any participant could disable DKG

vibs29:
A deal with a bogus signature would still get stored (before producing
an error). That would prevent a properly signed deal coming later from
being stored. This fixes the bug.

The test was also written in a flawed way. It measured two failure
situations ("wrong index" and "wrong deal") after a success situation. But
written that far down, those measurements could well end up measuring
failures due to a duplicate deal instead. This changeset also patches
the test flaw.

### DIFF
--- a/share/dkg/rabin/dkg.go
+++ b/share/dkg/rabin/dkg.go
@@ -282,7 +282,6 @@ func (d *DistKeyGenerator) ProcessDeal(dd *Deal) (*Response, error) {
 		return nil, err
 	}
 
-	d.verifiers[dd.Index] = ver
 	resp, err := ver.ProcessEncryptedDeal(dd.Deal)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


ineiti:
Looks good to me, thanks a lot! 

You will have to sign the CLA before your PR can be merged, though.

nikkolasg:
(just a reminder that I have rewrote the whole DKG the the drand fork in a single unit (not separating with VSS) which is more secure, has been audited internally and externally, more readable, and provides high level APIs to use it securely https://github.com/drand/kyber/tree/master/share/dkg)

ineiti:
> I have rewrote the whole DKG the the drand fork in a single unit

Do you think it would make sense to add this directory to the `dedis/kyber` repo under `share/dkg/xxx`, with some to-be-found value for `xxx`?

ineiti:
> That would need some review and that may or may not be to the kyber maintainers's liking 🤷‍♂️

OK, that's @pierluca's call then. Even though C4DT could lend a hand starting September...

nkcr:
Thanks a lot!

PR #465, opened 2022-07-07
Replaces Travis with Github Action

nkcr:
- updates the Makefile 
- applies `gofmt` and `go mod tidy`
- `staticcheck` is commented out because there are a lot to fix


.github/workflows/go_lint.yml
```diff
@@ -0,0 +1,28 @@
+name: Go lint
+
+on:
+  push:
```
ineiti, line 4: We don't use the `push` entry here, but rather protect the repository against direct pushes to `master` (which should be renamed to `main` :) and only allow merges when the tests pass.

This avoids running an unnecessary lint when you merge.

.github/workflows/go_tests.yml
```diff
@@ -0,0 +1,64 @@
+name: Go test
+
+on:
+  push:
```
ineiti, line 4: same as above - you can remove `push` if the repo is configured accordingly.

.github/workflows/go_tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          parallel: true
+        
+  # notifies that all test jobs are finished.
+  finish:
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: shogo82148/actions-goveralls@v1
+        with:
+          parallel-finished: true
```
ineiti, line 64: missing newline

Makefile
```diff
@@ -1,5 +1,13 @@
-.DEFAULT_GOAL := test
+lint:
+	# Coding style static check.
+	@go install honnef.co/go/tools/cmd/staticcheck@latest
+	@go mod tidy
+	#staticcheck `go list ./...`
```
ineiti, line 5: Do you want to put the staticcheck in its own target? That would make more sense, IMO.

Makefile
```diff
@@ -1,5 +1,13 @@
...
-Coding/bin/Makefile.base:
-	git clone https://github.com/dedis/Coding
-include Coding/bin/Makefile.base
+vet:
+	go vet ./...
+
+# target to run all the possible checks; it's a good habit to run it before
+# pushing code
+check: lint vet
+	go test ./...
```
ineiti, line 13: missing newline

ineiti (approved):
Nice - some small comments, but else looks good.

Makefile
nkcr: I prefer having one target that gathers all vetting scripts. Except if there is a case where we need to launch staticcheck but not the other linting tools.

PR #468, opened 2022-08-11
adding the Certified failing test to dkg_test

Mahsa-Bastankhah:


share/dkg/pedersen/dkg_test.go
```diff
@@ -1333,3 +1333,207 @@ func TestUserOnlyFlagFalseBehavior(t *testing.T) {
 
 	require.False(t, dkg1.dealer.PrivatePoly().Secret().Equal(dkg2.dealer.PrivatePoly().Secret()))
 }
+
+// check whether the nodes can certify at the end when one node is common.
+// almost the same as TestDKGResharingNewNodes except that this test calls dkg.Certified
+func TestDKGResharingNewNodesCertifyCheck(t *testing.T) {
```
ineiti, line 1339: Why not add the tests for `Certified` to the old test? If you really want to do two separate tests, then you could have a method that does the test, and then you call it twice: once without the call to `Certified`, and once with a follow-up call to `Certified`.

But IMO you should be OK to simply add the `Certified` to the old test.

nkcr: I agree, you don't need to duplicate the whole test and can just add `Certified` to the old test.

PR #471, opened 2022-08-18
Pedersen: Fixes Responses not sent while resharing

### LINKED ISSUE #469: Old dkg nodes doesn't certify after re-sharing
I added a new test to [dkg_test ](https://github.com/Mahsa-Bastankhah/kyber/blob/c3e2b5b7c6a3f6423bf03c3ec85e5f4c859cfc44/share/dkg/pedersen/dkg_test.go#L1339)in the PR #468. In this test I create a set of dkg nodes and did the initial dkg setup for them then I create a new set of dkg nodes to reshare the secret of the old nodes among the new dkg nodes. There is one common node between the old committee and the new committee(`oldDkgs[4]` , `newDkgs[0]`). The test is essentially the same as [this already existing test](https://github.com/Mahsa-Bastankhah/kyber/blob/c3e2b5b7c6a3f6423bf03c3ec85e5f4c859cfc44/share/dkg/pedersen/dkg_test.go#L901) except that it calls `dkg.Certified()` in the end to make sure that all the nodes certify the resharing. Surprisingly all the nodes of the old committee except for the one that is common between the new and the old committee fail in certifying. The new committee nodes however successfully certify. And even more surprising is that when I remove the common node the certify is successful for all the old and new dkg nodes so apparently the one common node is causing the problem. I tried to track the problem  to find its source. This is what I found out:
In resharing protocol old nodes are supposed to create a deal for each new node. [In line](https://github.com/Mahsa-Bastankhah/kyber/blob/c3e2b5b7c6a3f6423bf03c3ec85e5f4c859cfc44/share/dkg/pedersen/dkg_test.go#L1420) `oldDkgs[4]` creates one deal less than the other old nodes. In fact since `oldDkgs[4]` and `newDkgs[0]` are the same, `oldDkgs[4]` only creates deals for `newDkgs[1:5]`. Later in resharing protocol new nodes after receiving the deals create a respond and broadcast this response but `newDkgs[0]` never broadcast its respond to the deal received from  `oldDkgs[4]` to the other old dkg nodes. In fact, [line](https://github.com/Mahsa-Bastankhah/kyber/blob/c3e2b5b7c6a3f6423bf03c3ec85e5f4c859cfc44/share/dkg/pedersen/dkg_test.go#L1420) implicitly adds this respond to the respond sets of `newDkgs[0]` and later [in line](https://github.com/Mahsa-Bastankhah/kyber/blob/c3e2b5b7c6a3f6423bf03c3ec85e5f4c859cfc44/share/dkg/pedersen/dkg_test.go#L1439) this respond is implicitly added to the response set of `newDkgs[1:5]` but this single response is never added to the respond set of the old committee members. Each node has `oldN` aggregators/verifiers and each of them should have `newN` responses however [in line](https://github.com/Mahsa-Bastankhah/kyber/blob/b8734881a60b55845eda692c35a109a77ffdd527/share/dkg/pedersen/dkg_test.go#L1508) I am printing the number of responses that the old nodes receive and you can observe that they miss one response. In another word old nodes are not receiving the response that `newDkgs[0]` creates in respond to the deal that `oldDkgs[4]` builds.

nkcr:
This PR updates the Deals generation and processing to fix the case where an old committee would not receive responses from staying nodes during a resharing. It would then prevent the old committee from being certified.

Previously, the `Deals` function would not return the node's local deal, but rather process it locally. I added an additional condition that the node is not doing a resharing, otherwise it returns and not process its own deal. This allows for responses to be broadcast to the old committee.

I also made some minor formatting updates to the test, as well as changed the number of staying nodes from 1 to 2.

Closes #469.
Closes #468.

### DIFF
--- a/share/dkg/pedersen/dkg_test.go
+++ b/share/dkg/pedersen/dkg_test.go
@@ -897,35 +897,43 @@ func TestDKGResharingNewNodesThreshold(t *testing.T) {
 
 }
 
-// Test resharing to a different set of nodes with one common
+// Test resharing to a different set of nodes with two common.
 func TestDKGResharingNewNodes(t *testing.T) {
 	oldPubs, oldPrivs, dkgs := generate(defaultN, vss.MinimumT(defaultN))
 	fullExchange(t, dkgs, true)
 
 	shares := make([]*DistKeyShare, len(dkgs))
 	sshares := make([]*share.PriShare, len(dkgs))
+
 	for i, dkg := range dkgs {
 		share, err := dkg.DistKeyShare()
 		require.NoError(t, err)
 		shares[i] = share
 		sshares[i] = shares[i].Share
 	}
+
 	// start resharing to a different group
+
 	oldN := defaultN
 	oldT := len(shares[0].Commits)
 	newN := oldN + 1
 	newT := oldT + 1
 	newPrivs := make([]kyber.Scalar, newN)
 	newPubs := make([]kyber.Point, newN)
+
+	// old[-1], old[-2] = new[0], new[1]
 	newPrivs[0] = oldPrivs[oldN-1]
 	newPubs[0] = oldPubs[oldN-1]
-	for i := 1; i < newN; i++ {
+	newPrivs[1] = oldPrivs[oldN-2]
+	newPubs[1] = oldPubs[oldN-2]
+
+	for i := 2; i < newN; i++ {
 		newPrivs[i], newPubs[i] = genPair()
 	}
 
-	// creating the old dkgs and new dkgs
+	// creating the old dkgs
+
 	oldDkgs := make([]*DistKeyGenerator, oldN)
-	newDkgs := make([]*DistKeyGenerator, newN)
 	var err error
 	for i := 0; i < oldN; i++ {
 		c := &Config{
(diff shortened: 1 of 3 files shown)


share/dkg/pedersen/dkg_test.go
```diff
@@ -897,35 +897,43 @@ func TestDKGResharingNewNodesThreshold(t *testing.T) {
...
 	// start resharing to a different group
+
 	oldN := defaultN
 	oldT := len(shares[0].Commits)
 	newN := oldN + 1
 	newT := oldT + 1
 	newPrivs := make([]kyber.Scalar, newN)
 	newPubs := make([]kyber.Point, newN)
+
+	// old[-1], old[-2] = new[0], new[1]
```
ineiti, line 924: Isn't it the other way round?

```suggestion
	// new[0], new[1] = old[-1], old[-2]
```

ineiti (approved):
LGTM - so now I'm curious what other state is possible when `Deal` is called. Is it only the setting up and the resharing? And in the set-up phase, the nodes use their own deal, but not so in the resharing phase?

Perhaps you can add some lines to the documentation of `Deal`...

PR #473, opened 2022-09-09
bugfix share/dkg/rabin: allow n-t participants to be unavailable

### LINKED ISSUE #472: Bug: share/dkg/rabin: All N participants must be available
Hi. I'm having trouble generating a distributed key when some (allowed number of) participants is offline. Is my expectation wrong that some number are allowed to be offline? Otherwise, either I'm using the library incorrectly or it has a bug.

I've produced a small test file, threshold_test.go, that demonstrates the problem. It can simulate 1/7 participants being offline. It needs to be added to the share/dkg/rabin directory, from where you can then do "go test". If you set the index BAD to -1 (simulating all participants being online), you'll see the tests all pass. But if you set it to 1 (simulating the participant at that index being offline), then TestDKGSecretCommits2() fails, whereas I still expect it to pass. Shouldn't it pass even with BAD=1?

```
// share/dkg/rabin/threshold_test.go
package dkg

import (
	"testing"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
)

const GOOD int = 0
const BAD int = 1 /* index of dkg to simulate as offline, -1 will disable */

func TestDKGSecretCommits2(t *testing.T) {
	fullExchange2(t)
	dkg := dkgs[GOOD]
	_, err := dkg.SecretCommits()
	assert.Nil(t, err) //"dkg: can't give SecretCommits if deal not certified"
}

func fullExchange2(t *testing.T) {
	dkgs = dkgGen()
	// full secret sharing exchange
	// 1. broadcast deals
	resps := make([]*Response, 0, nbParticipants*nbParticipants)
	for i, dkg := range dkgs {
		if i != BAD {
			deals, err := dkg.Deals()
			require.Nil(t, err)
			for j, d := range deals {
				if j != BAD {
					resp, err := dkgs[j].ProcessDeal(d)
					require.Nil(t, err)
					require.Equal(t, true, resp.Response.Approved)
					resps = append(resps, resp)
				}
			}
		}
	}
	// 2. Broadcast responses
	for _, resp := range resps {
		for i, dkg := range dkgs {
			// ignore all messages from ourself
			if resp.Response.Index == dkg.index {
				continue
			}
			if i != BAD {
				j, err := dkg.ProcessResponse(resp)
				require.Nil(t, err)
				require.Nil(t, j)
			}
		}
	}
	// 3. make sure everyone has the same QUAL set
	for i, dkg := range dkgs {
		if i != BAD {
			dkg.SetTimeout() //line below fails otherwise
			require.True(t, dkg.Certified())
			for j, dkg2 := range dkgs {
				if j != BAD {
					require.True(t, dkg.isInQUAL(dkg2.index))
				}
			}
		}
	}
}
```

vibs29:
Fixes #472

### DIFF
--- a/share/dkg/rabin/dkg.go
+++ b/share/dkg/rabin/dkg.go
@@ -348,6 +348,7 @@ func (d *DistKeyGenerator) ProcessJustification(j *Justification) error {
 // SetTimeout triggers the timeout on all verifiers, and thus makes sure
 // all verifiers have either responded, or have a StatusComplaint response.
 func (d *DistKeyGenerator) SetTimeout() {
+	(*d).dealer.SetTimeout()
 	for _, v := range d.verifiers {
 		v.SetTimeout()
 	}
--- a/share/dkg/rabin/dkg.go
+++ b/share/dkg/rabin/dkg.go
@@ -306,7 +306,7 @@ func (d *DistKeyGenerator) ProcessDeal(dd *Deal) (*Response, error) {
 func (d *DistKeyGenerator) ProcessResponse(resp *Response) (*Justification, error) {
 	v, ok := d.verifiers[resp.Index]
 	if !ok {
-		return nil, errors.New("dkg: complaint received but no deal for it")
+		return nil, errors.New("dkg: response received but no deal for it")
 	}
 
 	if err := v.ProcessResponse(resp.Response); err != nil {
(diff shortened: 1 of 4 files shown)


ineiti:
Thanks for this PR - from what I understand, it looks reasonable. It seems that, indeed, the `SetTimeout` is not correctly implemented, nor tested. Perhaps @Mahsa-Bastankhah has some ideas, too, as she just worked on this part of the code, and at one moment we also thought that we'd need `SetTimeout`.

You should also add some documentation to this part, to make it clear that the caller needs to call `SetTimeout` if they want to continue after a timeout happened.

Before we look at it, please make sure that all tests pass.

PR #475, opened 2022-09-23
Embedding G1 for the bn256 package

ineiti:
Creating a PR for #383 so that all tests can pass.

@taybart - if you want changes, create a PR against this PR, and I'll merge it.

Sorry for the difficulties with sonarcloud.

### DIFF
--- a/examples/bn256_enc_test.go
+++ b/examples/bn256_enc_test.go
@@ -1,0 +1,86 @@
+package examples
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


taybart:
I would love git blame credit but if it gets merged I would be super happy. Thanks!

ineiti:
> I would love git blame credit but if it gets merged I would be super happy. Thanks!

I co-authored you in the commit - but unfortunately that doesn't give you git-blame-credits, sorry!

PR #478, opened 2022-11-25
permission should run on all platforms

jbsv:
it may happen that the job runs on windows or macos, before the permission job has run on ubuntu, which leads to a failed CI, even though the 'safe PR' label has been set.
This PR aims to correct this behaviour.


.github/workflows/go_tests.yml
```diff
@@ -28,13 +33,9 @@ jobs:
         echo "::error:: Could not start CI tests due to missing *safe PR* label."
         exit 1
 
-  test:
+  coverage:
```
pierluca, line 36: Maybe for clarity
```suggestion
  test_and_coverage:
```

.github/workflows/go_tests.yml
```diff
@@ -8,7 +8,12 @@ on:
 
 jobs:
   permission:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest, windows-latest]
```
pierluca, line 13: won't this lead to 3 comments being added to the PR ?

PR #482, opened 2023-02-09
Bumping go version for linter to 1.20

ineiti:
There seems to be an error when linting, due to an old go-version.


pierluca (approved):
LGTM !

PR #484, opened 2023-02-24
Make Rabin DKG work with two Suites.

kape1395:


ineiti:
Please give a short description of _why_ and _how_ you did this.

K1li4nL:
Hello and thanks for your contribution. Could you explain a bit why it would be desirable to have Rabin DKG work with two suites ? Some documentation would be necessary too.

ineiti:
I think you can close this. There has been no answer for 1 year, we don't know why they did it, and it breaks compatibility, because it changes the signature of an exported method.

PR #487, opened 2023-04-13
pairing: Add support for BLS12381 pairings using CIRCL library

### LINKED ISSUE #402: Is there a plan to add support for BLS12-381?

armfazh:
This PR adds the functionality of BLS12381 pairings using the CIRCL implementation as backend.

[CIRCL](https://github.com/cloudflare/circl/) implementation uses formally-verified implementation (using fiat-crypto) for the prime field arithmetic.

Fixes #402

### DIFF
--- a/pairing/bls12381/g1.go
+++ b/pairing/bls12381/g1.go
@@ -1,0 +1,94 @@
+package bls12381
+
+import (
+	"crypto/cipher"
+	"io"
+
+	circl "github.com/cloudflare/circl/ecc/bls12381"
+	"go.dedis.ch/kyber/v3"
+)
+
+var _ kyber.Point = &G1Elt{}
+
+type G1Elt struct{ inner circl.G1 }
+
+func (p *G1Elt) MarshalBinary() (data []byte, err error) { return p.inner.BytesCompressed(), nil }
+
+func (p *G1Elt) UnmarshalBinary(data []byte) error { return p.inner.SetBytes(data) }
+
+func (p *G1Elt) String() string { return p.inner.String() }
+
+func (p *G1Elt) MarshalSize() int { return circl.G1SizeCompressed }
+
+func (p *G1Elt) MarshalTo(w io.Writer) (int, error) {
+	buf, err := p.MarshalBinary()
+	if err != nil {
+		return 0, err
+	}
+	return w.Write(buf)
+}
+
+func (p *G1Elt) UnmarshalFrom(r io.Reader) (int, error) {
+	buf := make([]byte, p.MarshalSize())
+	n, err := io.ReadFull(r, buf)
+	if err != nil {
+		return n, err
+	}
+	return n, p.UnmarshalBinary(buf)
+}
+
+func (p *G1Elt) Equal(p2 kyber.Point) bool { x := p2.(*G1Elt); return p.inner.IsEqual(&x.inner) }
+
+func (p *G1Elt) Null() kyber.Point { p.inner.SetIdentity(); return p }
+
+func (p *G1Elt) Base() kyber.Point { p.inner = *circl.G1Generator(); return p }
+
+func (p *G1Elt) Pick(rand cipher.Stream) kyber.Point {
+	var buf [32]byte
+	rand.XORKeyStream(buf[:], buf[:])
+	p.inner.Hash(buf[:], nil)
+	return p
+}
+
+func (p *G1Elt) Set(p2 kyber.Point) kyber.Point { p.inner = p2.(*G1Elt).inner; return p }
+
+func (p *G1Elt) Clone() kyber.Point { return new(G1Elt).Set(p) }
+
+func (p *G1Elt) EmbedLen() int {
+	panic("bls12-381: unsupported operation")
+}
+
+func (p *G1Elt) Embed(data []byte, r cipher.Stream) kyber.Point {
+	panic("bls12-381: unsupported operation")
+}
+
+func (p *G1Elt) Data() ([]byte, error) {
+	panic("bls12-381: unsupported operation")
+}
+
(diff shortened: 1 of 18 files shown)


pairing/bls12381/g1.go
```diff
@@ -0,0 +1,94 @@
...
+	rand.XORKeyStream(buf[:], buf[:])
+	p.inner.Hash(buf[:], nil)
+	return p
+}
+
+func (p *G1Elt) Set(p2 kyber.Point) kyber.Point { p.inner = p2.(*G1Elt).inner; return p }
+
+func (p *G1Elt) Clone() kyber.Point { return new(G1Elt).Set(p) }
+
+func (p *G1Elt) EmbedLen() int {
```
ineiti, line 57: From what I remember, the `Embed` was always quite important in one way or another to make things work, for example with ElGamal encryption. But perhaps you don't want to encrypt with BLS pairings.

Or is this now replaced with `Hash`?

pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,404 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.dedis.ch/kyber/v3"
+	this "go.dedis.ch/kyber/v3/pairing/bls12381"
+	"go.dedis.ch/kyber/v3/util/random"
+)
+
+// Code extracted from kyber/utils/test
+// TODO: expose API in forked drand/kyber
```
ineiti, line 16: What is this about? Do we need this before merging? Or is this a post-merge operation? Which API?

pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,404 @@
...
+// Apply a generic set of validation tests to a cryptographic Group,
+// using a given source of [pseudo-]randomness.
+//
+// Returns a log of the pseudorandom Points produced in the test,
+// for comparison across alternative implementations
+// that are supposed to be equivalent.
+func testGroup(t *testing.T, g kyber.Group, rand cipher.Stream) []kyber.Point {
```
ineiti, line 23: Why can't you run this on `G1`, `G2`, and `GT`? Because they don't implement `Embed`?

pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,404 @@
...
+	type canCheckPrimeOrder interface {
+		IsPrimeOrder() bool
+	}
+	primeOrder := true
+	if gpo, ok := g.(canCheckPrimeOrder); ok {
+		primeOrder = gpo.IsPrimeOrder()
+	}
+
+	// Verify additive and multiplicative identities of the generator.
+	// TODO: Check GT exp
```
ineiti, line 76: This looks like something that you should do, no?

pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,404 @@
...
+	/*fmt.Println("Inverse of base")*/
```
ineiti, line 77: Remove the debugging statements. Here and below.

pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,404 @@
...
+			/*}*/
+		}()
+		p = g.Point().Pick(rand)
+		return
+	}
+
+	// Test randomly picked points
+	last := gen
+	for i := 0; i < 5; i++ {
+		// TODO fork kyber and make that an interface
```
ineiti, line 194: What is this TODO about?

pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,404 @@
...
+	abG := s.G1().Point().Mul(ab, nil)
+	bb.ResetTimer()
+	for i := 0; i < bb.N; i++ {
+
+		// e(aG, bG) = e(G,H)^(ab)
+		p1 := s.Pair(aG, bH)
+		// e((ab)G,H) = e(G,H)^(ab)
+		p2 := s.Pair(abG, s.G2().Point().Base())
+		if !p1.Equal(p2) {
+			panic("aie")
```
ineiti, line 326: Change panic message to something understandable.

pairing/bls12381/suite_test.go
```diff
@@ -0,0 +1,404 @@
...
+	a := s.G1().Scalar().Pick(s.RandomStream())
+	b := s.G2().Scalar().Pick(s.RandomStream())
+	aG := s.G1().Point().Mul(a, nil)
+	bH := s.G2().Point().Mul(b, nil)
+	ab := s.G1().Scalar().Mul(a, b)
+	abG := s.G1().Point().Mul(ab, nil)
+	bb.ResetTimer()
+	for i := 0; i < bb.N; i++ {
+		if !s.ValidatePairing(aG, bH, abG.Clone(), s.G2().Point().Base()) {
+			panic("aie")
```
ineiti, line 342: Same as above.

ineiti (changes requested):
If I understand this PR correctly, it will break existing users of the bn256 suite if they rely on the `Adapter`. Is that correct? If so, then this is a breaking change which should not go into the code without increasing the version from `v3` to `v4`.
Still assuming my understanding is correct, I'll put `adapter.go` back in the `pairing` directory.

I didn't go into many details, but looks good to me, except:
- some debugging statements left in the code
- `aie` as panic message is not understandable
- some `TODO`s where I'm not sure if they need to be implemented and what they mean
- the breaking change for `pairing/adapter`, if I got that correct

PR #490, opened 2023-06-27
refactor: Improve codebase

howjmay:
Improve error check and slice operations

### DIFF
--- a/share/dkg/pedersen/dkg.go
+++ b/share/dkg/pedersen/dkg.go
@@ -341,7 +341,10 @@ func (d *DistKeyGenerator) ProcessDeal(dd *Deal) (*Response, error) {
 		return nil, err
 	}
 
-	ver, _ := d.verifiers[dd.Index]
+	ver, ok := d.verifiers[dd.Index]
+	if !ok {
+		return nil, fmt.Errorf("missing verifiers")
+	}
 
 	resp, err := ver.ProcessEncryptedDeal(dd.Deal)
 	if err != nil {
--- a/share/dkg/pedersen/dkg_test.go
+++ b/share/dkg/pedersen/dkg_test.go
@@ -202,6 +202,7 @@ func TestDKGProcessResponse(t *testing.T) {
 	require.Nil(t, err)
 
 	resp12, err := rec.ProcessDeal(deals2[idxRec])
+	require.NoError(t, err)
 	require.NotNil(t, resp)
 	require.Equal(t, vss.StatusComplaint, resp12.Response.Status)
 	require.Equal(t, deals2[idxRec].Index, uint32(dkg2.nidx))
@@ -797,7 +798,7 @@ func TestDKGResharingNewNodesThreshold(t *testing.T) {
 		require.Equal(t, newDkgs[i].nidx, i)
 	}
 
-	//alive := oldT - 1
+	// alive := oldT - 1
 	alive := oldT
 	oldSelected := make([]*DistKeyGenerator, 0, alive)
 	selected := make(map[string]bool)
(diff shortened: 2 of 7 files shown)


share/dkg/pedersen/dkg.go
```diff
@@ -341,7 +341,10 @@ func (d *DistKeyGenerator) ProcessDeal(dd *Deal) (*Response, error) {
 		return nil, err
 	}
 
-	ver, _ := d.verifiers[dd.Index]
+	ver, ok := d.verifiers[dd.Index]
+	if !ok {
+		return nil, fmt.Errorf("verifiers not exists")
```
ineiti, line 346:
```suggestion
		return nil, fmt.Errorf("missing verifiers")
```

ineiti (changes requested):
LGTM, only the error message should be changed.

share/dkg/pedersen/dkg.go
howjmay: fixed

ineiti (approved):

howjmay:
Added!

share/poly.go
```diff
@@ -35,7 +35,7 @@ type PriShare struct {
 func (p *PriShare) Hash(s kyber.HashFactory) []byte {
 	h := s.Hash()
 	_, _ = p.V.MarshalTo(h)
-	_ = binary.Write(h, binary.LittleEndian, p.I)
+	_ = binary.Write(h, binary.LittleEndian, int32(p.I))
```
ineiti, line 38: Looking at it once more: could this be a breaking change? If the current `int` is 64 bit, now it will hash it as 32 bits.

It is definitely a good idea to cast the int here, I'm just not sure whether it will break protocols having used this with 64 bits. Any idea?

howjmay: I think `int` is architecture-dependent, which means both 32 and 64 bits should work. But it is definitely better to use a type with clear width

ineiti: Yes, that's what https://go.dev/tour/basics/11 says, too. So if a binary for a 64-bit platform used this library before the change, it will have used a 64 bit representation of `p.I` for the hashing. With this change it will use a 32 bit representation, and the comparison will fail, e.g., in a signature.

Same thing if you use `int64` - for a 32 bit platform, this might make it not working anymore. So I'm trying to figure out if somebody out there actually relies on this, and cannot come up with a good way of doing so :)

@pierluca or @jbsv, what is your take on this? Creating a new major version of kyber would be best, but it seems a bit too much, except if we do it for all hashes of ints.

howjmay: I am curious how about changing the datatype of `p.I` here? I think that should be the easier and safest too.

ineiti: Yes, hindsight is 20/20, that is what I would do now. But only after seeing problems like this and working for a couple of years with Rust :)

Now this change would be a breaking change, because then a lot of code would not even compile anymore. So let's wait for @pierluca 's return and see what he thinks. I'm currently thinking of keeping the `p.I` there without a cast. Which is wrong, but at least not in a breaking way!

If ever there is a version 4 of kyber, this is something that would need to be addressed at a moment...

pierluca: Apologies for the delay. First of all, thanks @howjmay for the good suggestions in this PR. Regarding this `int32`, I can only agree with @ineiti: it is a breaking change.
Furthermore, we're most commonly using this library on x64 systems on our end, so this would be very likely to break a few things.

This would indeed make sense for a version 4 of Kyber, for which right now we don't really have the resources. On the other hand, I think we're piling up quite a bit of changes that could justify a version 4, if the resources allow it on our side into 2024 (or there's interest on your side, @ineiti ), we should definitely plan for it.

ineiti: Added `v4` label :)

howjmay: @ineiti So I think if I remove this change, the rest of the change is ok for merging?

ineiti: Yes, here and on line 307. Both need to be removed.

ineiti (approved):
LGTM, I propose to merge these changes @pierluca 

In fact, @howjmay you should rebase against the latest base branch.

howjmay:
> LGTM, I propose to merge these changes @pierluca
> 
> In fact, @howjmay you should rebase against the latest base branch.

I just updated !

pierluca (approved):

PR #491, opened 2023-06-30
Add check for the dkg public key after resharing

### LINKED ISSUE #489: DKG Protocol - Public Key Change After Reshare
## DKG Protocol - Public Key Change After Reshare

**Description:**

I am currently working with the DKG protocol for threshold encryption and decryption as provided under the `share/dkg/pedersen` folder. 

While exploring the capabilities of the protocol, I've noticed that the DKG supports resharing of secrets, which is a helpful feature. However, it appears that the public key changes post-reshare, which is causing current threshold encryption fail in my application.

**Expected Behavior:**

Ideally, I would expect that the public key remains constant after the reshare operation. 

**Actual Behavior:**

Currently, the public key changes each time a reshare is performed.

**Steps to Reproduce:**

1. Initialize the DKG protocol
2. Perform the reshare operation
3. Observe the change in the public key

**Additional Information:**

If it is by design that the public key changes after a reshare, could you please explain the reasoning behind this? If there's a way to maintain the same public key post-reshare, I'd appreciate if you could guide me on how to do it. I noticed that checking public key remains the same is mentioned in https://github.com/dedis/kyber/blob/a008b2d45c5c8675d2f01d140b7232166757a013/share/dkg/pedersen/dkg_test.go#L657, but it is not implemented in the relevant code.

Looking forward to your support on this issue. 

Thank you in advance!

leetingo:
Check whether the DKG public key changes before and after resharing among the same group of nodes.

Closes #489

### DIFF
--- a/share/dkg/pedersen/dkg_test.go
+++ b/share/dkg/pedersen/dkg_test.go
@@ -668,6 +668,10 @@ func TestDKGResharing(t *testing.T) {
 	newSecret, err := share.RecoverSecret(suite, newSShares, thr, defaultN)
 	require.NoError(t, err)
 	require.Equal(t, oldSecret.String(), newSecret.String())
+	// 3.
+	for i := 0; i < len(dkgs); i++ {
+		require.Equal(t, shares[i].Public(), newShares[i].Public())
+	}
 }
 
 // Test resharing functionality with one node less


ineiti:
Please add a description in the PR.

TingoLee:
Added.

ineiti:
And you need to sign the CLA, if I read the failing tests correctly...

pierluca:
In particular, regarding the CLA, it seems the e-mail/username you used as part of your commit (litiange) is not associated with your GitHub account. This is why the CLA automation can't detect whether you've signed it already or not.

share/dkg/pedersen/dkg_test.go
```diff
@@ -667,6 +667,10 @@ func TestDKGResharing(t *testing.T) {
 	newSecret, err := share.RecoverSecret(suite, newSShares, thr, defaultN)
 	require.NoError(t, err)
 	require.Equal(t, oldSecret.String(), newSecret.String())
+	// 3.
+	for i := 0; i < len(dkgs); i++ {
+		require.Equal(t, shares[i].Public(), newShares[i].Public())
```
pierluca, line 672: the overall DKG public key should remain constant, not the public key of each share, right ?

ineiti: That's correct, but the `Public()` method returns the public key of the distributed key:

https://github.com/dedis/kyber/blob/a008b2d45c5c8675d2f01d140b7232166757a013/share/dkg/pedersen/structs.go#L24

So this should pass.

K1li4nL:
As stated in the related issue, this check has been added when we merged DRAND version of this pedersen dkg, thanks for your work !

PR #497, opened 2024-02-15
Increase test coverage

matteosz:
This PR increase the test coverage for the modules util/encoding and util/encryption

### DIFF
--- a/util/encoding/encoding_test.go
+++ b/util/encoding/encoding_test.go
@@ -2,6 +2,7 @@ package encoding
 
 import (
 	"bytes"
+	"io"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


util/encoding/encoding_test.go
```diff
@@ -60,3 +61,38 @@ func TestScalarHexString(t *testing.T) {
...
+	mockReader1 := &MockFailingReader{data: []byte("abc")}
+	_, err = ReadHexPoint(s, mockReader1)
+	require.Error(t, err, "Expected error when reader fails, but got nil")
+
+	// Test case: not enough bytes from stream
+	mockReader2 := &MockEmptyReader{data: []byte("abc")}
+	_, err = ReadHexPoint(s, mockReader2)
+	require.Error(t, err, "Expected error when not enough bytes from stream, but got nil")
+	if err.Error() != "didn't get enough bytes from stream" {
+		t.Errorf("Expected error message: didn't get enough bytes from stream, but got %s", err.Error())
```
AnomalRoil, line 96: This could be done using `require.ErrorContains`.

AnomalRoil (approved):
LGTM, just one comment.

util/encoding/encoding_test.go
matteosz: I used `require.EqualError` to exactly match it

AnomalRoil (dismissed):

util/random/rand_test.go
```diff
@@ -84,3 +85,59 @@ func TestIncorrectSize(t *testing.T) {
...
+		bitlen uint // input bit length
+		exact  bool // whether the exact bit length should be enforced
+	}{
+		{bitlen: 128, exact: false},
+		{bitlen: 256, exact: true},
+		{bitlen: 512, exact: false},
+		{bitlen: 1024, exact: true},
+	}
+
+	for _, tc := range testCases {
```
pierluca, line 100: Nice pattern, but to improve test results reporting you'll want to use the Go sub-test pattern:
https://go.dev/blog/subtests

```
    for _, tc := range testCases {
        t.Run(fmt.Sprintf("bitlen: %d exact: %s", tc.bitlen, tc.exact), func(t *testing.T) {
            ...
        })
    }
```

same obviously applies to the other test :)

util/encoding/encoding_test.go
```diff
@@ -60,3 +61,36 @@ func TestScalarHexString(t *testing.T) {
...
+}
+
+func (m *MockFailingReader) Read(p []byte) (n int, err error) {
+	return copy(p, m.data), io.EOF
+}
+func (m *MockEmptyReader) Read(p []byte) (n int, err error) {
+	return 0, nil
+}
+
+func TestReadHexPointError(t *testing.T) {
```
pierluca, line 80: I'm going to be annoying on a point that's not common in this codebase, that is a bit of a pet peeve of mine, but which I think is important nonetheless:

This test contains 3 different test cases. When possible, try to have independent unit tests.
- This way, when the test fails, it tells you directly that it's the "reader fails" that fails, you don't have to look up which line this failed at.
- Furthermore, if the 3rd test case passes, you won't be able to tell (from the test results) if everything's grouped in a single test.
- Finally, having separate tests makes them more (trivially) independent of one another, and potentially parallelizable (not that it matters here)

pierluca (commented):
Good increment, I've suggested some improvements :)

pierluca (approved):

AnomalRoil (approved):

PR #498, opened 2024-02-16
Enable linter in the CI

K1li4nL:
Golanci-lint has been added and most of the errors it found were fixed. (TODO and "deprecated" warnings were disabled)


Makefile
```diff
@@ -7,6 +7,8 @@ generate: tidy
 
 # Coding style static check.
 lint: tidy
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.56.2
```
ineiti: I wouldn't put this in the Makefile target, as this would have side-effects every time somebody calls `make lint` on his local machine.

K1li4nL: It is fixed now !

group/curve25519/curve.go
```diff
@@ -81,6 +81,41 @@ func (c *curve) NewKey(stream cipher.Stream) kyber.Scalar {
 	return secret
 }
 
+func initBasePoint(c *curve, self kyber.Group, p *Param, fullGroup bool, base point) {
```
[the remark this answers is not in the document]
K1li4nL: Yes you are right, fixed !

PR #500, opened 2024-03-04
Added fuzzer tests

matteosz:


sign/cosi/cosi_test.go
```diff
@@ -28,15 +28,31 @@ func (m *cosiSuite) RandomStream() cipher.Stream { return m.r }
 var testSuite = &cosiSuite{edwards25519.NewBlakeSHA256Ed25519(), blake2xb.New(nil)}
 
 func TestCoSi(t *testing.T) {
-	testCoSi(t, 2, 0)
-	testCoSi(t, 5, 0)
-	testCoSi(t, 5, 2)
-	testCoSi(t, 5, 4)
+	message := []byte("Hello World Cosi")
+	testCoSi(t, 2, 0, message)
+	testCoSi(t, 2, 0, message)
```
pierluca, line 33: why repeat this twice ?

matteosz: Yeah probably I should have misclicked a ctrl+v

sign/tbls/tbls_test.go
```diff
@@ -10,22 +10,39 @@ import (
...
 	suite := bn256.NewSuite()
-	n := 10
-	t := n/2 + 1
+	th := n/2 + 1
+
 	secret := suite.G1().Scalar().Pick(suite.RandomStream())
-	priPoly := share.NewPriPoly(suite.G2(), t, secret, suite.RandomStream())
+	priPoly := share.NewPriPoly(suite.G2(), th, secret, suite.RandomStream())
 	pubPoly := priPoly.Commit(suite.G2().Point().Base())
 	sigShares := make([][]byte, 0)
```
AnomalRoil, line 35: I wonder if this setup phase really needs to be repeated when fuzzing or whether fuzzing should occur on the message only, WDYT?

sign/schnorr/schnorr_test.go
```diff
@@ -125,3 +126,19 @@ func TestSchnorrMalleability(t *testing.T) {
...
 	assert.Error(t, err, "schnorr signature malleable")
 }
+
+func FuzzSchnorr(f *testing.F) {
+	f.Fuzz(func(t *testing.T, msg []byte) {
+		if (len(msg) < 1) || (len(msg) > 1000) {
+			t.Skip("msg must have byte length between 1 and 1000")
+		}
+		suite := edwards25519.NewBlakeSHA256Ed25519()
+		kp := key.NewKeyPair(suite)
```
AnomalRoil, line 136: this should probably be outside of the fuzzing method since it doesn't need to be re-instantiated for each fuzzing pass

sign/cosi/cosi_test.go
```diff
@@ -28,15 +28,31 @@ func (m *cosiSuite) RandomStream() cipher.Stream { return m.r }
...
+		if f < 0 || f >= n {
+			t.Skip("f must be between 0 and n-1")
+		}
+
+		testCoSi(t, n, f, msg)
+	})
+}
 
+func testCoSi(t *testing.T, n, f int, message []byte) {
 	// Generate key pairs
```
AnomalRoil, line 56: key generation probably needs to happen prior to fuzzing here too

sign/tbls/tbls_test.go
matteosz: I was thinking that fuzzing on the parameter n would have been more interesting than using just; I'd say that the cost of repeating these operations is not that expensive and leads some value in increased exploration. But if you feel it's relevant to fuzz only on n I can change it

sign/cosi/cosi_test.go
matteosz: Key generation uses the value of n though, which is different for every fuzzer

sign/tbls/tbls_test.go
AnomalRoil: In general, fuzzing should be repeatable. Given the same "fuzzing corpus of input messages" I should get the same outputs/results in all runs.

Could we change this code so that there is no randomness outside of the fuzzer msg? 
Maybe seed the `RandomStream` using `msg`.

sign/schnorr/schnorr_test.go
```diff
@@ -125,3 +126,19 @@ func TestSchnorrMalleability(t *testing.T) {
 	err = Verify(suite, kp.Public, msg, s)
 	assert.Error(t, err, "schnorr signature malleable")
 }
+
+func FuzzSchnorr(f *testing.F) {
+	suite := edwards25519.NewBlakeSHA256Ed25519()
+	kp := key.NewKeyPair(suite)
```
AnomalRoil, line 132: Could you please use a hardcoded keypair for fuzz tests?

sign/cosi/cosi_test.go
AnomalRoil: Actually having a random keypair in the fuzz tests is an issue, see other comments. We need either:
- an hardcoded keypair for fuzzing
- a seeded, deterministic randomness stream for fuzz tests.
-

AnomalRoil (changes requested):

sign/tbls/tbls_test.go
```diff
@@ -7,25 +7,45 @@ import (
...
+		}
+		if (len(msg) < 1) || (len(msg) > 1000) {
+			t.Skip("msg must have byte length between 1 and 1000")
+		}
+		TBLSRoutine(t, msg, n)
+	})
+}
+
+func TBLSRoutine(test *testing.T, msg []byte, n int) {
 	suite := bn256.NewSuite()
```
AnomalRoil, line 30: I think you also need to assign the below deterministic stream to the suite's stream to avoid having the Suite default to proper randomness.

AnomalRoil (changes requested):

AnomalRoil (approved):
LGTM. Should we wait on the big merge first?

pierluca (approved):
LGTM

AnomalRoil (approved):

PR #501, opened 2024-03-11
Add some benchmarks

K1li4nL:


encrypt/ecies/ecies_test.go
```diff
@@ -44,3 +48,36 @@ func TestECIESFailCiphertext(t *testing.T) {
...
+			var ct []byte
+			b.Run("Encrypt/"+suite.String(), func(b *testing.B) {
+				ct, _ = Encrypt(suite, public, message, nil)
+
+			})
+
+			b.Run("Decrypt/"+suite.String(), func(b *testing.B) {
+				_, _ = Decrypt(suite, private, ct, nil)
+			})
+		}
```
AnomalRoil, line 81: This looks like the for loops should be in the sub-benchmarks, rather than in the outer one.

proof/proof_test.go
```diff
@@ -246,3 +248,64 @@ func Example_or2() {
...
+
+			sval["x"+index] = s
+			predicateBuilder = append(predicateBuilder, "x"+index)
+			predicateBuilder = append(predicateBuilder, "B"+index)
+			pval["B"+index] = suite.Point().Base()
+
+			P = suite.Point().Add(P, publicPoint)
+		}
+
+		pval["P"] = P
```
AnomalRoil, line 285: this looks like it should be in a sub-benchmark, no?

proof/proof_test.go
```diff
@@ -246,3 +248,64 @@ func Example_or2() {
...
+		})
+
+		b.Run(suite.String()+"/ProofVerify", func(b *testing.B) {
+			verifier := pred.Verifier(suite, pval)
+			err = HashVerify(suite, "TEST", verifier, proof)
+			if err != nil {
+				b.Log(err.Error())
+				b.Fail()
+			}
+		})
```
AnomalRoil, line 309: These are missing for loops in the subbenchmarks

AnomalRoil (changes requested):

encrypt/ecies/ecies_test.go
K1li4nL: Fixed !

proof/proof_test.go
K1li4nL: Fixed !

proof/proof_test.go
K1li4nL: I did not put it in a sub-benchmark because this wasn't really the part I wanted measured, only the time for the prover to build the actual proof and the time the verifier takes, what do you think ?

AnomalRoil: I meant the entire for `i := 0; i < b.N; i++ {` block, not just the `pval` setting.

Why keep it in the "upper function" instead of having it in its own sub-benchmark?

Issue #441, opened 2020-09-07
The blake2b XOF is not standard


cgrigis:
_Security Code Review observation:_

The way the XOF is built from a keyed `blake2b` instance in Kyber is unusual:
it is taking the `seed`, and if the seed is bigger than `blake2b` key size, it is splitting it into a `seed1` value of the key size and an extra `seed2`  value that is then fed to the state of the `blake2b` instance using `Write`. 

This means that if one were to implement the `Reset()` operation on that XOF, it might be difficult to feed again the `seed2` value to the state, since the original `seed` is not stored after initialization and only `seed1` is stored to.

PR #506, opened 2024-03-26
Fix issue #441

K1li4nL:
Attempt to fix issue #441 by adding the original seed to the XOF struct and defining Reset()

### DIFF
--- a/xof/blake2xb/blake.go
+++ b/xof/blake2xb/blake.go
@@ -66,8 +66,10 @@ func (x *xof) Reseed() {
 }
 
 func (x *xof) Reset() {
-	y := New(x.seed)
-	x.impl = y.(*xof).impl
+	x.impl.Reset()
+	if len(x.seed) > blake2b.Size {
+		x.impl.Write(x.seed[blake2b.Size:])
+	}
 }
 
 func (x *xof) XORKeyStream(dst, src []byte) {
--- a/xof/blake2xb/blake.go
+++ b/xof/blake2xb/blake.go
@@ -9,6 +9,7 @@ import (
 
 type xof struct {
 	impl blake2b.XOF
+	seed []byte
 	// key is here to not make excess garbage during repeated calls
 	// to XORKeyStream.
 	key []byte
@@ -33,7 +34,11 @@ func New(seed []byte) kyber.XOF {
 			panic("blake2b.XOF.Write should not return error: " + err.Error())
 		}
 	}
-	return &xof{impl: b}
+	seedCopy := make([]byte, len(seed))
+	copy(seedCopy, seed)
+	x := xof{impl: b, seed: seedCopy}
+
+	return &x
 }
 
 func (x *xof) Clone() kyber.XOF {
@@ -58,6 +63,11 @@ func (x *xof) Reseed() {
 	x.Read(x.key)
 	y := New(x.key)
 	// Steal the XOF implementation, and put it inside of x.
+	x.impl = y.(*xof).impl
+}
+
+func (x *xof) Reset() {
+	y := New(x.seed)
 	x.impl = y.(*xof).impl
 }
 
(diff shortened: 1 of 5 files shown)


xof/blake2xb/blake.go
```diff
@@ -33,7 +34,11 @@ func New(seed []byte) kyber.XOF {
 			panic("blake2b.XOF.Write should not return error: " + err.Error())
 		}
 	}
-	return &xof{impl: b}
+	seedCopy := make([]byte, len(seed))
+	copy(seedCopy, seed)
+	x := xof{impl: b, seed: seedCopy}
+
+	return &x
```
AnomalRoil, line 41:
```suggestion
	return &xof{impl: b, seed: seedCopy}
```

xof/blake2xb/blake.go
```diff
@@ -61,6 +66,11 @@ func (x *xof) Reseed() {
 	x.impl = y.(*xof).impl
 }
 
+func (x *xof) Reset() {
+	y := New(x.seed)
+	x.impl = y.(*xof).impl
```
AnomalRoil, line 71: This is making too many useless allocations.
Please, refactor `New` to use `Reset` after having done the `blake2b.NewXOF` step,  and use the fact that `impl` has its own `Reset` function if you can.

xof/blake2xs/blake.go
```diff
@@ -61,6 +66,11 @@ func (x *xof) Reseed() {
 	x.impl = y.(*xof).impl
 }
 
+func (x *xof) Reset() {
+	y := New(x.seed)
```
AnomalRoil, line 70: Same as above. Use Reset in New, not the other way around.

xof/keccak/keccak.go
```diff
@@ -36,6 +39,11 @@ func (x *xof) Reseed() {
 	x.sh.Write(x.key)
 }
 
+func (x *xof) Reset() {
+	x.sh.Reset()
+	x.sh.Write(x.seed)
```
AnomalRoil, line 44: yes, this is nice

AnomalRoil (changes requested):

xof/blake2xb/blake.go
K1li4nL: Should be fixed now !

xof/blake2xs/blake.go
K1li4nL: Should be fixed now !

xof/xof_test.go
```diff
@@ -214,6 +215,43 @@ func testReseed(t *testing.T, s kyber.XOFFactory) {
 	}
 }
 
+func TestResetNoSeed(t *testing.T) {
+	for _, impl := range impls {
```
AnomalRoil, line 219: I think this is only testing `blake2xb`, no? Could you add `blake2xs` to the list of `impls` too?

AnomalRoil (changes requested):

xof/blake2xs/blake.go
```diff
@@ -61,6 +65,13 @@ func (x *xof) Reseed() {
 	x.impl = y.(*xof).impl
 }
 
+func (x *xof) Reset() {
+	x.impl.Reset()
+	if len(x.seed) > blake2s.Size {
+		x.impl.Write(x.seed[blake2s.Size:])
+	}
+}
```
AnomalRoil, line 73: Is it really necessary to store the entire seed then? Wouldn't storing just `seed2` be enough and test here if its len is bigger than 0?

AnomalRoil (changes requested):
Okay, one last comment I guess

xof/blake2xs/blake.go
K1li4nL: Now that I am not using New() in Reset() as you suggested, you are right we can do that. Fixed !

AnomalRoil (approved):
lgtm

PR #509, opened 2024-04-01
Merging the drand fork

Robingoumaz:
This PR is about merging the drand fork back into the dedis dyber.

Many files only got minimal changes, (comments, typos, moved to another folder) and some had bigger changes applied to them.


AnomalRoil:
We are trying to reconcile the drand/kyber fork with the upstream dedis/kyber, would you mind signing DEDIS CLAs ? 

- [ ] @nikkolasg
- [ ] @CluEleSsUK
- [ ] @armfazh
- [ ] @kevincharm
- [ ] @matejpavlovic

kevincharm:
Done!

nikkolasg:
Done !

xof/blake2xs/blake.go
```diff
@@ -22,18 +23,21 @@ func New(seed []byte) kyber.XOF {
...
-			panic("blake2s.XOF.Write should not return error: " + err.Error())
-		}
+	_, err = b.Write(seed2)
+	if err != nil {
+		panic("blake2s.XOF.Write should not return error: " + err.Error())
 	}
-	return &xof{impl: b}
+
+	seedCopy := make([]byte, len(seed2))
+	copy(seedCopy, seed2)
```
AnomalRoil, line 38: Aren't these changes part of another PR already?

xof/keccak/keccak.go
AnomalRoil, line 1: Weren't these changes part of another PR?

encrypt/ibe/ibe_test.go
```diff
@@ -1,5 +1,6 @@
 package ibe
 
+/*
```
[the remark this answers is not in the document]
Robingoumaz: I uncommented them and used the circl bls12381 instead but there is a difference between the scalar of circl bls and the one of the other pairing curve which makes it not pass the tests. I am trying to change the circl implementation so that it uses mod.Int

ineiti:
This is a huge PR. Unless you want to keep up with the drand branch of kyber, I propose to split it in different PRs:

- one for changes in comments / typos
- one PR per existing and new module

This would make it more manageable, but would also make it difficult for future exchanges between the drand fork and dedis/kyber.

AnomalRoil:
@ineiti the goal is to get rid of the drand/kyber fork and only have one unique kyber repo :D

ineiti:
> @ineiti the goal is to get rid of the drand/kyber fork and only have one unique kyber repo :D

Then I would definitely pack this up in several easy-to-digest PRs.

share/dkg/dkg.go
AnomalRoil, line 1: Here I think you should stick with the file structure from dedis/kyber and keep both Pedersen and Rabin subfolders in `share/dkg/` instead of deleting Rabin and replacing everything with `drand` implementation.

share/dkg/rabin/dkg.go
AnomalRoil, line 1: See my comment about keeping both pedersen and rabin packages, but having the drand dkg move to the pedersen folder rather than deleting and replacing with drand version

sign/dss/dss.go
AnomalRoil, line 1: Why deleting it?

util/random/rand.go
```diff
@@ -66,19 +66,21 @@ func (r *randstream) XORKeyStream(dst, src []byte) {
...
 		if err != nil {
 			nerr++
+			errors = append(errors, err.Error())
 		}
 		b.Write(buff[:n])
 	}
 
 	// we are ok with few sources being insecure (i.e., providing less than
 	// readerBytes bytes), but not all of them
 	if nerr == len(r.Readers) {
```
AnomalRoil, line 82: you could get rid of `nerr` and use `len(errors)` instead here.

sign/test/threshold.go
AnomalRoil, line 1: This doesn't seem to conform to the `_test` nomenclature for Go tests and functions in it also don't use the proper `TestXYZ` format.

Please replace to use the proper Go test framework.

sign/test/scheme.go
AnomalRoil, line 1: same remark about it not being a proper test package as per the Go test nomenclature. Need `_test.go` in file name and `TestName` functions

sign/bls/bls_test.go
AnomalRoil, line 1: Why commenting it all?

share/dkg/dkg.go
AnomalRoil: Okay, so @nikkolasg told me actually that the Pedersen `dedis/kyber` is not the same implementation as the `drand/kyber`, the latter being the secure one, while the former might be somewhat faulty.

So I think we should 100% replace the `dedis/kyber/share/dkg/pedersen` with the `drand/kyber/share/dkg` code.
I don't have strong opinion about keeping or not the Rabin one, but he told me that one is very slow, which means it's inferior to Pedersen.
We could either:
 - keep both and make sure to benchmark them to warn users cc @matteosz 
 - remove Rabin and keep only the drand one, which is what this PR is currently doing.
 
 Preferences @pierluca ?

ineiti: I vote for having both, and adding a benchmark is a good idea. Except if keeping the kyber dkg is a lot of work.

Thinking of kyber as a research library, so it's nice to have various algorithms for comparisons, or in case somebody has a brilliant idea to update one of the algorithms.

encrypt/ibe/ibe_test.go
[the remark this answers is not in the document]
armfazh: The conversion is not needed if rejection sampling is updated as in: 
https://github.com/Robingoumaz/kyber-drand/pull/1

pierluca:
I see that the tests failed, which means Sonar wasn't run.
It might be worth it to bypass the tests once (in the CI) to let Sonar run and gather some feedback.
Sonar might be able to make a first pass to spot problematic issues.
Alternatively, you could also setup Sonar to analyze your own repo and provide a pointer here :)

sign/test/threshold.go
AnomalRoil: To do in a next PR

sign/test/scheme.go
AnomalRoil: To do in a next PR

PR #510, opened 2024-04-21
Add hash to point from RFC9380

K1li4nL:
Add Hash() on Ed25519


group/edwards25519/point.go
```diff
@@ -290,3 +294,240 @@ func (P *point) IsCanonical(s []byte) bool {
...
+	feMul(&gx1, &gx1, &x1n)
+	feAdd(&gx1, &gx1, &tv2)
+	feMul(&gx1, &gx1, &x1n)
+	feSquare(&tv3, &gxd)
+	feSquare(&tv2, &tv3)
+	feMul(&tv3, &tv3, &gxd)
+	feMul(&tv3, &tv3, &gx1)
+	feMul(&tv2, &tv2, &tv3)
+
+	// compute y11 = tv2 ^ c4
```
AnomalRoil, line 448: like this, but for each step

group/edwards25519/point.go
```diff
@@ -290,3 +294,240 @@ func (P *point) IsCanonical(s []byte) bool {
...
+	}
+
+	for i := 0; i < len(dst); i++ {
+		dst[i] = b1[i] ^ b2[i]
+	}
+
+	return dst, nil
+}
+
+func curve25519Elligator2(u fieldElement) (xn, xd, yn, yd fieldElement) {
```
AnomalRoil, line 411: Please add inline comments for each of the Elligator2 operations. Did you implement [these ones](https://datatracker.ietf.org/doc/html/rfc9380#ell2-opt)?

AnomalRoil, line 411: Can't this also be used for Curve25519? Why put it in Edwards25519?

AnomalRoil (changes requested):
Please, add better comments to refer to which part of the RFC you've implemented, that will make review easier.

group/edwards25519/point_test.go
```diff
@@ -62,3 +87,165 @@ func Test_PointIsCanonical(t *testing.T) {
...
+		res, err := expandMessageXMD(h, []byte(inputsTestVectRFC9380[i]), dst, outputLength[1])
+		resHex := hex.EncodeToString(res)
+
+		assert.NoError(t, err)
+		assert.Equal(t, expectedHex128byte[i], resHex)
+	}
+}
+
+// Test vectors from: https://datatracker.ietf.org/doc/rfc9380
+func TestExpandMessageXMDSHA512(t *testing.T) {
```
AnomalRoil, line 134: Could you also implement the tests with SHAKE128 ? 
- https://datatracker.ietf.org/doc/html/rfc9380#name-expand_message_xofshake128
- https://datatracker.ietf.org/doc/html/rfc9380#name-expand_message_xofshake128-
-  and the long DST ones for SHA-256 ? https://datatracker.ietf.org/doc/html/rfc9380#name-expand_message_xmdsha-256-l

 It should be easy in Go: https://pkg.go.dev/golang.org/x/crypto/sha3#ShakeSum128

group/edwards25519/fe.go
```diff
@@ -245,6 +248,31 @@ func feToBytes(s *[32]byte, h *fieldElement) {
 	s[31] = byte(h[9] >> 18)
 }
 
+func feToBn(dst *big.Int, src *fieldElement) {
```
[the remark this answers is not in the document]
K1li4nL: Not exactly, it converts to big.Int, its fixed !

group/edwards25519/fe.go
```diff
@@ -245,6 +248,31 @@ func feToBytes(s *[32]byte, h *fieldElement) {
...
+	half := len(b) / 2
+	l := len(b) - 1
+	for i := 0; i < half; i++ {
+		b[i], b[l-i] = b[l-i], b[i]
+	}
+
+	dst.SetBytes(b[:])
+}
+
+func feFromBn(dst *fieldElement, src *big.Int) {
```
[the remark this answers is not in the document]
K1li4nL: Not exactly, it converts from big.Int, its fixed !

group/edwards25519/point.go
K1li4nL: Yes it can, though our Curve25519 doesn't use `fieldElement` but only `big.Int`, it would make sense to put it there but it would be of no use unfortunately.

group/edwards25519/point.go
K1li4nL: for `curve25519Elligator2` I used: https://datatracker.ietf.org/doc/html/rfc9380#name-curve25519-q-5-mod-8-k-1
That was put in the function doc for clarity.

group/edwards25519/point_test.go
K1li4nL: The current implementation doesn't suppor DST of length superior to 255 byte. Do you want to implement that support as well ?

K1li4nL: `expand_message_xof`[ is actually a different function](https://datatracker.ietf.org/doc/html/rfc9380#name-expand_message_xof),  I could implement it if you want ?

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	go.dedis.ch/protobuf v1.0.11
 	golang.org/x/crypto v0.21.0
 	golang.org/x/sys v0.18.0
+	golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7
```
AnomalRoil, line 14: Do we really need to add `xerrors`?

K1li4nL: not necessarily, I removed it.

group/edwards25519/point_test.go
K1li4nL: Added the xof version of expand message

group/edwards25519/fe_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.dedis.ch/kyber/v3/xof/blake2xb"
+)
+
+func Test_feToBnZero(t *testing.T) {
```
si-co, line 11: I'd add some additional edge cases (e.g. all 1's) and put those cases in a list, so that it's easy to add new ones (like here: https://github.com/FiloSottile/edwards25519/blob/main/field/fe_test.go#L47).

group/edwards25519/fe_test.go
```diff
@@ -0,0 +1,64 @@
...
+	var f0 fieldElement
+	feZero(&f0)
+
+	b0 := big.NewInt(0)
+	feToBn(b0, &f0)
+
+	assert.Equal(t, big.NewInt(0).Cmp(b0), 0)
+}
+
+func Test_feBnConversionRandom(t *testing.T) {
```
si-co, line 21: I agree that property testing is a good way to test the conversion, but maybe you can use other properties that involve multiplication (e.g., distributive) to make the test stronger.

group/edwards25519/point.go
```diff
@@ -290,3 +296,319 @@ func (P *point) IsCanonical(s []byte) bool {
...
+	P.Add(q0, q1)
+
+	// Clear cofactor
+	P.Mul(cofactorScalar, P)
+
+	return P
+}
+
+func hashToField(m []byte, dst string, count int) []fieldElement {
+	l := 48 // L param in RFC9380
```
si-co, line 313: I guess this comes from Section 5 in the RFC, right? I'd add a direct pointer to the correct (sub)section.

group/edwards25519/point.go
```diff
@@ -290,3 +296,319 @@ func (P *point) IsCanonical(s []byte) bool {
...
+		u[i] = fe
+	}
+
+	return u
+}
+
+func expandMessageXMD(h hash.Hash, m []byte, domainSeparator string, byteLen int) ([]byte, error) {
+	r := float64(byteLen) / float64(h.Size()>>3)
+	ell := int(math.Ceil(r))
+	if ell > 255 || ell < 0 || byteLen > 65535 {
```
si-co, line 333: The RFC specifies that the domain separator MUST have nonzero length: I'd add this to the invalid parameters test (cf. Section 3.1 in the RFC).

group/edwards25519/point.go
```diff
@@ -290,3 +296,319 @@ func (P *point) IsCanonical(s []byte) bool {
...
+	mPrime = append(mPrime, byteLenStr...)
+	mPrime = append(mPrime, zeroPad...)
+	mPrime = append(mPrime, dstPrime...)
+
+	// b0 = H(msg_prime)
+	h.Reset()
+	h.Write([]byte(mPrime))
+	b0 := h.Sum(nil)
+
+	// b_1 = H(b_0 || I2OSP(1, 1) || DST_prime
```
si-co, line 369: Missing `)` in comment.

group/edwards25519/point.go
```diff
@@ -290,3 +296,319 @@ func (P *point) IsCanonical(s []byte) bool {
...
+
+	return bFinal[:byteLen], nil
+}
+
+func expandMessageXOF(h sha3.ShakeHash, m []byte, domainSeparator string, byteLen int) ([]byte, error) {
+	if byteLen > 65535 {
+		return nil, errors.New("invalid parameters")
+	}
+
+	if len(domainSeparator) > 255 {
```
si-co, line 404: You have this twice, maybe create a function that handles long domain separators?

K1li4nL: The second one takes a different type of hash function, I don't think it make much sense to have a function for that, I did de-duplicated the repeated string though.

group/edwards25519/point.go
K1li4nL: Done !

group/edwards25519/point.go
K1li4nL: Done !

group/edwards25519/fe_test.go
K1li4nL: Done !

azixus (changes requested):
I just have one comment on `expandMessageXOF`, the rest looks good to me :)

group/edwards25519/point.go
```diff
@@ -290,3 +297,319 @@ func (P *point) IsCanonical(s []byte) bool {
...
+		return nil, err
+	}
+
+	if n != byteLen {
+		return nil, fmt.Errorf("read %d byte instead of expected %d from xof", n, byteLen)
+	}
+
+	h.Reset()
+	return uniformBytes, nil
+}
```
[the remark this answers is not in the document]
K1li4nL: Its fixed, thank you very much !

azixus (approved):
Looks all good to me now, thank you for the work!

PR #512, opened 2024-04-29
Final merge PR for drand/kyber

AnomalRoil:
This is a new PR for #509 to allow everyone to work against the `drandmerge` branch on the Dedis repo.


hash.go
```diff
@@ -1,8 +1,15 @@
...
+import (
+	"hash"
+)
 
 // A HashFactory is an interface that can be mixed in to local suite definitions.
 type HashFactory interface {
 	Hash() hash.Hash
 }
+
+// HashablePoint is an interface implemented by n
```
CluEleSsUK, line 12: implemented by `n`?

internal/test/scheme.go
```diff
@@ -0,0 +1,95 @@
...
+		sig, err := s.Sign(private, msg)
+		require.Nil(tt, err)
+		_, public := s.NewKeyPair(random.New())
+		if s.Verify(public, msg, sig) == nil {
+			tt.Fatal("verification succeeded unexpectedly")
+		}
+	})
+}
+
+// AggregationTesting test an aggrgatable scheme
```
CluEleSsUK, line 43:
```suggestion
// AggregationTesting tests an aggregatable scheme
```

pairing/bn254/curve.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (c *curvePoint) Neg(a *curvePoint) {
+	c.x.Set(&a.x)
+	gfpNeg(&c.y, &a.y)
+	c.z.Set(&a.z)
+	c.t = gfP{0}
+}
+
+// Clone makes a hard copy of the curve point
```
CluEleSsUK, line 249:
```suggestion
// Clone makes a deep copy of the curve point
```

pairing/bn254/gfp12.go
```diff
@@ -0,0 +1,169 @@
...
+	t1.MulTau(t1).Sub(t2, t1)
+	t2.Invert(t1)
+
+	e.x.Neg(&a.x)
+	e.y.Set(&a.y)
+	e.MulScalar(e, t2)
+	return e
+}
+
+// Clone makes a hard copy of the field
```
CluEleSsUK, line 162:
```suggestion
// Clone makes a deep copy of the field
```

pairing/bn254/gfp2.go
```diff
@@ -0,0 +1,165 @@
...
+
+func gfP2Decode(in *gfP2) *gfP2 {
+	out := &gfP2{}
+	montDecode(&out.x, &in.x)
+	montDecode(&out.y, &in.y)
+	return out
+}
+
+func (e *gfP2) String() string {
+	return "(" + e.x.String() + ", " + e.y.String() + ")"
```
CluEleSsUK, line 21:
```suggestion
	return fmt.Sprintf("(%s, %s)", e.x.String(), e.y.String())
```
may need to add an import too

pairing/bn254/gfp2.go
```diff
@@ -0,0 +1,165 @@
...
+	inv.Invert(t1)
+
+	gfpNeg(t1, &a.x)
+
+	gfpMul(&e.x, t1, inv)
+	gfpMul(&e.y, &a.y, inv)
+	return e
+}
+
+// Clone makes a hard copy of the field
```
CluEleSsUK, line 158:
```suggestion
// Clone makes a deep copy of the field
```

pairing/bn254/gfp6.go
```diff
@@ -0,0 +1,224 @@
...
+// http://eprint.iacr.org/2006/471.pdf.
+
+// gfP6 implements the field of size p⁶ as a cubic extension of gfP2 where τ³=ξ
+// and ξ=i+9.
+type gfP6 struct {
+	x, y, z gfP2 // value is xτ² + yτ + z
+}
+
+func (e *gfP6) String() string {
+	return "(" + e.x.String() + ", " + e.y.String() + ", " + e.z.String() + ")"
```
CluEleSsUK, line 14: `fmt.Sprintf` would be nice for consistency

pairing/bn254/gfp6.go
```diff
@@ -0,0 +1,224 @@
...
+
+	F.Invert(F)
+
+	e.x.Mul(C, F)
+	e.y.Mul(B, F)
+	e.z.Mul(A, F)
+	return e
+}
+
+// Clone makes a hard copy of the field
```
CluEleSsUK, line 215:
```suggestion
// Clone makes a deep copy of the field
```

pairing/bn254/optate.go
```diff
@@ -0,0 +1,270 @@
...
+
+	aAffine := &twistPoint{}
+	aAffine.Set(q)
+	aAffine.MakeAffine()
+
+	bAffine := &curvePoint{}
+	bAffine.Set(p)
+	bAffine.MakeAffine()
+
+	minusA := &twistPoint{}
```
CluEleSsUK, line 133: might be able to avoid an allocation sometimes if you move this inside the switch

pairing/bn254/point.go
```diff
@@ -0,0 +1,723 @@
...
+	return p
+}
+
+func (p *pointGT) Set(q kyber.Point) kyber.Point {
+	x := q.(*pointGT).g
+	p.g.Set(x)
+	return p
+}
+
+// Clone makes a hard copy of the point
```
CluEleSsUK, line 552:
```suggestion
// Clone makes a deep copy of the point
```

pairing/bn254/point.go
```diff
@@ -0,0 +1,723 @@
...
+// Borrowed from: https://github.com/kilic/bls12-381/blob/master/hash_to_field.go
+func expandMsgXmdKeccak256(domain, msg []byte, outLen int) []byte {
+	h := sha3.NewLegacyKeccak256()
+	domainLen := uint8(len(domain))
+	if domainLen > 255 {
+		panic("invalid domain length")
+	}
+	// DST_prime = DST || I2OSP(len(DST), 1)
+	// b_0 = H(Z_pad || msg || l_i_b_str || I2OSP(0, 1) || DST_prime)
+	_, _ = h.Write(make([]byte, h.BlockSize()))
```
CluEleSsUK, line 288: in the `gnarkExpandMsgXmd` we check/return errors but here we don't. Is there a reason to swallow them here? Admittedly they're unlikely to happen, but it seems like good hygiene

pairing/bn254/twist.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (c *twistPoint) Neg(a *twistPoint) {
+	c.x.Set(&a.x)
+	c.y.Neg(&a.y)
+	c.z.Set(&a.z)
+	c.t.SetZero()
+}
+
+// Clone makes a hard copy of the point
```
CluEleSsUK, line 207:
```suggestion
// Clone makes a deep copy of the point
```

pairing/circl_bls12381/adapter.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"go.dedis.ch/kyber/v3"
+)
+
+// SuiteBLS12381 is an adapter that implements the suites.Suite interface so that
+// bls12381 can be used as a common suite to generate key pairs for instance but
+// still preserves the properties of the pairing (e.g. the Pair function).
+//
+// It's important to note that the Point function will generate a point
+// compatible with public keys only (group G2) where the signature must be
```
CluEleSsUK, line 12: is this not the reverse of the standard case? It seems that most bls libs put pub keys on G1 and sigs on G2. In drand we do the opposite.
It seems circl supports both

pairing/bn254/point.go
K1li4nL: The `HashablePoint.Hash` doesn't have an error in its return, I tried to handle the errors but the cascading effect goes out of hand. This pr might not be the place to do that ?

AnomalRoil:
It seems this was mistakenly merged instead of rebased on top of master. I recommend reverting it and doing a rebase lest future git bisect will basically end up here more often than not

PR #515, opened 2024-05-08
Fix dkg bug

K1li4nL:


share/dkg/dkg_test.go
```diff
@@ -294,7 +305,7 @@ func TestDKGSkipIndex(t *testing.T) {
 	// remove a random node from the previous group
 	offlineToRemove := uint32(rand.Intn(n))
 	for _, node := range tns {
-		if node.Index == offlineToRemove {
+		if node.Index == uint32(oldIndexes[offlineToRemove]) {
 			continue
 		}
```
AnomalRoil, line 310: Why not instead:
```
	offlineToRemove := uint32(rand.Intn(n))
	for i, node := range tns {
		if i == offlineToRemove {
			continue
		}
```
???

K1li4nL: no reason really, love the 2 line fix, its done !

AnomalRoil (approved):

share/dkg/dkg_test.go
ineiti: @K1li4nL sorry to chime in late - but are the `node.Index` always starting from 0, and increasing one by one?

K1li4nL: Hi, no problem, that's how it is done in the tests at least, I don't think such requirement exist beyond that, not completely familiar with this protocol. @AnomalRoil solution (which was adopted) make this question a non-issue since it relies only on the indexes of the array.

ineiti: OK, I see. It's in the test, and there the index of the node equals `node.Index`. I think this is not given in a general run, but here it should hold.

All good then :)

K1li4nL: The test `TestDKGSkipIndex` also makes it slightly general by making a node with index 53 instead 1, the source of the bug

PR #517, opened 2024-05-14
Uncomment bls_test and internal folder

Robingoumaz:
Uncommenting sign/bls/bls_test.go and fixed issues arising. 
Also moved test tools into an internal folder.


AnomalRoil (approved):
Why is CI failing here?

PR #519, opened 2024-05-16
Change import names from v3 to v4

Robingoumaz:
Just a PR to change import and mod names from kyber/v3 to kyber/v4

### DIFF
--- a/encrypt/ibe/ibe.go
+++ b/encrypt/ibe/ibe.go
@@ -9,10 +9,10 @@ import (
 	"errors"
 	"fmt"
 
-	"go.dedis.ch/kyber/v3"
-	"go.dedis.ch/kyber/v3/group/mod"
-	"go.dedis.ch/kyber/v3/pairing"
-	"go.dedis.ch/kyber/v3/util/random"
+	"go.dedis.ch/kyber/v4"
+	"go.dedis.ch/kyber/v4/group/mod"
+	"go.dedis.ch/kyber/v4/pairing"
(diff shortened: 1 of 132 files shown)


encrypt/ibe/ibe.go
```diff
@@ -9,10 +9,10 @@ import (
 	"errors"
 	"fmt"
 
-	"go.dedis.ch/kyber/v3"
-	"go.dedis.ch/kyber/v3/group/mod"
-	"go.dedis.ch/kyber/v3/pairing"
-	"go.dedis.ch/kyber/v3/util/random"
+	"go.dedis.ch/kyber/v4"
```
CluEleSsUK, line 12: I'm guessing there's a reason this is all commented out, but it seems suspicious - could just nuke it

go.mod
```diff
@@ -1,4 +1,4 @@
-module go.dedis.ch/kyber/v3
+module go.dedis.ch/kyber/v4
 
 go 1.18
```
CluEleSsUK, line 3: have we considered bumping the go version to something newer? does it break things?

CluEleSsUK (approved):

go.mod
AnomalRoil: we actually have upgraded it from Go 1.13 to Go 1.18 already! 
In truth it would probably be nice to update to Go 1.22 to get the new Loop behaviour, but at the same time a library like Kyber kinda wants to support older codebases too I guess.

AnomalRoil (approved):

PR #520, opened 2024-05-21
Add challenge check PVSS

K1li4nL:


share/pvss/pvss.go
```diff
@@ -35,52 +35,65 @@ var errorTooFewShares = errors.New("not enough shares to recover secret")
 var errorDifferentLengths = errors.New("inputs of different lengths")
 var errorEncVerification = errors.New("verification of encrypted share failed")
 var errorDecVerification = errors.New("verification of decrypted share failed")
+var errorChallengeVerification = errors.New("verification of the challenge failed")
```
CluEleSsUK, line 38: I'd suggest exposing these errors and prefixing them with just `Err` as per the [go naming conventions](https://littleroot.org/blog/go-naming-errors/), e.g. `ErrChallengeVerification`.
That will allow users to react to errors in a meaningful way without doing string matching

share/pvss/pvss.go
```diff
@@ -35,52 +35,65 @@ var errorTooFewShares = errors.New("not enough shares to recover secret")
...
+
+type PubRun struct {
```
CluEleSsUK, line 40: `PubRun` sounds like what I do on a friday night. Would suggest a more descriptive name or a comment explaining it

share/pvss/pvss.go
```diff
@@ -159,13 +172,36 @@ func VerifyDecShare(suite Suite, G kyber.Point, X kyber.Point, encShare *PubVerS
...
+	for i, el := range pubRun.Shares {
+		indToShare[el.S.I] = el
+		indToPk[el.S.I] = pubRun.X[i]
+	}
+
+	filteredDecShares := make([]*PubVerShare, 0)
+	filteredEncShares := make([]*PubVerShare, 0)
+	filteredPks := make([]kyber.Point, 0)
+	for _, decShare := range decShares {
+		if encShare, ok := indToShare[decShare.S.I]; ok {
```
CluEleSsUK, line 188: is there a use case for returning valid decrypted shares but silently ignoring that some are invalid?

K1li4nL: If enough are valid we can still retrieve the secret, otherwise the user would need to do the filtering themselves, we also don't have a way to log the event

share/pvss/pvss.go
K1li4nL: I went for `PubVSSData` and added a comment, hopefully better now

share/pvss/pvss.go
```diff
@@ -152,39 +166,94 @@ func DecShareBatch(suite Suite, H kyber.Point, X []kyber.Point, sH []kyber.Point
...
+
 	var D []*PubVerShare // good decrypted shares
-	for i := 0; i < len(X); i++ {
-		if err := VerifyDecShare(suite, G, X[i], encShares[i], decShares[i]); err == nil {
-			D = append(D, decShares[i])
+	for i := 0; i < len(filteredPks); i++ {
+		if err := VerifyDecShare(suite, G, filteredPks[i], filteredEncShares[i], filteredDecShares[i]); err != nil {
+			continue
+		}
+		if !pubRun.Shares[i].P.C.Equal(cExp) {
```
AnomalRoil, line 205: I find this very surprising that all shares are sharing the same challenge, and also shouldn't you rather be checking the response ?

Each of them should have a different challenge, and you should instead be using `dleq.Verify` to verify the proof, which requires knowing `G`, `H`,  `xG` and `xH`, as well as the proof.

See here:
https://asecuritysite.com/encryption/go_dleq

Please can you add a test to `dleq_test.go` using the values on this page to check it does work as expected?

AnomalRoil (changes requested):

share/pvss/pvss.go
```diff
@@ -152,39 +166,94 @@ func DecShareBatch(suite Suite, H kyber.Point, X []kyber.Point, sH []kyber.Point
...
+	}
+
 	var D []*PubVerShare // good decrypted shares
-	for i := 0; i < len(X); i++ {
-		if err := VerifyDecShare(suite, G, X[i], encShares[i], decShares[i]); err == nil {
-			D = append(D, decShares[i])
+	for i := 0; i < len(filteredPks); i++ {
+		if err := VerifyDecShare(suite, G, filteredPks[i], filteredEncShares[i], filteredDecShares[i]); err != nil {
+			continue
+		}
```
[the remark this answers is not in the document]
K1li4nL: In the function documentation: _The function returns the the valid decrypted shares._

This behavior makes sense to me, as a threshold based protocol, it is expected that some parts may not be available / corrupt, if there are enough valid share I am not sure why we should stop ? 

I can add a comment yes

share/pvss/pvss.go
```diff
@@ -152,39 +166,94 @@ func DecShareBatch(suite Suite, H kyber.Point, X []kyber.Point, sH []kyber.Point
...
+	}
+	for _, x := range vG {
+		x.MarshalTo(h)
+	}
+	for _, x := range vH {
+		x.MarshalTo(h)
+	}
+	cb := h.Sum(nil)
+
+	cExp := suite.Scalar().Pick(suite.XOF(cb))
```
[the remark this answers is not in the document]
K1li4nL: Sorry for the _terrible_ variable name, I will replace it by _challengeExpected_ for clarity, the point is to compare the challenge included in the proof against the one we compute using the public values. The challenge is just recomputed as was done in `dleq.go:NewDLEQProof()`

share/pvss/pvss.go
```diff
@@ -31,64 +31,78 @@ type Suite interface {
...
 
-	return encShares, pubPoly, nil
+	return p, nil
 }
 
 // VerifyEncShare checks that the encrypted share sX satisfies
 // log_{H}(sH) == log_{X}(sX) where sH is the public commitment computed by
 // evaluating the public commitment polynomial at the encrypted share's index i.
 func VerifyEncShare(suite Suite, H kyber.Point, X kyber.Point, sH kyber.Point, encShare *PubVerShare) error {
 	if err := encShare.P.Verify(suite, H, X, sH, encShare.S.V); err != nil {
```
azixus, line 104: We should include the verification of the DLEQ proof challenge `c` here. It's perhaps better to do that verification into `VerifyEncShareBatch` directly, since the "encrypted shares" have a common challenge.

share/pvss/pvss.go
```diff
@@ -121,19 +135,19 @@ func DecShare(suite Suite, H kyber.Point, X kyber.Point, sH kyber.Point, x kyber
 	G := suite.Point().Base()
 	V := suite.Point().Mul(suite.Scalar().Inv(x), encShare.S.V) // decryption: x^{-1} * (xS)
 	ps := &share.PubShare{I: encShare.S.I, V: V}
-	P, _, _, err := dleq.NewDLEQProof(suite, G, V, x)
+	P, com, _, err := dleq.NewDLEQProof(suite, G, V, x)
```
azixus, line 138: Same remark as `EncShares` here - we don't need to output `com`. To verify the proof of a `decShare`, we need (G, X[i], V, encShares[i]) <-> $(G, y_i, S_i, Y_i)$. Three parameters are already public, $(G, y_i, Y_i)$, and $S_i$ is output by `DecShare` alongside the proof.

share/pvss/pvss.go
```diff
@@ -152,39 +166,95 @@ func DecShareBatch(suite Suite, H kyber.Point, X []kyber.Point, sH []kyber.Point
...
 	var D []*PubVerShare // good decrypted shares
-	for i := 0; i < len(X); i++ {
-		if err := VerifyDecShare(suite, G, X[i], encShares[i], decShares[i]); err == nil {
-			D = append(D, decShares[i])
+	for i := 0; i < len(filteredPks); i++ {
+		if err := VerifyDecShare(suite, G, filteredPks[i], filteredEncShares[i], filteredDecShares[i]); err != nil {
+			// Expected behavior is to accept some shares may be invalid and keep going
+			continue
+		}
+		if !pubRun.Shares[i].P.C.Equal(expectedChallenge) {
```
azixus, line 206: The wrong challenge is verified here. It seems that you are verifying the common challenge for the so-called "encrypted shares" here. This should have already been verified in `VerifyEncShare` / `VerifyEncShareBatch`.

Nonetheless, we also need to verify each of the $n$ unique challenges for the `decShares` within `VerifyDecShare`.

share/pvss/pvss.go
```diff
@@ -152,39 +166,95 @@ func DecShareBatch(suite Suite, H kyber.Point, X []kyber.Point, sH []kyber.Point
...
+	}
+	for _, x := range vG {
+		x.MarshalTo(h)
+	}
+	for _, x := range vH {
+		x.MarshalTo(h)
+	}
+	cb := h.Sum(nil)
+
+	expectedChallenge := suite.Scalar().Pick(suite.XOF(cb))
```
azixus, line 245: As mentioned in previous comments, the challenge derivation should be moved to `VerifyEncShareBatch`.

azixus (changes requested):

share/pvss/pvss.go
```diff
@@ -31,64 +31,78 @@ type Suite interface {
...
 
-	for i := 0; i < n; i++ {
+	for i := 0; i < p.N; i++ {
 		ps := &share.PubShare{I: indices[i], V: sX[i]}
-		encShares[i] = &PubVerShare{*ps, *proofs[i]}
+		p.Shares[i] = &PubVerShare{*ps, coms[i], *proofs[i]}
 	}
 
-	return encShares, pubPoly, nil
+	return p, nil
```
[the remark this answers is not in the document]
K1li4nL: Hello and thank you very much for your review ! Given your comments, I started from zero, indeed no need to change the output there.  I added a function to compute the coms[i] instead.

share/pvss/pvss.go
K1li4nL: I moved the common challenge check in `VerifyEncShareBatch`

azixus: Great point! I agree, it's probably better to add the param to `VerifyEncShare`.

share/pvss/pvss.go
K1li4nL: Fixed !

share/pvss/pvss.go
K1li4nL: I added the verification of the decrypted shares challenge

azixus (approved):
I just left one comment on `computeCommitments`. The function could be slightly improved by using Horner's method to compute the commitments. Other than that, it looks good now.

share/pvss/pvss.go
```diff
@@ -83,27 +87,108 @@ func EncShares(suite Suite, H kyber.Point, X []kyber.Point, secret kyber.Scalar,
...
+			expScalar.SetInt64(exp.Int64())
+			com := suite.Point().Mul(expScalar, cj) // C_j ** (i**j)
+			acc.Add(acc, com)
+		}
+
+		coms[ith] = acc
+	}
+
+	return coms, nil
+}
```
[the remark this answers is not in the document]
K1li4nL: This is much more elegant, thank you, I implemented it.

share/pvss/pvss.go
```diff
@@ -151,8 +229,21 @@ func DecShareBatch(suite Suite, H kyber.Point, X []kyber.Point, sH []kyber.Point
...
+	cb := h.Sum(nil)
+	expDecChallenge := suite.Scalar().Pick(suite.XOF(cb))
+
+	if !decShare.P.C.Equal(expDecChallenge) {
+		return ErrDecShareChallengeVerification
+	}
+
 	if err := decShare.P.Verify(suite, G, decShare.S.V, X, encShare.S.V); err != nil {
-		return errorDecVerification
+		return ErrDecVerification
```
AnomalRoil, line 246:
```suggestion
		return fmt.Errorf("didn't verify: %w", ErrDecVerification)
```

share/pvss/pvss.go
```diff
@@ -161,7 +252,7 @@ func VerifyDecShare(suite Suite, G kyber.Point, X kyber.Point, encShare *PubVerS
 // slices of decrypted shares. The function returns the the valid decrypted shares.
 func VerifyDecShareBatch(suite Suite, G kyber.Point, X []kyber.Point, encShares []*PubVerShare, decShares []*PubVerShare) ([]*PubVerShare, error) {
 	if len(X) != len(encShares) || len(encShares) != len(decShares) {
-		return nil, errorDifferentLengths
+		return nil, ErrDifferentLengths
```
AnomalRoil, line 255:
```suggestion
		return nil, fmt.Errorf("didn't verify: %w", ErrDifferentLengths)
```

AnomalRoil (changes requested):

AnomalRoil (approved):

PR #521, opened 2024-05-22
Benchmarking platform

matteosz:
This PR implements a benchmarking visualization tool for kyber and the script to run the benchmarks itself. 

Steps to ensure it works correctly:
- GITHUB_TOKEN must have read and write permissions enabled
- GITHUB pages must be configured to deploy from gh-pages branch and use ./root folder

home

comparison

ranking

sign


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
CluEleSsUK, line 19: I'd suggest upping to 22 unless there's a compatibility reason not to

docs/benchmark-app/public/index.html
```diff
@@ -0,0 +1,43 @@
...
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>React App</title>
```
CluEleSsUK, line 27:
```suggestion
    <title>Kyber benchmarks</title>
```

docs/benchmark-app/public/logo192.png
CluEleSsUK, line 1: a cool dedis logo favicon would be nice ;p

docs/benchmark-app/src/App.test.js
CluEleSsUK, line 1: can nuke this

docs/benchmark-app/src/data/data.json
CluEleSsUK, line 1: should this be empty?

docs/benchmark-app/src/setupTests.js
CluEleSsUK, line 1: can nuke this as there are no tests

docs/benchmark-app/src/data/data.json
matteosz: Oh now I executed the script and then ctrl+c, and then forgot to restart 😅

AnomalRoil:
Do you think we should do another repo "kyber-benchmark" with the web bits to get a dedicated Github Page and not clutter the kyber repo too much?

Leaving all the Go bits here, but moving the HTML, static resources, etc. to their own repo and pulling the data we need from this repo in CI.

matteosz:
> Do you think we should do another repo "kyber-benchmark" with the web bits to get a dedicated Github Page and not clutter the kyber repo too much?
> 
> Leaving all the Go bits here, but moving the HTML, static resources, etc. to their own repo and pulling the data we need from this repo in CI.

I don't know, in kyber there's already the docs folder which hosts the webpage with documentation and other stuff, and so I see the benchmarks as an extension to that; also everything centralized is easier to manage and update in the future, I feel that a separate repo might be forgotten and leaking behind through the development of the repo. Plus, it's very isolated from the library and only 1.4 MB so I guess my preference would be to keep it in Kyber, but I could move it if someone in DEDIS created a custom repo for it.

PR #524, opened 2024-05-23
Nonce removal from vss/EncryptedDeal struct

### LINKED ISSUE #379: GCM Nonce field can be removed from rabin.vss/EncryptedDeal
[...since its value is always 0.](https://github.com/dedis/kyber/blob/89ca87/share/vss/rabin/vss.go#L224) 

```golang
	nonce := make([]byte, gcm.NonceSize())
	dealBuff, err := protobuf.Encode(d.deals[i])
	if err != nil {
		return nil, err
	}
	encrypted := gcm.Seal(nil, nonce, dealBuff, d.hkdfContext)
```

It's probably OK to leave the nonce constant because the key is ephemeral, but if so the field is just deadweight on the protocol's bandwidth.

matteosz:
This PR resolves #379 by removing the nonce from the Encrypted Deal struct (as it's always 0)

### DIFF
--- a/share/vss/pedersen/vss.go
+++ b/share/vss/pedersen/vss.go
@@ -69,8 +69,6 @@ type EncryptedDeal struct {
 	DHKey []byte
 	// Signature of the DH key by the longterm key of the dealer
 	Signature []byte
-	// Nonce used for the encryption
-	Nonce []byte
 	// AEAD encryption of the deal marshalled by protobuf
 	Cipher []byte
 }
@@ -206,7 +204,6 @@ func (d *Dealer) EncryptedDeal(i int) (*EncryptedDeal, error) {
 	return &EncryptedDeal{
 		DHKey:     dhBytes,
 		Signature: signature,
-		Nonce:     nonce,
 		Cipher:    encrypted,
 	}, nil
 }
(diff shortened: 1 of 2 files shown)


share/vss/pedersen/vss.go
```diff
@@ -408,7 +405,8 @@ func (v *Verifier) decryptDeal(e *EncryptedDeal) (*Deal, error) {
 	if err != nil {
 		return nil, err
 	}
-	decrypted, err := gcm.Open(nil, e.Nonce, e.Cipher, v.hkdfContext)
+	nonce := make([]byte, gcm.NonceSize())
```
CluEleSsUK, line 408: is it intentional to pass in an empty slice here? it seems like it'd be better to remove it entirely (though this breaks the interface for other users potentially)

matteosz: I'm not sure I understood well your point but I'll try to answer anyway. 

It's not possible to not pass the nonce, as it's required from the gcm interfact; additionally, if we were to pass nil instead that wouldn't work either, as it has to match the NonceSize(). Also, this is a v4 change, so it's okay to break our interfaces (the encrypted deal in this case).

AnomalRoil (approved):

PR #525, opened 2024-05-23
Update to go 1.21 + set go version in CI

### LINKED ISSUE #522: Use specific go versions in workflows and go.mo
The current workflows all have the following:

```
      - name: Set up Go ^1.20
        uses: actions/setup-go@v3
        with:
          go-version: ^1.20
```

Or version 1.13. But this will always install the latest version of go, as of May 2024 this is 1.22.

I propose to choose an explicit go version, even though it needs to be updated from time to time. And then sync this with the version in `go.mod`, which currently is at 1.13...

For the cothority, we tested the two latest go versions with a matrix, which would be 1.21 and 1.22 here. And put the lower version in go.mod, so it's a little bit more backwards-compatible.

K1li4nL:
Closes #522


ineiti:
Now you see that you have to update a lot of go-versions in the `test_on_pr` :) This is one of the reasons I prefer the matrix thingie...

K1li4nL:
> Now you see that you have to update a lot of go-versions in the `test_on_pr` :) This is one of the reasons I prefer the matrix thingie...

You want to test for both go 1.21 and 1.22 ? I added a variable for each workflow, this reduces the amount of changes needed, but it doesn't seem to be what you expect

ineiti:
> You want to test for both go 1.21 and 1.22 ? I added a variable for each workflow, this reduces the amount of changes needed, but it doesn't seem to be what you expect

Adding a variable is good, I think using a matrix is even better - see my comment in https://github.com/dedis/kyber/pull/502#discussion_r1611863282 . I added another 'dimension' to the matrix in https://github.com/ineiti/test_ubuntu_32bits/blob/main/.github/workflows/test.yaml and now use `exclude` to not test the windows-32bit entries. If you look a the [action](https://github.com/ineiti/test_ubuntu_32bits/actions/runs/9219724969) you can see that it iterates through windows only for 64-bit, and for the ubuntu it does both 64-bit and 32-bit. Plus both golang versions.

You still only have to write the golang-versions once, and all the rest in `test_on_push.yml` is much shorter.

AnomalRoil:
Why Go 1.21 and not Go 1.22 or Go 1.18 ?

K1li4nL:
done in #526

PR #526, opened 2024-05-24
CI tests for 32bit correction + version of Go update

### LINKED ISSUE #522: Use specific go versions in workflows and go.mo

matteosz:
It addresses [@ineiti's comments](https://github.com/dedis/kyber/pull/502#discussion_r1611863282) of using a matrix to simplify the ci.

Resolves #522


.github/workflows/test_on_pr.yml
```diff
@@ -29,67 +29,81 @@ jobs:
...
     env:
       DBGSYNCLOG: trace
       DBGSYNCON: true
+
     steps:
-    - name: Set up Go ^1.13
-      uses: actions/setup-go@v3
+    - name: Matrix is ${{matrix.os}} / ${{matrix.size}} - Go is ${{matrix.golang}}
+      run: date
+
```
ineiti, line 53: You can remove those - this was for debugging :)

.github/workflows/test_on_pr.yml
```diff
@@ -29,67 +29,81 @@ jobs:
...
+        go version
+      shell: alpine.sh {0}
 
-    - name: Test without coverage
-      run: make test
+    - name: getconf x64
+      if: ${{ matrix.size == '64b' }}
+      run: |
+        getconf LONG_BIT
+        go version
```
ineiti, line 90: Remove debugging for the final PR.

.github/workflows/test_on_pr.yml
```diff
@@ -29,67 +29,81 @@ jobs:
...
+    - name: Test without coverage (Windows, MacOS)
+      if: ${{matrix.os == 'macos-latest' || matrix.os == 'windows-latest'}}
+      run: make test
 
-    - name: Check out code into the Go module directory
-      uses: actions/checkout@v3
-      with:
-        ref: ${{ github.event.pull_request.head.sha }}
+    - name: Test without coverage (Ubuntu x86)
+      if: ${{matrix.os == 'ubuntu-latest' && matrix.size == '32b'}}
```
ineiti, line 97: Proposition:

```suggestion
      if: ${{matrix.size == '32b'}}
```

It's shorter, but misses some of the context. So I'm not sure it's a good proposition...

.github/workflows/test_on_push.yml
```diff
@@ -5,70 +5,80 @@ on:
...
     env:
       DBGSYNCLOG: trace
       DBGSYNCON: true
+
     steps:
-    - name: Set up Go ^1.13
-      uses: actions/setup-go@v3
+    - name: Matrix is ${{matrix.os}} / ${{matrix.size}} - Go is ${{matrix.golang}}
+      run: date
+
```
ineiti, line 28: Remove debugging output

.github/workflows/test_on_push.yml
```diff
@@ -5,70 +5,80 @@ on:
...
+        getconf LONG_BIT
+        go version
+      shell: alpine.sh {0}
+
+    - name: getconf x64
+      if: ${{ matrix.size == '64b' }}
+      run: |
+        getconf LONG_BIT
+        go version
+
```
ineiti, line 66: Remove debugging output

.github/workflows/test_on_push.yml
```diff
@@ -5,70 +5,80 @@ on:
...
-      with:
-        go-version: ^1.13
-
-    - name: Check out code into the Go module directory
-      uses: actions/checkout@v3
-      with:
-        ref: ${{ github.event.pull_request.head.sha }}
-        fetch-depth: 0
+    - name: Test without coverage (Ubuntu x86)
+      if: ${{matrix.os == 'ubuntu-latest' && matrix.size == '32b'}}
```
ineiti, line 72: Proposition: like above - can be made shorter, but might be confusing.

ineiti (changes requested):
Some small changes. And here is how you can have different golang versions in Alpine:

https://github.com/ineiti/test_ubuntu_32bits/blob/main/.github/workflows/test.yaml

TLDR: download the go-binary from the official website and add it to the PATH.

ineiti:
What is your reasoning behind testing go 1.18 and 1.22?

In some past projects we had a policy of "the last two stable versions". You _might_ fail on go 1.19, 1.20, or 1.21 with your policy.

.github/workflows/test_on_pr.yml
```diff
@@ -64,46 +61,39 @@ jobs:
...
-      if: ${{matrix.os == 'macos-latest' || matrix.os == 'windows-latest'}}
+      if: ${{ matrix.os == 'macos-latest' || matrix.os == 'windows-latest' }}
       run: make test
 
     - name: Test without coverage (Ubuntu x86)
-      if: ${{matrix.os == 'ubuntu-latest' && matrix.size == '32b'}}
-      run: make test
+      if: ${{ matrix.size == '32b' }}
+      run: |
+        GOVERSION=${{ matrix.golang == '1.21' && '1.21.10' || '1.22.3' }}
```
ineiti, line 82: Please put 1.21.10 and 1.22.3 in the matrix, so you don't have to test here. This will most probably be overlooked the day somebody will adjust the versions in the matrix above, and then things will start to fail.

ineiti (changes requested):
- The matrix in `test_on_pr` still has go 1.18 in it
- why is there a `test_on_pr` and a `test_on_push` with similar tests in it? This leads to errors, like you can see yourself: one has the correct golang-version matrix, the other doesn't. One still has the debug output in the `Setup Alpine Linux`, the other doesn't. 
 I propose to remove the `test_on_push` - if I read correctly, only the `Send coverage` is unique to it. Either include it in the `test_on_pr`, or add an `if` to it.

.github/workflows/tests.yml
```diff
@@ -1,27 +1,120 @@
...
-concurrency:
-  group: ci-${{ github.ref }}-test
-  cancel-in-progress: true
-
 jobs:
+  permission:
+    if: ${{ github.event_name == 'pull_request_target' }}
+    strategy:
+      matrix:
+        platform: [macos-latest, windows-latest, ubuntu-latest]
```
ineiti, line 14: I think a matrix is overkill here, and a simple `ubuntu-latest` should be enough, no?

ineiti (approved):
Perfect - I think you could remove the matrix from the `permission` job for a more-than-perfect CI/CD :)

.github/workflows/tests.yml
```diff
@@ -1,27 +1,117 @@
...
+    types: [opened, synchronize, reopened, labeled]
   push:
-    branches:
-      - master
-  pull_request:
     branches: [ master ]
 
-concurrency:
-  group: ci-${{ github.ref }}-test
-  cancel-in-progress: true
```
AnomalRoil, line 12: Why remove this early cancellation of already running tests when a new commit is pushed in quick succession on a PR?

AnomalRoil (approved):

.github/workflows/tests.yml
matteosz: Nice spot, it was an unwanted removal indeed

ineiti: We're entering Nirvana territory :)

Issue #384, opened 2019-05-08
Curve-renaming


ineiti:
After discussion with @Daeinar , there seems to be a need to clean up the `group/`-directory (again). As this is a backward-incompatible change, I don't know how and when this change should go in. But we should discuss it...

The question came up with ByzGen about our use of Curve25519. This curve can be represented in two different ways:

- Twisted Edward Curve: https://tools.ietf.org/html/rfc8032
- Montgomery Curve: https://tools.ietf.org/html/rfc7748

The two representations are isogenic, which means that you can get from one representation to the other with a change of coordinates.

The two (three) curve25519 implementation in kyber are as follows:
- `Ed25519` in `group/ed25519` is either constant-time, or variable-time, Twisted Edwards Curve implementation
- `Curve25519` in `group/curve25519` is a variable-time, Twisted Edwards Curve implementation

As of 2019, it seems that a consensus is materializing, where `Edwards25519` refers to the Twisted Edwards representation, while `Curve25519` refers to the Montgomery representation.

Additionally, the `nist` package holds only the `p256` curve.

This is why we propose to:
- Keep `Ed25519` as is
- Change the name of `Curve25519`, or remove it, and keep only the additional curves in there
- Change the `nist` package to `p256`
- Add a Montgomery representation of `Curve25519` to the repo and call this one `Curve25519`

pierluca:
Following #528, it's my understanding that the renaming has happened and the only missing element is adding the Montgomery representation of `Curve25519`. I'll create a separate issue mentioning this one.

PR #528, opened 2024-06-03
Curve renaming

matteosz:
This PR addresses partially #384, by:
- Keeping Ed25519 as is
- Changing name to variable Ed25519 (previously named curve25519)
- Changing name of nist to p256

For a future PR we could add the curve25519 (Montgomery representation).


group/var_ed25519/param.go
```diff
@@ -10,7 +10,7 @@
 // are isomorphic to curves having c == 1.
 //
 // For details see Bernstein et al, "Twisted Edwards Curves", http://eprint.iacr.org/2008/013.pdf
-package curve25519
+package var_ed25519
```
AnomalRoil, line 13: Need to change package comment on line 1 here.

group/p256/doc.go
```diff
@@ -1,3 +1,3 @@
 // Package nist implements cryptographic groups and ciphersuites
```
AnomalRoil, line 1: need to change package level comment here

group/var_ed25519/param.go
AnomalRoil: Also renaming below "`func Param25519() *Param`" to "`func ParamEd25519() *Param`" might be best.

group/var_ed25519/suite.go
AnomalRoil, line 1: I think all 25519 function in this file should be renamed "Ed25519", no ?

AnomalRoil (commented):
Overall seeing there are no test vectors for Ed25519 using actual given values and checking computations are correct makes me worried. 
But that's for another issue/PR I guess.

group/edwards25519/point.go
```diff
@@ -1,5 +1,5 @@
 // Package edwards25519 provides an optimized Go implementation of a
-// Twisted Edwards curve that is isomorphic to Curve25519. For details see:
+// Twisted Edwards curve that is isomorphic to var_ed25519. For details see:
```
AnomalRoil, line 2: This seems like an unwanted replacement

group/var_ed25519/curve.go
```diff
@@ -67,7 +67,7 @@ func (c *curve) PointLen() int {
 	return (c.P.BitLen() + 7 + 1) / 8
 }
 
-// NewKey returns a formatted curve25519 key (avoiding subgroup attack by requiring
+// NewKey returns a formatted var_ed25519 key (avoiding subgroup attack by requiring
 // it to be a multiple of 8). NewKey implements the kyber/util/key.Generator interface.
 func (c *curve) NewKey(stream cipher.Stream) kyber.Scalar {
```
AnomalRoil, line 72: This seems like an undesirable change. 
Also @ineiti @pierluca this function seems to be applying the clamping from Ed25519, which probably is not the right thing to do for all the other params provided in the param.go file... It might introduce bias or big issues.

AnomalRoil (changes requested):

group/var_ed25519/curve.go
AnomalRoil: maybe `returns a scalar suitable for use as a secret key` instead of naming the curve @matteosz?

AnomalRoil (approved):

PR #529, opened 2024-06-09
Simplified 32-bit fail fix

matteosz:
This PR is a simplification for #502 . The error was due to the protobuf decoding of an int field in the PriShare struct, only changing that field should solve the issue and minimise the risk since the changes are now minimal.


AnomalRoil (approved):
I'm not super happy with it I guess because now some indices are uint32 and some are int, so consistency isn't great.

I guess changing the protobuf package will have to happen before v5 xD

PR #540, opened 2024-08-16
Remove BLS signatures aggregate

K1li4nL:
As the BLS signature aggregate scheme is vulnerable to rogue public-key attack, this pr removes the "aggregate" part of the code and use BDN where BLS aggregates were used.


pairing/bls12381/bls12381_test.go
```diff
@@ -730,31 +733,47 @@ func BLSBenchmark(b *testing.B, curveOption string) {
...
+	}
 	for _, n := range numSigs {
+
 		// Benchmark aggregation of public keys
 		b.Run(fmt.Sprintf("AggregatePublicKeys-G1 on %d signs", n), func(bb *testing.B) {
 			for j := 0; j < bb.N; j++ {
-				result = schemeOnG1.AggregatePublicKeys(pubKeysOnG1[:n]...)
+				result, err = schemeOnG1.AggregatePublicKeys(maskG1)
+				if err != nil {
+					panic(err)
```
pierluca, line 752: why panic instead of a `require.NoError`?

pierluca (commented):
Overalls LGTM, but this could benefit from @AnomalRoil 's review.

pairing/bls12381/bls12381_test.go
K1li4nL: No good reason apparently, I didn't expect `*time.T` and `*time.B` to be swappable, but it seems to work that way :+1:

K1li4nL:
Turning this back into draft, I didn't properly ran the benchmarks... It seems that our bdn implementation assumes kyber.Scalar to be mod.Int, making it incompatible to use with circl_bls12381

sign/bls/bls.go
```diff
@@ -2,7 +2,7 @@
 // was introduced in the paper "Short Signatures from the Weil Pairing". BLS
 // requires pairing-based cryptography.
 //
-// Deprecated: This version is vulnerable to rogue public-key attack and the
+// This version is vulnerable to rogue public-key attack and the
```
AnomalRoil, line 5: Maybe specify it's in case of aggregation only and that regular BLS signatures are fine

sign/bdn/bdn.go
```diff
@@ -61,14 +61,39 @@ func hashPointToR(pubs []kyber.Point) ([]kyber.Scalar, error) {
...
+		b, err := mod.NewIntBytes(out[i*16:(i+1)*16], modulus128, kyber.LittleEndian).MarshalBinary()
+		if err != nil {
+			return nil, err
+		}
+		if g.Scalar().ByteOrder() == kyber.BigEndian {
+			reverse(b, b)
+		}
+
+		coefs[i] = g.Scalar()
+		coefs[i].SetBytes(b)
```
AnomalRoil, line 73: `g.Scalar` is of size 32 whereas the coefs in `bdn` are meant to be of size 128 bits, e.g. 16.
I think that's explaining the issues you've had so far.

Also why not just:

```suggestion
		b, err := mod.NewIntBytes(out[i*16:(i+1)*16], modulus128, g.Scalar().ByteOrder()).MarshalBinary()
		if err != nil {
			return nil, err
		}

		coefs[i] = b
```

K1li4nL: Although it looks cleaner, this is a breaking change, these bytes in `out` were always interpreted in little-endian mod 2^128 with no regard to the endianness of the suite.

sign/bdn/bdn.go
```diff
@@ -61,14 +61,39 @@ func hashPointToR(pubs []kyber.Point) ([]kyber.Scalar, error) {
...
 	}
 
 	return coefs, nil
 }
 
+// reverse copies src into dst in byte-reversed order and returns dst,
+// such that src[0] goes into dst[len-1] and vice versa.
+// dst and src may be the same slice but otherwise must not overlap.
+func reverse(dst, src []byte) []byte {
+	if dst == nil {
```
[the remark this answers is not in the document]
K1li4nL: good point, I enforced the equality, there and where I copied this function in `mod/int`

sign/bdn/bdn.go
AnomalRoil: I think this is getting changed in #553 anyway now, no?

K1li4nL: Yes, I will remove that part and rebase on #553

AnomalRoil (approved):

PR #543, opened 2024-08-26
Add P256 test vectors

K1li4nL:


pierluca (commented):
What is the license for this data ? 🤔 ideally, it'd be good to include that and verify that it's compatible with MPL 2.0, Kyber's license.

K1li4nL:
> What is the license for this data ? 🤔 ideally, it'd be good to include that and verify that it's compatible with MPL 2.0, Kyber's license.

So far I can't find any either, I will try to contact the author

pierluca:
Also, is Sonar very wrong here or am I missing something ? 🤔

PR #549, opened 2024-09-17
bls12-381: Add edge case signature

Kubuxu:
This signature does not validate when using kilic@master
Add it to tests to highlight the issue and prevent this implementation from updating to it.
    
The issue appears in this commit: https://github.com/kilic/bls12-381/commit/7536ae1cb938a818cffabc34c4f9a7335b1c7f9a and is not related to assembly code (it reproduces even on M series Macs).

Additionally, synchronise the default DST values between circl and kilic.


pairing/bls12381/circl/g2.go
```diff
@@ -97,5 +97,7 @@ func (p *G2Elt) Mul(s kyber.Scalar, q kyber.Point) kyber.Point {
 
 func (p *G2Elt) IsInCorrectGroup() bool { return p.inner.IsOnG2() }
 
-func (p *G2Elt) Hash(msg []byte) kyber.Point       { p.inner.Hash(msg, nil); return p }
+var domainG2 = []byte("BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_")
+
+func (p *G2Elt) Hash(msg []byte) kyber.Point       { p.inner.Hash(msg, domainG2); return p }
 func (p *G2Elt) Hash2(msg, dst []byte) kyber.Point { p.inner.Hash(msg, dst); return p }
```
AnomalRoil, line 103: I wonder if we really need Hash2 still in V4 if we make the Hash take a DST arg by default. Cc @K1li4nL

pairing/bls12381/circl/g1.go
```diff
@@ -97,5 +97,7 @@ func (p *G1Elt) Mul(s kyber.Scalar, q kyber.Point) kyber.Point {
 
 func (p *G1Elt) IsInCorrectGroup() bool { return p.inner.IsOnG1() }
 
-func (p *G1Elt) Hash(msg []byte) kyber.Point       { p.inner.Hash(msg, nil); return p }
+var domainG1 = []byte("BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_NUL_")
```
AnomalRoil, line 100: I wonder if the DST for the curves shouldn't be the ones as per RFC9380, and the BLS signature DST shouldn't be specified in the BLS and BDN packages rather.
Cc @K1li4nL as well.

pairing/bls12381/circl/g2.go
K1li4nL: Unless there are good reasons to have this flexibility I suppose we can get rid of `Hash2()` and have a default DST yes

pairing/bls12381/circl/g1.go
K1li4nL: Ideally yes, this would allow us to include the RFC test vectors easily for that curve

Kubuxu: This was to match the behaviour of circl to kilic. If there is anything I can do here, please don't hesitate to let me know.

Kubuxu: With the current interface (I guess that is the change you discussed in the other thread), there is no way for the BLS package to specify the DST, otherwise I would have done it there.

PR #553, opened 2024-09-22
Use the group-specific scalar type when hashing in BDN

Stebalien:
Previously, `hashPointToR` would always use `mod.Int` but that only works with the Kilic backend. This change makes the BDN scheme work with all backends.

This PR is currently based on #546 as it touches masks. You can ignore everything but the last commit.

### DIFF
--- a/sign/bdn/bdn.go
+++ b/sign/bdn/bdn.go
@@ -13,25 +13,20 @@ import (
 	"crypto/cipher"
 	"errors"
 	"fmt"
-	"math/big"
+	"slices"
 
 	"go.dedis.ch/kyber/v4"
-	"go.dedis.ch/kyber/v4/group/mod"
 	"go.dedis.ch/kyber/v4/pairing"
 	"go.dedis.ch/kyber/v4/sign"
 	"go.dedis.ch/kyber/v4/sign/bls"
 	"golang.org/x/crypto/blake2s"
 )
 
-// modulus128 can be provided to the big integer implementation to create numbers
-// over 128 bits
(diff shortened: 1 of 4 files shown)


AnomalRoil:
@Stebalien seems your PR didn't like your other PR getting squashed, sorry for that I somehow thought Github handled that automagically.

Also it seems you might be touching some of the same bits and bobs as in #540 which removes insecure aggregation from the `bls` package, not sure which one we should be merging first. @K1li4nL opinion?

Stebalien:
Rebased (although something is wrong with CI).

This should only touch the bdn package so it _shouldn't_ interfere too much with that PR, although it does change the `NewMask` constructor.

K1li4nL:
> @Stebalien seems your PR didn't like your other PR getting squashed, sorry for that I somehow thought Github handled that automagically.
> 
> Also it seems you might be touching some of the same bits and bobs as in #540 which removes insecure aggregation from the `bls` package, not sure which one we should be merging first. @K1li4nL opinion?

I stumbled upon the group specific problem when removing the insecure part of BLS, I solved it pretty similarly though this fix is cleaner than mine. I can remove my fix from my pr and rebase on top of that one.

K1li4nL:
> Rebased (although something is wrong with CI).
> 
> This should only touch the bdn package so it _shouldn't_ interfere too much with that PR, although it does change the `NewMask` constructor.

The linter is complaining about commented code left over, I guess that's why the rest of the pipeline got canceled

Stebalien:
Ah, that would make sense. Let's see if this works...

AnomalRoil (approved):

PR #559, opened 2024-11-08
upgrade to go 1.23

jbsv:


AnomalRoil (commented):
I think we should take advantage of the golangci-lint-action and the way it's nicely integrated with Github Action and reports the linter issue with nice formatting rather than using `make lint` in Github Actions.

PR #567, opened 2025-02-19
Upgrade go and dependencies

jbsv:
upgrade to go 1.22
upgrade project dependencies to avoid dependabot


AnomalRoil (commented):
This looks like (almost?) the same PR as #559, no?

